PR #246, opened 2016-03-09
Bunch of fixes for warnings issued by GCC 6

iljah:


galfthan:
I will close this one since time went past it. It is still a very valid effort to do it though again, but probably easier to do from scratch.

Issue #277, opened 2016-08-24
Fix test_fp to be runnable and used in the testpackage again


ykempf:
test_fp does not run at the moment, a few small project changes make it call base class functions instead of the derived ones.

Due to that it hasn't been used in the testpackage in the last year or so.
- [ ] Fix the default cfg.
- [ ] Define a good test for the test package -- more than one cell might be useful.

PR #281, opened 2016-08-24
Fixes to test_fp and testpackage reintegration

ykempf:
This first commit makes test_fp run again. The cfg should be updated and the integration to the test package discussed (which setup, the one here initialises more than one cell, which might be more useful to test).

Linked with #277.


ursg:
I can confirm that this indeed fixed the test_fp run of the testpackage. So I approve this to be merged.

PR #291, opened 2016-09-27
Ldz whistler investigations

ykempf:
NOT TO BE MERGED

This is only to have a colourful diff at hand.


galfthan:
I guess the only relevant things here, not covered by #305 is the last comment on B which is not recomputed. 

In my opinion we should just recompute Bvol if and when that is more accurate. Performance can then be gained in other ways (fsgrid etc.).

We should open an issue on that, or even a new branch with pull request, and continue discussion there after closing this.

ykempf:
#310 created.

PR #301, opened 2016-12-05
IP Shock project

markusbattarbee:
This is a pull request for the interplanetary shock investigations, from August-October 2016, somewhat cleaned up. Included is also IDL code for calculation of input parameters using R-H equation assumptions.

Code was extended to support tangential components in both y and z directions, but these have been tested only lightly. The well-tested plane is x-z.

This pull request also tests methodology after switching to public repositories.


markusbattarbee:
Would it be preferable that I hide the development history of this branch and do a new pull request from a "clean" branch? Should I also include the COPYING and rules of the road stuff?

ykempf:
I don't think there is a need to obfuscate the development history, keeping it visible makes future developments/debugging easier. Make sure that you have the correct texts included (copying etc.) in any new files you commit.

galfthan:
We are ready to merge this, but the initial text needs to be added first. Urs will do that today.

ursg:
I now have a commit on top of this in https://github.com/ursg/vlasiator/tree/IPshock (for which I apparently can't create a pull request to go on top of this one). Once you merge that, we're good to go here.

markusbattarbee:
There, license headers added!

PR #303, opened 2017-01-30
New machine fix

ursg:
A couple of small code fixes to build on a freshly-installed machine with GCC 6.

Includes the removal of boost::array (replaced by std::array) and some missing headers.
Tested to run with a small Magnetosphere project run on my new machine at HY.


galfthan:
Seems fine. The only thing is that the asserts are already fixed in master, I do not know why they show up in the diff here In any case the second commit in this pull is thus in principle not needed.

galfthan:
I looked into it, and it is because github only shows the diff to the last common ancestor, not the the latest version. See https://developer.atlassian.com/blog/2015/01/a-better-pull-request/

Issue #276, opened 2016-08-24
Update moments during acceleration subcycling


ykempf:
See below for the description:

galfthan:
Re-named original repo. Changes in https://github.com/galfthan/vlasiator-old/commit/05618dd7967b5f88250a8d746a3fe7adec6b26f7

PR #308, opened 2017-02-07
Fix update moments in acc

galfthan:
This branch fixes issue #276 

Verification of this code changes still to be done.


ursg:
This creates zero differences in the test package, but the test package doesn't properly cover subcycling... so this still needs to be tested some other way.

At least it doesn't break anything when there is no subcycling.

galfthan:
Let's add the acc subcycling test, and then re-test. I'll do that

ursg:
Well, as we discussed, I'll do this afterall. :)

ursg:
Indeed, the substeps test shows a bit of differences:

```
running acctest_3_substeps 
--------------------------------------------------------------------------------------------
acctest_3_substeps  -  Verifying 344ee07903fba0468733343d23d4b4f1efed3877___DACC_SEMILAG_PQM__DTRANS_SEMILAG_PPM__DDP__DDPF__DVEC4D_AGNER__DACC_SEMILAG_PQM__DTRANS_SEMILAG_PPM__DDP__DDPF__DVEC4D_AGNER against current
--------------------------------------------------------------------------------------------
NA        NA         NA
  variable     |     absolute diff     |     relative diff | 
rho_0                8.55601e-05                 2.13898e-11    
rho_v_0                2.66324e+07                 0.0449613    
rho_v_1                2.51376e+07                 0.0657062    
rho_v_2                1.49609e+07                 0.0223951    
B_0                0                 0    
B_1                0                 0    
B_2                0                 0    
E_0                0                 -1    
E_1                0                 -1    
E_2                0                 -1    
--------------------------------------------------------------------------------------------
   Distribution function diff                                                               
--------------------------------------------------------------------------------------------
INFO Reading in two files.
File names: /mnt/btrfs/work/uganse/testpackage/testpackage_data/current/acctest_3_substeps/fullf.0000001.vlsv & /mnt/btrfs/work/uganse/testpackage/run_2017.02.22_16.37.37/acctest_3_substeps/fullf.0000001.vlsv
NonIdenticalBlocks:      9
IdenticalBlocks:         2188
Absolute_Error:          9.55971e-11
Mean-Absolute-Error:     2.22376e-15
Max-Absolute-Error:      2.39817e-13
Absolute-log-Error:      436.994
Mean-Absolute-log-Error: 0.0101653
```

ursg:
f1

The differences in distribution functions is indeed miniscule. So the ~5% difference in rho_v apparently is primarily a result of the changed moment calculation?

ursg:
Looking at the actual rho_v vectors: they have the same length, but with this pull request, they are rotated 2.15 degrees further than on master. This makes sense, because the output that's written should be a couple of substeps later.

Except the cfg is setup in such a way that one step rotates by exactly 10 degrees, and one substep should be 1 degree difference. So it doesn't fit that this is 2-and-a-bit substeps off?

galfthan:
So you compare master against new version, both with subcycles? Which is more correct when compared to the original value? Since we do 10 complete rotations it would in a perfect world be the same.

ursg:
substeps

Here's the comparison of velocity space shape (pseudocolor) and bulk velocity vector (black arrows) between master without substeps, with substeps, this branches' substeps and the same with twice the vspace resolution. If the black arrows aren't visible, they are to small to be rendered.

ursg:
the black arrows are multiplied in length by 3e3, just to be visible. :)

PR #311, opened 2017-02-21
Updated testpackage (subcycle test and polar magnetosphere)

ursg:
This adds two additional test cases to the testpackage:

acctest_3_substeps, based on acctest_2

and

test_fp_substeps, based on test_fp_fsolver_only_3D

For now, they are simply comparing their results to the reference data version of themselves, but we can consider also comparing the subcycled versions to the non-subcycled ones.


ursg:
To make this a little more readable, the diffs between the specific tests are:

```
--- acctest_2_maxw_500k_100k_20kms_10deg/acctest_2_maxw_500k_100k_20kms_10deg.cfg	2016-09-07 11:09:54.000000000 +0300
+++ acctest_3_substeps/acctest_3_substeps.cfg	2017-02-08 11:49:09.461841285 +0200
@@ -64,6 +64,10 @@
 diagnostic = RhoLossAdjust
 diagnostic = RhoLossVelBoundary
 
+[vlasovsolver]
+maxSlAccelerationRotation = 1
+maxSlAccelerationSubcycles = 20
+
 [sparse]
 minValue = 1.0e-16

```

and

```
--- test_fp_fsolver_only_3D/test_fp_fsolver_only_3D.cfg	2016-09-07 11:09:54.000000000 +0300
+++ test_fp_substeps/test_fp_substeps.cfg	2017-02-21 10:59:52.013558190 +0200
@@ -69,3 +69,8 @@
 Bdirection = 3
 shear = 0
 
+[fieldsolver]
+maxSubcycles = 20
+maxCFL = .15
+minCFL = .05
+
```

ursg:
Ah, apparently, these should have static dt, otherwise they are not directly comparable.

ykempf:
Yes, if both versions don't hit the same final time, then e.g. test_fp variations are meaningless.

ursg:
Now added a change to vlasiator, so that subcycling can be used even if a fixed timestep is specified in the config file. This is neccessary for the testpackage to actually employ subcycling, and still provide reproducible results.

ursg:
Ok, I confirm this all works and produces good test results. Nothing in the way of merging this!

PR #305, opened 2017-02-07
Field solver stepping improvements

ykempf:
This wraps up the 20161014 comment on #280. As mentioned there, this is still not stable enough but very likely a significant improvement to have in.

PR #313, opened 2017-02-22
Interface to call project once every step.

ykempf:
This is done now mostly for Dispersion.

- [x] @ursg can you test with the testpackage? Just to test that it does not crash with different projects.

I ran Dispersion and got identical results compared to the last test I did in #305 so it did not affect anything, as expected obviously.

Suggestions on the implementation and naming welcome.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -244,7 +244,7 @@ version.o: version.cpp
 	 ${CMP} ${CXXFLAGS} ${FLAGS} -c version.cpp
 
 amr_refinement_criteria.o: ${DEPS_COMMON} velocity_blocks.h amr_refinement_criteria.h amr_refinement_criteria.cpp object_factory.h
-	${CMP} ${CXXFLAGS} ${FLAGS} ${MATHFLAGS} -c amr_refinement_criteria.cpp
+	${CMP} ${CXXFLAGS} ${FLAGS} ${MATHFLAGS} -c amr_refinement_criteria.cpp ${INC_DCCRG}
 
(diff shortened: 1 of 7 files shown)


galfthan:
would project->hook(location, myRank, cells, mpiGrid);  be better? location would be an enum where we could initially only define atTimeStepEnd. Later other locations, if need arises. 

Perhaps even project->hook(location,  mpiGrid);, and fetch the others internally.

ykempf:
As for location, maybe, do you foresee a need for that in the near- or mid-term?

I passed the vector and rank because I didn't want to have an extra call to getLocalCells() and MPI_Rank(). If you prefer to hide that away instead, I can do that.

galfthan:
It may be useful for other validation/debugging tests, and may make them faster by reducing the need to write vlsv files and call tools on them. This would be the most obvious benefit. E.g. writing velocity space as vlsv when doing the amr version, without adding hooks by hand. 

The cells are already cached here: https://github.com/fmihpc/vlasiator/blob/master/parameters.h#L72 so one does not really need to pass them, or call dccrg to get local cells. MPI_Rank() is not too heavy, I would not hesitate to call that inside the hook.

ykempf:
Better?

ykempf:
Tested with Dispersion, that works. @ursg I still think one could pro forma test the checkbox above.

ursg:
Can confirm that the testpackage runs are unaffected by this. Checking the box.

PR #307, opened 2017-02-07
Fs subcycle cfl fix

ursg:
This replaces pull request #229, it was simply rebased onto current master.

The testpackage runs fine with it, but that is unsurprising, since the testpackage doesn't subcycle.

PR #316, opened 2017-02-23
Revert last commit of #307: Moment derivatives are actually required.

ursg:
Since within the substeps, the moments are being accessed at both RK2
times alternatingly, they need to be computed back and forth.

This *might* cost some performance, or then again it might be
negligible. If it is to costly, they could also be precomputed and
stored.

### DIFF
--- a/fieldsolver/derivatives.cpp
+++ b/fieldsolver/derivatives.cpp
@@ -36,7 +36,6 @@ extern map<CellID,uint> existingCellsFlags; /**< Defined in fs_common.cpp */
  * \param cellCache Field solver cell cache
  * \param sysBoundaries System boundary conditions existing
  * \param RKCase Element in the enum defining the Runge-Kutta method steps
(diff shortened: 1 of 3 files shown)


ursg:
Looking at testpackage performance, this doesn't seem to have a big impact at all. Some tests actually ran faster (then again, this is a desktop machine).

ykempf:
The communication pattern can be optimised after https://github.com/fmihpc/vlasiator/blob/master/fieldsolver/derivatives.cpp#L413, as the moments won't change (until we *polate them).

ykempf:
As discussed with Urs now, the new scheme will be:
- always recalculate the derivatives of moments during fs subcycling (can be optimised, but would require a doubling of the moment derivative fields);
- communicate the moments only once (new optimisation, they were always communicated when computing the derivatives, along with the PERB communication).

ursg:
Testpackage'd. Runs properly, and is negligibly faster. I suggest merging it.

PR #326, opened 2017-03-15
Allow a maximum memory per node bailout trigger to be set.

ursg:
In order to prevent simulations from simply crashing when they run out of memory, this limit can be set to gracefully bailout once a certain memory level per node (given in GB) is exceeded.

### DIFF
--- a/memoryallocation.cpp
+++ b/memoryallocation.cpp
@@ -27,6 +27,8 @@
 #include <limits>
 #include "logger.h"
 #include "memoryallocation.h"
+#include "common.h"
+#include "parameters.h"
 #ifdef PAPI_MEM
 #include "papi.h" 
 #endif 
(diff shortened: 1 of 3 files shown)


ykempf:
Have you tested its performing as expected?

ursg:
Hmm, if this parameter is not set, it's default value is `std::numeric_limits<Real>::max()`, which *should* be reasonable for it. However, when run like that, the code warns:
```
Problems casting value to double bailout.max_memory = 1.797693134862316e+308
```
Why is the parameter code not able to cast a valid double value... to double?

memoryallocation.cpp
```diff
@@ -156,6 +158,7 @@ void report_process_memory_consumption(){
          logFile << "(MEM) High water mark per node (GiB) avg: " << sum_mem_papi[0]/nNodes/GiB << " min: " << min_mem_papi[0]/GiB << " max: "  << max_mem_papi[0]/GiB <<
             " sum (TiB): " << sum_mem_papi[0]/TiB << " on "<< nNodes << " nodes" << endl;
          
+         bailout(sum_mem_papi[0]/nNodes/GiB > Parameters::bailout_max_memory, "Memory high water mark per node exceeds bailout threshold", __FILE__, __LINE__);
```
ykempf: I would suggest to use the maximum here?

ursg:
Ok, fixed. Also, the default value for this parameter is now 1 Exabyte/Node, so the double conversion doesn't warn anymore.

parameters.cpp
```diff
@@ -145,6 +145,7 @@ Real P::hallMinimumRho=1.0;
 
 bool P::bailout_write_restart = false;
 Real P::bailout_min_dt = NAN;
+Real P::bailout_max_memory = 1e9;
```
ykempf: Well, Sebastian was using TiB and GiB, maybe make this EiB?

ursg: Do you mean in the comments? Or is that defined as a numeric constant?

ykempf: https://github.com/fmihpc/vlasiator/pull/326/files#diff-e4f6a462e1cc134401251768bee66a90R113
There.

PR #329, opened 2017-05-04
Marconi support

ursg:
Starting with a makefile, this pull request is supposed to collect all required changes to make vlasiator work properly on marconi.

### DIFF
--- a/MAKE/Makefile.marconi
+++ b/MAKE/Makefile.marconi
@@ -1,0 +1,84 @@
+#MARCONI - KNL based cluster
+#
+#The following module commands should give you the correct environment for this makefile
+#module purge
+#module load env-knl
+#module load profile/advanced
+#module load intel
+#module load intelmpi
+#module load boost
+#module swap intelmpi mvapich2/2.2--intel--pe-xe-2017--binary
+
+
+#======== Vectorization ==========
+#Set vector backend type for vlasov solvers, sets precision and length. 
+#NOTE this has to have the same precision as the distribution function define (DISTRIBUTION_FP_PRECISION)
+#Options: 
+# AVX:	    VEC4D_AGNER, VEC4F_AGNER, VEC8F_AGNER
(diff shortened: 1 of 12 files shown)


ursg:
My .profile on marconi lists the following, in order to build vlasiator:
```
module load intel
module load intelmpi
module load boost
module load env-knl

export VLASIATOR_ARCH=marconi

umask 0007
```

ursg:
In the main Makefile, jemalloc also needs to be disabled (since it doesn't work well on KNL), by commenting out the ```-DUSE_JEMALLOC``` line.

Instead, intel thread building blocks' own malloc is being used, and for it to be properly reacheable in the environment, ```. /cineca/prod/opt/compilers/intel/pe-xe-2017/binary/tbb/bin/tbbvars.sh intel64``` needs to be run. I suppose that should be part of the startup files as well.

ursg:
Our working directory on marconi is ```/marconi_work/Pra14_3521```, where the libraries are also located.

ursg:
In order to run vlasiator, the LD_LIBRARY_PATH needs to point to the correct places (so that our self-built shared libraries can be found):

````
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:~/work/libraries/phiprof/lib
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:~/work/libraries/papi/lib
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$BOOST_LIB
LD_PRELOAD=$TBBROOT/lib/intel64/gcc4.7/libtbbmalloc_proxy.so.2
````

galfthan:
Also remember to set VEC16F in main Makefile.  Perhaps this, and USE_JEMALLOC could even be relocated to machine specific makefiles in this branch.

This will be a never-ending struggle otherwise between runs on KNL and Xeon

ursg:
This has been merged into, and superseded by pull request #355

PR #335, opened 2017-06-02
Warn users in project.cpp if spatial cell would be empty

iljah:
With too high minimum dist func threshold I got 0 densities and looks like that was a feature not a bug. A warning would've been nice so this adds it.


galfthan:
One should check/fix to see that it plays nicely with multipop. Only if none of the populations have blocks in a cell does it matter

PR #336, opened 2017-06-09
Multipop

ursg:
TODO

  - [ ] Update testpackage to do the same tests as before, and add multipop test
  - [x]  fix the "cosmetic" problem of the boundary population fields not being updated at restart. This will confuse us in cases like oxygen coming out of the ionosphere but not "visible" in oxygen/rho.
  - [x] Projects in the projects/ directory which do *not* yet support being run with multipop should abort with a warning if more than one population is present.
  - [ ] Split off the remaining boundary options to be population-wise.
  - [ ] Update list of output and diagnostic variables in parameters.cpp to reflect the new system.

The testruns of the last couple of days start to look good, so I think it is time to open a proper pull request  for this.

Multiple populations can now be specified in the configuration files. Warning: this version is not backwards-compatible to old proton-only configs!


ykempf:
We need to fix the boundary parsing. At the moment it does not allow to have e.g. proton, helium maxwellian inflow without specifying a file for oxygen.

ykempf:
We also need to make the ionosphere/magnetosphere tapering radius per-population. Otherwise e.g. ionospheric oxygen densities are present in the same large radius where proton velocity is reducing.

ykempf:
We will also need to fix the "cosmetic" problem of the boundary population fields not being updated at restart. This will confuse us in cases like oxygen coming out fo the ionosphere but not "visible" in oxygen/rho.

ursg:
This has been completely merged into, and superseded by pull request #355

PR #337, opened 2017-07-19
Replace toolutil::getFiles with an implementation based on glob()

ursg:
This fixes the weird vlsvextract bug we have been scratching our heads about for two days, where we had a ```restart.xyz.vlsv``` and some ```restart.xyz.vlsv_part_x_152``` files in the same directory, and the pattern matching code randomly matched the partial files before the actual restart.

Instead, the pattern matching now works as expected from the shell and other unix tools.

### DIFF
--- a/tools/vlsv_util.cpp
+++ b/tools/vlsv_util.cpp
@@ -27,7 +27,7 @@
 
 #include <cstdlib>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


ursg:
This is completely merged into master by #339 already.

PR #319, opened 2017-02-24
FsGrid

ursg:
Since the other acute pull requests are now merged, it's time to focus on getting this into a working state and merging it into master.

This pull request in it's current form is based on the state from last march, and does not cleanly apply to current master -- however, the decision was made to get it running first, and then (painfully) rebase/merge it to work on the current master.

The state of the code is that it manages to complete timesteps without crashing under *some* circumstances, but it is unreliable due to bugs. The number of MPI tasks and size of the simulated system seem to have an impact here, so I suspect the grid coupling setup still has a bug (or the way it is invoked).

Furthermore, not all field data is being transfered back into the mpiGrid before output, so that is still incomplete.

PR #338, opened 2017-07-20
initializeFieldPropagator always, also for test-Vlasov.

ykempf:
Will also be done properly after the fsgrid #319 is merged.


ursg:
Alrighty, ran it through the test package, which yielded zero differences.

Issue #340, opened 2017-08-31
1D shock run blowing up: possible problem in acceleration


ykempf:
Run test location: /proj/vlasov/2D/IPshock/yann/ABA_sw_ABA_dx_1D_1e-20 on Sisu.

Run type: 1D shock, 1000 cells, runs on 5-10 cells in a couple of hours. restart files lie around, the one from t=259 s is before problems start and runs to the crash site in about 20 minutes. Take later ones for quicker testing.

Symptoms: after t=270 s, densities drop at CellID 133, E blows up (O(1)), after which the run doesn't recover and becomes significantly different from the same run with higher vdf thresholds.

I noticed that RHO_R and RHO_V start to differ by a factor of almost 10 just when it starts misbehaving badly (tstep 4627-4628).

Tested versions so far:
4c5e4b5f  misbehaves
25c8d49b Merge pull request #321 from galfthan/ipcc-optimization-acceleration misbehaves 
6139c83dad Merge pull request #313 from ykempf/project_once_per_step_function survives

So it seems that 25c8d49b is the culprit.

First suggestion by @galfthan: If you want to try, you can add -dv here https://github.com/fmihpc/vlasiator/blob/master/vlasovsolver/cpu_acc_map.cpp#L299 , and +dv on the next. This could avoid some potential round off issues.

ykempf:
The suggestion above is run in test_01 and does not seem to help.

ykempf:
Setting the map order to 0 or 2 works in this test, map_order 1 or the nominal random cause the problem to appear.

galfthan:
There is a potential fix in: https://github.com/galfthan/vlasiator/tree/fix-acceleration

The problem seems to arise at https://github.com/fmihpc/vlasiator/blob/master/vlasovsolver/cpu_acc_map.cpp#L309, where we divide an int with WID. The int can (legally) be negative, and in that case the result will be a very large integer since WID is uint. See https://stackoverflow.com/questions/32175191/output-of-dividing-int64-t-by-uint64-t for further discussion.

Seems safer to just make WID int, since we may have other similar uses in the code. Other similar parameters could potentially also be converted to int to be safe.

Further testing needed to see if this fixed all issues, but the test is clearly more stable than before.

ykempf:
As far as I am concerned indeed the blowup of 1D shocks has been fixed. Thanks for your help!

markusbattarbee:
Was there a explanation for why WID was a negative number?

PR #341, opened 2017-09-03
Fix acceleration

galfthan:
Fix for #340 

At https://github.com/fmihpc/vlasiator/blob/master/vlasovsolver/cpu_acc_map.cpp#L309,  we divide an int with WID. The int can (legally) be negative, and in that case the result will be a very large integer since WID is uint. See https://stackoverflow.com/questions/32175191/output-of-dividing-int64-t-by-uint64-t for further discussion.

Seems safer to just make WID int, since we may have other similar uses in the code. Other similar parameters could potentially also be converted to int to be safe.


markusbattarbee:
Close this as it's all already included in #343?

PR #343, opened 2017-10-04
Implementation of per-face outflow reset.

ursg:
This adds parameters to only reapply boundaries in specified directions upon restart.

This was neccessitated (and hence, also successfully tested) in the AEF run, where the incoming shock didn't allow us to reinitialize the ±y boundaries, but we had to switch behaviour of the back wall.


ursg:
This pull request is now outdated, since the same functionality has been merged with vlasiator-4.0 (#355)

Issue #334, opened 2017-05-31
NaNs in E and B after init


iljah:
Using this code:
```
	for infilename in sys.argv[1:]:
		infile = pytools.vlsvfile.VlsvReader(infilename)

		mesh_size = infile.get_spatial_mesh_size()
		cells = [c for c in range(1, int(mesh_size[0] * mesh_size[1] * mesh_size[2] + 1))]

		plot_dims = []
		if mesh_size[0] > 1:
			plot_dims.append(mesh_size[0])
		if mesh_size[1] > 1:
			plot_dims.append(mesh_size[1])
		if mesh_size[2] > 1:
			plot_dims.append(mesh_size[2])

		var = infile.read_variable('B', cellids = cells)
		var = numpy.reshape(var, plot_dims + [3])
		if numpy.sum(numpy.isnan(var)) > 0:
			exit('NaNs encountered when plotting B')
```
I get NaNs from the bulk.0000000.vlsv file. I'm using this config and pretty sure I haven't touched anything related to field solver, how should I debug this?
```
project = Magnetosphere

[io]
diagnostic_write_interval = 1
write_initial_state = 0
restart_walltime_interval = 21000
number_of_restarts = 1000


system_write_t_interval = 1
system_write_file_name = bulk
system_write_distribution_stride = 0
system_write_distribution_xline_stride = 0
system_write_distribution_yline_stride = 0
system_write_distribution_zline_stride = 0


system_write_t_interval = 25
system_write_file_name = distributions
system_write_distribution_stride = 0
system_write_distribution_xline_stride = 15
system_write_distribution_yline_stride = 15
system_write_distribution_zline_stride = 15


system_write_t_interval = 500
system_write_file_name = fullf
system_write_distribution_stride = 1
system_write_distribution_xline_stride = 0
system_write_distribution_yline_stride = 0
system_write_distribution_zline_stride = 0


[gridbuilder]
x_length = 50
y_length = 1
z_length = 50
x_min = -2.0e8
x_max = 2.0e8
y_min = -4e6
y_max = 4e6
z_min = -2.0e8
z_max = 2.0e8
vx_min = -1.0e6
vx_max = +1.0e6
vy_min = -1.0e6
vy_max = +1.0e6
vz_min = -1.0e6
vz_max = +1.0e6
vx_length = 10
vy_length = 10
vz_length = 10


timestep_max = 10000000
t_max = 1


[loadBalance]
rebalanceInterval = 10

[variables]
output = Rho
[...]

PR #345, opened 2017-11-07
Improvements to output variable specification

ykempf:
One fix to issue #334, one replacement of a TODO with a new variable output option, one silent fix to my own local Makefile.

Hold on, I have one more commit in a few minutes.


ykempf:
This can go from my point of view. I tested the #334 fix and the fact that the X, Y, Z etc fields are in the file. Nothing here that would impact the test package.

PR #348, opened 2017-11-14
Line dipole and DRO profiling+threading

ykempf:
This is a bit of a mixed bag but I got entangled in git commits and branches.


ykempf:
I have a separate branch DRO_threading for that. So I'll try to clean this mess up a bit.

PR #349, opened 2017-11-14
Line dipole paper documentation from Sanni.

ykempf:
Here it is! Lo and behold!

Do we want a wiki page on it? If not, can be merged right away.


markusbattarbee:
No need to keep this pull hanging, I think, but I created an issue for the wiki.

ykempf:
Mister lead developer will take care of this shortly.

ursg:
The cheap way out would be a wiki page that simply links to this pdf.

Or were you thinking of a more proper explanation of what's happening?

markusbattarbee:
At some point someone should document it and perhaps type it out in LaTeX. I might do this quite soon. But let's not leave this hanging.

ykempf:
It also hangs on the wall here.

PR #350, opened 2017-11-14
DRO profiling and threading

ykempf:
We had many unthreaded DROs and some (Pressure and its tensor components) threaded within the DRO. This was an attempt at threading all DROs over cells, which is probably more sensible at scale. Open for inspection and criticism, not especially urgent but not hurting either I suppose.

### DIFF
--- a/MAKE/Makefile.yann
+++ b/MAKE/Makefile.yann
@@ -3,6 +3,22 @@ LNK = mpic++
 
 PAPI_FLAG =
 
+#======== Vectorization ==========
+#Set vector backend type for vlasov solvers, sets precision and length. 
+#NOTE this has to have the same precision as the distribution function define (DISTRIBUTION_FP_PRECISION)
+#Options: 
+# AVX:	    VEC4D_AGNER, VEC4F_AGNER, VEC8F_AGNER
+# AVX512:   VEC8D_AGNER, VEC16F_AGNER
+# Fallback: VEC4D_FALLBACK, VEC4F_FALLBACK, VEC8F_FALLBACK
+
+ifeq ($(DISTRIBUTION_FP_PRECISION),SPF)
+#Single-precision        
+	VECTORCLASS = VEC8F_AGNER
+else
+#Double-precision
+	VECTORCLASS = VEC4D_AGNER
+endif
+
 FLAGS =
 #CXXFLAGS = -I $(HOME)/include  -L $(HOME)/lib -g  -funroll-loops -std=c++0x -fopenmp -W -Wall -pedantic -Wno-unused -fabi-version=0 -mavx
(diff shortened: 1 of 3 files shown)


ykempf:
So it seems that it is broken, as data is jumbled in the output file.

ykempf:
Indeed, we need to make the relevant class member variables threadprivate.

ykempf:
I submitted a test of the latest commit.

ursg:
Phew, I can't really say that I like this latest patch.

I mean, sure, making the DataReductionOperator members static and threadprivate probably solves this concrete issue, but it is highly inelegant. And I guess it'll open up an entire new can of worms when it comes to understinding memory allocation and variable lifetimes. When exactly are the thread-local copies of these varibles created? When are they initialized? And when destroyed?

Unfortunately, fixing this properly would probably entail a proper rewrite of the datareduction section of the code. Instead of this weird threadprivate-static-class frankenstein, I would prefer to have an actual thread-local instantiation of the datareductionoperator classes (into a local variable), preferably through some sort of factory pattern. But: that's lots of work.

ykempf:
Well I apologise for the ugliness of the solution, but it was the quickest I came up with... I can confirm that the result is not jumbled any more. I put this on top of my Marconi version to test and launched a small Magnetosphere.

galfthan:
Per cell threading is a good thing, and should be done. 

Is this implementation sensible to be merged is another thing. Many of the DRO's do not need this local storage of variables and the splitting into setSpatialCell() and reduceData() is not needed. Are there any DRO's which actually need it?

By removing the local variables and setSpatialCell() one could make it stateless and easy to thread without all of this complexity.

ykempf:
Well this is "historic" and it takes a day or two of work to clean it up. Maybe less though. So maybe we backtrack on the merge that was done – sorry for that – and do the clean up when we have a bit more time. Or merge this pathc and set a deadline by which the cleanup has to be done.

There was certainly good reasons for the two-stage thing originally otherwise @sandroos wouldn't have done it like so. Possibly this is needed for the operators writing directly to disk. But actually some of the more recent DROs coded by @ohannuks or me for example already only have a `return true;` statement in `setSpatialCell()`.

We could also review the population moments DROs. @sandroos had coded a generic DRO doing all population densities. My current implementation creates one DRO per species. But it probably doesn't make a lot of difference.

galfthan:
Vlasiator has enough cruft as it is, so I think it best not to merge to main version things which increase that.

ykempf:
Yeah but the current master is broken. So let's backtrackon this merge, @ursg...

ursg:
This is not actually in master, but in the vlasiator-4.0 branch, indeed. I'll revert it there.

markusbattarbee:
Since our DROs are not handled by the ARCH looping interface, any threading analysis and improvements should go through there. Thus, closing this.

Issue #332, opened 2017-05-18
Fix IO on Marconi


ykempf:
Initial testing shows that IO is very slow for small files but excellent for big files (order of 10 MB/s for bulk files, 10 GB/s for restarts).

One find is that collective IO seems to be as fast as single IO for small data. So for now a workaround by @sandroos will be tested while waiting for suggestions by the CINECA helpdesk. This seems not to depend on the MPI implementation.

ursg:
We ran on marconi and it wasn't too great. But now that is long done.

Issue #230, opened 2016-02-10
Multiple ion species production ready


galfthan:
Most of the code for using multiple species exists but known and unknown issues exist.

This is the list of known issues:
- [ ] In cpu_acc_transformation the hall term and electron pressure term do not take into account the charge density off all ion species when computing electron charge density as being equal to ion charge density
- [x] Check that charge magnitude and sign is taken into acocunt in all relevant places, e.g., cpu_acc_transformation
- [x] Check that species mass is taken into acocunt in all relevant places, e.g., cpu_acc_transformation
- [ ] Find a suitable solution for the Whistler Speed problem, or a conclusive reasoning why the proton whistler is the only relevant one (see below)

galfthan:
There was a duplicate issue #271

galfthan:
This tracking issue has not been used, @ursg can you update or close?

ursg:
A full campagin of multipop production runs has been performed (and indeed, a number of places where multiple populations had not been taken into account correctly have been found only in that process).

Apart from the fact that results *look* ok now, I'll do another complete read-through of the field solver before I acknowledge this as fixed.

ursg:
Gyration direction of negatively charged particles (compare sebastian's first comment in this issue) is now fixed in https://github.com/fmihpc/vlasiator/pull/368

ursg:
Copied over from pull request #366, regarding which ions to take into account in wave dispersion calculation:

- [ ] Find a suitable solution for the Whistler Speed problem, or a conclusive reasoning why the proton whistler is the only relevant one

As for the whistler (and, to a limited degree, also alfven and soundwave) speed calculation: I dug into the literature a little bit, and found 
https://www.ann-geophys.net/26/1605/2008/ (for the whistler dispersion) and 
http://iopscience.iop.org/article/10.1088/0253-6102/43/2/026/meta (for alfven dispersion).

Both show that the dispersion behaviour of whistler and alfven waves are not *strongly* affected by the presence of heavier ions. For Alfven and sound waves, higher particle mass simply leads to a reduction of wave speeds, in which case our solvers overestimating them would not be too critical.

The Whistler speed however might increase with heavier ions, if only by a tiny amount. The example plots with 10% Oxygen in the paper above do not look bothersome to me, but we should determine up to which limit we want to tolerate departure from the correct behaviour.

ykempf:
We can decide to let it be for now and remember that this is the likely cause if/when we get a new blowup in a future run.

PR #355, opened 2017-11-24
Vlasiator-4.0

ursg:
This pull requests supersedes the previous pull requests #336 (Multipop), #329 (Marconi support) and #319 (Fsgrid), which are merged together and cleanly updated to current master here. It also encompasses issuse #230 and #332.

It has been used for production runs on Marconi (both AED and Yann's shock tests).

Consolidating the TODO lists remaining from the earlier requests, there are still some open ends to be tied here:

  - [x] Update testpackage to do the same tests as before, and add multipop test
  - [x] Split off the remaining boundary options to be population-wise.
  - [x] Split off backstream parameters ? (Discussion below)
  - [x] Fix fsgrid/cellparams Hall term E components IO output. (See below.)
 - [x] Test and merge https://github.com/ursg/vlasiator/pull/19
 - [x] Review all data reduction operators (At least TWO reviewers!). See https://github.com/ursg/vlasiator/pull/21 for details.
 - [ ] Review all data reduction operators (At least TWO reviewers!).

As soon as these are resolved, and current production runs with the old codebase are through, I suggest merging this into master.


ykempf:
I suppose the backstream parameters could/should also be defined population-wise, for the sake of some generality.

ykempf:
I spotted a small bug for IO. Hall E is not being pulled out of fsgrid to cellparams for IO. I will put that into a commit later/ping me about it if I forget. The fix includes Hall E and GradPe E if they are to be written out.

ykempf:
The later commit I also need to compare the new version with a reference test I was running right now.

ykempf:
Whoops there was a wrong parenthesis closure there. Glad I tested and caught it.

I now confirm that the change related to the Hall term in the two commits c80869d and a173ad1 does not modify the E given out and it yields EHall components if the output is set.

ykempf:
I'll take care of the backstreaming parameter splitting.

ykempf:
`ERROR in vlsv::ParallelReader! I only got -32766/0 bytes in vlsv_reader_parallel.cpp:519
ERROR, failed to read BLOCKVARIABLE in ioread.cpp:349
ERROR, failed to read BLOCKIDS in ioread.cpp:345`

Excerpt from Marconi output files.

ykempf:
Also:
slurmstepd: error: Detected 1 oom-kill event(s) in step 586688.1 cgroup.
srun: error: r106c02s02: tasks 5351,5358: Out Of Memory
srun: Terminating job step 586688.1
slurmstepd: error: Detected 1 oom-kill event(s) in step 586688.1 cgroup.
slurmstepd: error: *** STEP 586688.1 ON r066c11s04 CANCELLED AT 2018-03-31T23:03:11 ***

But this happens after the logfile said we're exiting.

ykempf:
https://github.com/fmihpc/vlsv/pull/33 Will fix the former (of the last 2 reported errors).

ykempf:
@ursg, please note the latest commit, I had the population backstream options mistakenly with an underscore instead of a full stop.

sysboundary/outflow.cpp
```diff
@@ -127,6 +89,78 @@ namespace SBC {
...
+        }
+        if(!Readparameters::get(pop + "_outflow.vlasovScheme_face_z+", vlasovSysBoundarySchemeName[4])) {
+           if(myRank == MASTER_RANK) cerr << __FILE__ << ":" << __LINE__ << " ERROR: This option has not been added for population " << pop << "!" << endl;
+           exit(1);
+        }
+        if(!Readparameters::get(pop + "_outflow.vlasovScheme_face_z-", vlasovSysBoundarySchemeName[5])) {
+           if(myRank == MASTER_RANK) cerr << __FILE__ << ":" << __LINE__ << " ERROR: This option has not been added for population " << pop << "!" << endl;
+           exit(1);
+        }
+        for(uint i=0; i<6 ; i++) {
```
ykempf: @ursg I don't like embedded for loops with the same index... Even though you told me it should work all right...

ykempf:
The current <population>/PTensorDiagonal and <population>/PTensorOffDiagonal have the same data as the DRO does a loop over all species. Either this is stored as the global PTensor and does not need to be duplicated per species, or this is supposed to be the partial pressure (and the summation can be done later if needed).

Opinions?

markusbattarbee:
This should definitely be per-species, so we can analyse T_parallel and T_perpendicular per-species.

ykempf:
Here it is.

ursg:
With all the recend feature branches merged into this one, and all production runs of Vlasiator 3.0 complete, I think it is time to merge this into master. The conflcts below already have a recorded preimage in the vlasiator-4.0-preview branch, so they will present no problem.

PR #366, opened 2018-06-12
Code cleanup of fieldsolver

ursg:
Related to #230, I read through the field solver code completely and cleaned up some things here and there.

Still to Do:
- [x] Update Doxygen comments to stop referencing mpiGrid, and instead documenting the FSGrids
- [ ] Find a suitable solution for the Whistler Speed problem, or a conclusive reasoning why the proton whistler is the only relevant one


ursg:
As for the whistler (and, to a limited degree, also alfven and soundwave) speed calculation: I dug into the literature a little bit, and found 
https://www.ann-geophys.net/26/1605/2008/ (for the whistler dispersion) and 
http://iopscience.iop.org/article/10.1088/0253-6102/43/2/026/meta (for alfven dispersion).

Both show that the dispersion behaviour of whistler and alfven waves are not *strongly* affected by the presence of heavier ions. For Alfven and sound waves, higher particle mass simply leads to a reduction of wave speeds, in which case our solvers overestimating them would not be too critical.

The Whistler speed however might increase with heavier ions, if only by a tiny amount. The example plots with 10% Oxygen in the paper above do not look bothersome to me, but we should determine up to which limit we want to tolerate departure from the correct behaviour.

ursg:
As there are no further comments on this, and it doesn't change any actual running code (only touches dead code, comments and formatting), I'll merge this now.

ykempf:
We didn't solve the second checkbox above though, want to open an issue for it?

ursg:
I updated issiue #230 (that was starting this pull request) with the open question there.

PR #368, opened 2018-06-25
Proper gyration of negatively charged particles

ursg:
The velocity space update matrix calculation was effectively using abs(charge), so all particle populations had right-hand gyration, regardless of their actual charge sign.

This pull request also adds a multipop gyration test (proton/antiproton) to the testpackage and fixes the MultiPeak project so that it actually works properly.

### DIFF
--- a/MAKE/Makefile.marconi
+++ b/MAKE/Makefile.marconi
@@ -12,7 +12,6 @@
 
 #======== Vectorization ==========
 #Set vector backend type for vlasov solvers, sets precision and length. 
-#NOTE this has to have the same precision as the distribution function define (DISTRIBUTION_FP_PRECISION)
 #Options: 
 # AVX:	    VEC4D_AGNER, VEC4F_AGNER, VEC8F_AGNER
 # AVX512:   VEC8D_AGNER, VEC16F_AGNER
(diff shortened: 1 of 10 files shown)


galfthan:
Good. You are right that we could also allow gyroperiod to be negative, keep fabs in computation of substep dt, and then substep radians is ok when there is no fabs. Would that be nicer?

ursg:
That would probably be nicer, but I was confused about coordinate systems and whether the gyration of a positively charged particle should have positive or negative angle sign here. So I went the cheap way, and just multiplied it with -1 for negatively charged ones. I'll fix this properly then.

ursg:
Ok, while trying to figure out wether the plus or the minus sign is the right one, I was completely confused by the fact that both of them looked wrong.

Turns out, I screwed up operator precedence in the previous commit, and had always +1 or -1 radians of rotation in every substep. Whoops!

Good that you complained. Now it's fixed in a better way.

ursg:
Just for comparison: this is the behaviour of the proton-antiproton gyration test before this fix (both sitting on top of each other, because gyrating in the same direction):
before0000
And this is the behaviour after (protons gyrating the same way as before, antiprotons the other way):
after0000

PR #369, opened 2018-06-25
Add basic function instrumentation support, for run-time tracing.

ursg:
This allows, for example, to observe the functions called from
vlasiator.cpp without the need to have a debugger attached, and limited
to the master rank. As such, it might be used as a poor man's "debugging at scale".

We're currently testing this for debugging BFA restart failures on marconi, and will then decide whether it is useful to merge or not.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,10 @@ COMPFLAGS += -DNDEBUG
 # CXXFLAGS += -DDEBUG_SOLVERS
 # CXXFLAGS += -DDEBUG_IONOSPHERE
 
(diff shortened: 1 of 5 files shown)


ursg:
This, in the end, turned out to work as intended, but not sufficiently useful to debug a crash on marconi.

The crash went away by itself, with an unmodified executable, and was probably the system's fault.

What do we do with this pull request? Should we merge it and leave it deactivated by default? Should we just leave it dangling? Or get rid of it?

ykempf:
Since it's been worked out and implemented, why not keep it? I guess the extra code for the object wrapper function is not needed generally? Should there be some more documentation in the code for future reference?

PR #371, opened 2018-07-09
Issue 322 calculate wave speed skip

ykempf:
Fixing 322, compiles, not tested yet but "what can possibly go wrong"? If @ursg you could provide me with a location with reference data, I can run the test package myself if you prefer.


ykempf:
Ööh, this of course only includes the last commit as a meaningful change I suppose.

ykempf:
Ööh, this of course only includes the last commit as a meaningful change I suppose. And the line dipole calculation, it seems.

ursg:
Uh, do you want to rebase this so it only contains the relevant commit? What does the line dipole diff contain?

ykempf:
The pdf version of the pencil-and-paper calculations by Sanni.

ursg:
Surprisingly, this gives zero changes with the testpackage, so let's apply it right away!

ykempf:
We just skip useless calculations, no surprise there.

Issue #249, opened 2016-04-11
Magnetosphere distribution picker spatial/velocity sampling asymmetry


markusbattarbee:
In https://github.com/fmihpc/vlasiator/blob/master/projects/Magnetosphere/Magnetosphere.cpp#L165
if only one of nSpaceSamples and nVelocitySamples is different from 1, the picker defaults that one to 1. Using a logical OR for this line is not enough, as in that case for the domain where only 1 sample is required, it would be offset to the side.

ykempf:
I pushed this one out now. It could do with some testing to make sure there is no problem, of course.

Question to @markusbattarbee or @ursg, should I copy this over to all projects with these sampling cfg parameters?

ykempf:
I could also make sure #256 is fixed at the same time.

PR #372, opened 2018-07-11
Fixed sisu testpackage

ursg:
A bunch of little things were missing here and there.
The most important change in this pull request: for the tests where only a single spatial cell is present, the testpackage runs only using a single MPI task, since otherwise fsgrid aborts the run.

Tested on Sisu and produced new reference data for vlasiator-4.0.


ykempf:
Once this is integrated I can also help in testing with some branches like #249/#256.

ursg:
This should now work sufficiently well to be useable for testing other branches.

Some tests still don't work properly though:
- acctest_4_helium doesn't have protons, thus doesn't produce any diff values for those. Would have to have a different mechanism to determine which population to compare
- acctest_5_proton_antiproton only compares the protons for the same reason
- Magnetosphere_small crashes with an MPI error on Sisu for me. This will have to be debugged.

Apart from that, the testpackage *with* this pull request works much better than without already. So I suggest merging.

ykempf:
Does this test "correctly" for restarts? The version I was working on didn't do that correctly yet.

ursg:
Now the "restart_read" test finally properly tests it's output. So I assume that means yes.

Issue #373, opened 2018-07-17
VEC4D_FALLBACK does not compile on Sisu


ykempf:
I am getting the following error:
```
vlasovsolver/cpu_trans_map.cpp:275:71: error: '_MM_HINT_T0' was not declared in this scope
          _mm_prefetch((char *)(blockDatas[b + VLASOV_STENCIL_WIDTH]), _MM_HINT_T0);
                                                                       ^~~~~~~~~~~
vlasovsolver/cpu_trans_map.cpp:275:82: error: '_mm_prefetch' was not declared in this scope
          _mm_prefetch((char *)(blockDatas[b + VLASOV_STENCIL_WIDTH]), _MM_HINT_T0);
```

markusbattarbee:
VEC_FALLBACK was re-activated as part of the openacc branch, but as those changes are not yet in master or dev, leaving this open.

ykempf:
Sisu is decommissioned, FALLBACK was fixed.

Issue #256, opened 2016-06-09
Enable 1 to inf nVelocitySamples and nSpaceSamples in all projects


ykempf:
At least Flowthrough breaks with 1 velocity sample. This should be made consistent across projects.

ykempf:
Note that Alfven uses a global vSpaceSamples whereas e.g. Magnetosphere has a population-specific one. Which one do we prefer? Or do we leave it inconsistent for now?

markusbattarbee:
Nice find - I see no reason to leave it unfixed.

ykempf:
I'll first port the general version allowing 1 to many, and in a second phase check whether I make nSpaceSamples population-specific or not.

ykempf:
So I did see your suggestion @markusbattarbee to put this in a wrapper function, which I like a lot as it will save me a whole lot of copy-pasting, but I don't see it here. Did you remove the comment? I'll start with that anyway, and be most likely done by the time you have a chance to comment back, but I think it is very sensible to have one function instead of 15ish identical implementations.

markusbattarbee:
The response about the wrapper was made to the actual commit :)

PR #374, opened 2018-07-17
Issue 249 vdf sampling

### LINKED ISSUE #249: Magnetosphere distribution picker spatial/velocity sampling asymmetry

ykempf:
Do not merge yet!

I am baffled by this branch. I am trying to resolve #249 and #256 with it. As one can see in the diff, this only touches projects, i.e. initialisation.

However, in conjunction with #372 I want this to sail through the testpackage, and it doesn't.

Some of the testacc and the testtrans fail. In testacc_3 I get differences in the phase space density of order 10⁻¹⁴, where the peak phase space density is of order 10⁻¹⁰ (figure below). This is not acceptable and I have no idea where it could come from. These tests all rely on the MultiPeak project.

I am compiling with testpackage flags and I am comparing to my own master version compiled for this test as well, so no stale reference data here. I tried to run the fallback vectorisation but that won't compile (#373).

@markusbattarbee suggested to check memory allocation; that's done outside of the project. And I do believe that the functions are all declared within their respective class properly, so I don't think that MultiPeak would call another project's `calcPhaseSpaceDensity` or `getDistribValue`.


ykempf:
image

There is differences in the initial state, I shouldn't have trusted the diff of bulk variables. The hunt continues.

ykempf:
Now I think of it, the project functions should use Realf and not Real to avoid problems. But that isn't the cause in DP DPF (testpackage flags).

ykempf:
So indeed @markusbattarbee and I agree that the initialisation of MultiPeak has one, possibly more bugs.

markusbattarbee:
As per our discussion, the bugs should still result in deterministic behaviour, just not exactly as intentioned. Let's see how the comparison goes after fixing them.

ykempf:
The latest commit, applied both to this branch and master, does not change things. I now even compared initial-grid.0000000.vlsv, but I still get small diffs.

ykempf:
Indeed when decreasing the optimisation level of `getDistribValue()` to `O1` or `O0` I get zero diffs. So apparently the fact that the function was virtualised made its optimisation change somehow. Is this somethign we should worry about/investigate further?

ursg:
Hmm, this is indeed odd. Changing the function definition to be virtual of course precludes some of the heavier optimizations to, for example, inline its code into wherever it gets called.
As a result, it *could* be that the order of some floating point operations changes, and thus the values are changing on the order of floating point accuracy, but the magnitudes that we see there seem to large for that.
Especially, I would expect the differences to be mostly pronounced close to the velocity space cutoff, but the plots you show have the biggest diffs in the middle of the distribution function.

So, hum. I'm not sure what to do about it. 10^-23 certainly doesn't look like a *large* error. But diffs are bad nonetheless.

ykempf:
Well the diff is small at initial-grid at t=0, of the order of 10⁻²¹ when the phase space density peaks at O(10⁻¹⁰), but of course it grows quite quickly.

ykempf:
I give up with getting the recursive initialisation of the vdf in MultiPeak to match in the master version and this branch. I am quite confident it is correct now. Of course it should be inspected by others.

ykempf:
As I noticed (again?) that the project functions weren't using Realf, I tried to do that. However often the variables entering a calculation before a Realf is returned are Real. Is this going to create problems? Should we scrap this commit and just use functions with Real instead?

PR #375, opened 2018-07-31
Add a BeforeSetCell hook to Project, for complicated data juggling.

ursg:
Specifically, this is needed for a mechanism to restart a small
simulation as a subset of a larger one, as Thiago is implementing now.


ykempf:
So far so good.

PR #380, opened 2018-09-25
Spatial cell update

tkoskela:
Two minor updates:

- Changed C-style arrays in spatial_cell member variables to c++11 std::arrays. This allows us to use the default copy constructor and operator= for spatial_cell objects.

- Updated dccrg calls to get_neighbors_of() and initialize() to be compatible with the current dccrg master repo (commit 63699d6).

Tested in testpackage and saw no differences to master.


ursg:
Great stuff!

Logistical question: since this requires updating DCCRG at the same time (otherwise people get confusing error messages at compile time), should we just push this through quickly now, before any big electron runs are started, or should we wait with updating DCCRG until after you have merged this, @tvbrito?

tvbrito:
You can go ahead and update this quickly. I can merge it later before doing the next big run.

PR #388, opened 2019-03-19
Fixed a phiprof bug in multipop IO

ykempf:
This fixes a small bug in multipop IO with phiprof. I also updated my local Makefile.


ykempf:
This was tested, by the way.

PR #391, opened 2019-03-28
Update Fluctuations project example cfg file to be multipop-compatible

ursg:


ursg:
This was quickly testrun on my fridge, and seems to work now.

PR #399, opened 2019-05-02
Setup boundary flags directly on fsgrid

tkoskela:


sysboundary/antisymmetric.cpp
```diff
@@ -121,6 +125,41 @@ namespace SBC {
...
          }
       }
+
+         // Assign boundary flags to local fsgrid cells
+      const std::array<int, 3> gridDims(technicalGrid.getLocalSize());  
+      for (int k=0; k<gridDims[2]; k++) {
+         for (int j=0; j<gridDims[1]; j++) {
+            for (int i=0; i<gridDims[0]; i++) {
+               const auto& coords = technicalGrid.getPhysicalCoords(i,j,k);
+               const auto refLvl = mpiGrid.get_refinement_level(mpiGrid.get_existing_cell(coords));
```
galfthan: I believe the coords from the corner of the cell. Then when you get the existing cell, these coordinates are used for dccrgrid. This is also a corner for dccrg (at least at finest level), It would be numerically more stable to add 0.5 DXYZ to coordinates?

sysboundary/ionosphere.cpp
```diff
@@ -156,48 +156,83 @@ namespace SBC {
...
             mpiGrid[cells[i]]->sysBoundaryFlag = this->getIndex();
          }
       }
+
+            // Assign boundary flags to local fsgrid cells
+      const std::array<int, 3> gridDims(technicalGrid.getLocalSize());  
+      for (int k=0; k<gridDims[2]; k++) {
+         for (int j=0; j<gridDims[1]; j++) {
+            for (int i=0; i<gridDims[0]; i++) {
+               const auto& coords = technicalGrid.getPhysicalCoords(i,j,k);
```
galfthan: same comment

sysboundary/outflow.cpp
```diff
@@ -214,28 +214,65 @@ namespace SBC {
...
+            mpiGrid[dccrgId]->sysBoundaryFlag = this->getIndex();
+         }         
+      }
+
+      // Assign boundary flags to local fsgrid cells
+      const std::array<int, 3> gridDims(technicalGrid.getLocalSize());  
+      for (int k=0; k<gridDims[2]; k++) {
+         for (int j=0; j<gridDims[1]; j++) {
+            for (int i=0; i<gridDims[0]; i++) {
+               const auto& coords = technicalGrid.getPhysicalCoords(i,j,k);
```
galfthan: same comment

sysboundary/antisymmetric.cpp
```diff
@@ -121,6 +125,41 @@ namespace SBC {
...
+               creal dz = P::dz_ini * pow(2,-refLvl);
+               creal x = coords[0] + 0.5 * P::dx_ini * pow(2,-P::amrMaxSpatialRefLevel);
+               creal y = coords[1] + 0.5 * P::dy_ini * pow(2,-P::amrMaxSpatialRefLevel);
+               creal z = coords[2] + 0.5 * P::dz_ini * pow(2,-P::amrMaxSpatialRefLevel);
+
+               isThisCellOnAFace.fill(false);
+               doAssign = false;
+               
+               determineFace(isThisCellOnAFace.data(), x, y, z, dx, dy, dz);
+               for(uint j=0; j<6; j++) doAssign = doAssign || (facesToProcess[j] && isThisCellOnAFace[j]);
```
galfthan: You are shadowing a int j used in a loop above on line 132. 

Also, why uint, vlasiator normally uses int for these.

sysboundary/antisymmetric.cpp
tkoskela: Since I add the .5DXYZ on the following line, it's straightforward to do it before this call and also use the updated coords for the get_existing_cell call.

sysboundary/antisymmetric.cpp
tkoskela: The uint is copy-paste from previous code (the loop over dccrg cells above this), but I can change it.

sysboundary/antisymmetric.cpp
tkoskela: Ah, no, it's not that straightforward since I need the refinement level to know how much is 0.5 DX

tkoskela: I could add 0.5 * width of the highest refinement level cell

sysboundary/antisymmetric.cpp
```diff
@@ -121,6 +125,46 @@ namespace SBC {
...
+      // Assign boundary flags to local fsgrid cells
+      const std::array<int, 3> gridDims(technicalGrid.getLocalSize());  
+      for (int k=0; k<gridDims[2]; k++) {
+         for (int j=0; j<gridDims[1]; j++) {
+            for (int i=0; i<gridDims[0]; i++) {
+               const auto& coords = technicalGrid.getPhysicalCoords(i,j,k);
+
+               // Shift to the center of the fsgrid cell
+               auto cellCenterCoords = coords;
+               cellCenterCoords[0] += 0.5 * P::dx_ini * inv2powMaxRefLvl;
```
galfthan: This code is quite complex to get the fsgrid dx, it also makes the assumption fsgriddx is the same as finest level dccrg grid one (though we do that in many places so ok). Now fsgrid leavs the DX DY DZ as public variables, perhaps just use those? I guess somebody may want to later add a get function for those

sysboundary/antisymmetric.cpp
galfthan: see my comment, but otherwis ok

galfthan:
I added a comment that could be changed, but ok also as is. Ok for me, but did not run or check that this works

sysboundary/antisymmetric.cpp
tkoskela: Oh, okay, I did not realize they are public. But indeed we do make the assumption that fsgrid has the same resolution as the finest level dccrg in a lot of places right now.

PR #401, opened 2019-05-09
Added calculation for tiltable x-limited vector potential dipole field

markusbattarbee:


markusbattarbee:
WIP, I'll add option to scale to IMF field at upstream boundary

markusbattarbee:
dipolemodels_vectors_1
dipolemodels_vectors_2
dipolemodels_streamlines_1
dipolemodels_streamlines_2
vecpotdip_verify_streamlines
Screenshot from 2019-05-14 10-54-30

markusbattarbee:
Good to go, IMO

ykempf:
- [ ] Could you include a documentation of that, either in the doc directory or as a wiki page? Both mathematical and practical (what the user should set to achieve a given setup)? And maybe also explaining what the python files are for (should they be in a separet (sub)directory or not?)?
- [ ] Are the resulting configurations always curl-free?
- [ ] What are the *_unused files for?

markusbattarbee:
_unused files were for a dipole which scales away radially instead of as a function of x. That is what I originally coded, but decided that it's not worth implementing as the field becomes really squished. Can be removed.

However, all that might be moot, as the field is not curl-free. I guess it's not worth implementing then after all.

ykempf:
Then it'd have to be put into the perturbed field instead and we'd have to live with the potential problems of having worse derivatives and averages in regions of strong gradients.

Can the curl-freedom be used to constrain the solution?

markusbattarbee:
Analytic calculations show that nope, there's no elegant way past this.

So this ended up being an interesting exercise, no more. 
Next step: implement instead tilting of the dipole+mirror dipole and automate placement of the mirror dipole based on FSgrid values.

ursg:
*technically* of course, also the dipole itself is not curl free (as the https://en.wikipedia.org/wiki/Helmholtz_decomposition requires), but it's curly areas are completely located within the inner boundary. I guess any other solution in which the curl would be contained to the boundary domains would be ok, but I'm doubtful that those would give us any reasonable field configurations....

markusbattarbee:
Screenshot from 2019-05-21 16-20-05

I now modified it so that the BGB field has just a single tandard dipole field, and the perturbed field contains the difference between that and the vector potential field. That way, all the currents generated by the curl of this field are placed in the perturbed field and accounted for properly.

Ran a test for 20 seconds until it crashed, looked ok so far

markusbattarbee:
Current progress after 7h

6dtest_init
6dtest_50s
6dtest_110s_tilt2
6dtest_110s_tilt1

markusbattarbee:
test6d_110s_xz
test6d_110s_xy

Tail lobe generation is nice and fast, but the fieldsolver timestep went way down at some point.

markusbattarbee:
[Magnetosphere]
constBgBX = 0.0
constBgBY = 0.0
constBgBZ = -5.0e-9
noDipoleInSW = 1.0
dipoleType = 4
#dipoleMirrorLocationX = 5.64e8                                                                                    
dipoleTiltPhi = 0.0
dipoleTiltTheta = 0
dipoleXFull = 9.5565e7 # 15 RE                                                                                     
dipoleXZero = 2.5e8
dipoleInflowBX = 0.0
dipoleInflowBY = 0.0
dipoleInflowBZ = 0.0
refine_L2radius = 9.5565e7 # 15 RE                                                                                 
refine_L2tailthick = 3.1855e7 # 5 RE                                                                               
refine_L1radius = 1.59275e8 # 25 RE                                                                                
refine_L1tailthick = 6.371e7 # 10 RE

PR #402, opened 2019-05-09
Added a loss cone population DRO for precipitation

maximegrandin:
Added a data reduction operator to save the loss-cone population of ion species in each cell.
- At the moment, the number of energy bins (16) as well as the minimum and maximum bin values (100 eV and 100 keV) are hardcoded. Energy bins are logarithmically spaced. Particles with energies under 100 eV (above 100 keV) are counted with the lowest (highest) energy bin.
- The loss-cone opening angle has a fixed value which is the same in the whole simulation domain. The value was set to 10 deg, which should be ok as a first approximation, since the phase-space density is averaged inside the loss cone (not integrated).
- At the moment, the loss-cone population is saved everywhere, including in regions where it is not relevant for particle precipitation.
- Differential ion number flux values are saved in units used by observers [ion/(cm2 s sr eV)], which is a deviation from the convention of manipulating SI units only.

### DIFF
--- a/datareduction/datareductionoperator.h
+++ b/datareduction/datareductionoperator.h
@@ -507,6 +507,25 @@ namespace DRO {
       std::string popName;
    };
    
+   // Precipitation directional differential number flux
+   class VariablePrecipitationDiffFlux: public DataReductionOperator {
+   public:
+      VariablePrecipitationDiffFlux(cuint popID);
+      virtual ~VariablePrecipitationDiffFlux();
+      
+      virtual bool getDataVectorInfo(std::string& dataType,unsigned int& dataSize,unsigned int& vectorSize) const;
+      virtual std::string getName() const;
+      virtual bool reduceData(const SpatialCell* cell,char* buffer);
+      virtual bool setSpatialCell(const SpatialCell* cell);
+      
+   protected:
+      uint popID;
+      std::string popName;
+      int nChannels;
+      Real emin, emax;
+      Real cosAngle;
+      std::vector<Real> channels, dataDiffFlux;
+   };
 } // namespace DRO
 
 #endif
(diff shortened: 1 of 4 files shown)


ykempf:
I suggest that you make the hard-coded parameters cfg parameters. It's best if someone on-site shows you the basics, or if you have a bit of time, look at how it's done in parameters.cpp and parameters.h. Add a declaration (add) and a retrieval function (get) like other such scalar parameters, as part of the variables group.

testpackage/tests/Magnetosphere_polar_small/Magnetosphere_polar_small.cfg
```diff
@@ -32,7 +32,7 @@ y_min = -5.0e6
 y_max = 5.0e6
 z_min = -250.0e6
 z_max = 250.0e6
-t_max = 20.05
+t_max = 1000.05
```
markusbattarbee: I think we don't want to edit the testpackage cfg files ;)

datareduction/datareductionoperator.h
```diff
@@ -507,6 +507,25 @@ namespace DRO {
...
+      virtual bool reduceData(const SpatialCell* cell,char* buffer);
+      virtual bool setSpatialCell(const SpatialCell* cell);
+      
+   protected:
+      uint popID;
+      std::string popName;
+      int nChannels;
+      Real emin, emax;
+      Real cosAngle;
+      std::vector<Real> channels, dataDiffFlux;
```
markusbattarbee: You don't actually end up using the channels vector for anything

markusbattarbee:
Note, parameters.cpp needs to also be updated with the name of the data reducer

ykempf:
> Note, parameters.cpp needs to also be updated with the name of the data reducer

What do you mean @markusbattarbee?

markusbattarbee:
See new pull request introducing how DROs can write parameters to the .vlsv file

maximegrandin:
Will be included in upcoming pull request.

PR #404, opened 2019-05-14
Dev fix ionosphere

ykempf:
Any comments?

### DIFF
--- a/parameters.cpp
+++ b/parameters.cpp
@@ -89,7 +89,7 @@ std::vector<std::pair<std::string,std::string>> P::systemWriteHints;
 
 Real P::saveRestartWalltimeInterval = -1.0;
 uint P::exitAfterRestarts = numeric_limits<uint>::max();
-uint64_t P::vlsvBufferSize;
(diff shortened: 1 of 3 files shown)


ursg:
Uh, is this still reasonably fast? Doesn't this potentially cause the layer identification loop to go from being evaluated 3 times to being evaluated multiple thousands of times?

Maybe we can early-exit that loop, if no more cells will be tagged as the next layers?

ykempf:
Nah. Let's do one pass over all cells setting the layer. Then a second classic pass, as is done above with DCCRG.

ykempf:
BRB

ykempf:
Oh nooo, something is tangled in my git now apparently...

ykempf:
This looks better now... Doesn't it?

tkoskela:
Looks good to me

ykempf:
With the last commit this allows the testpackage tests to all run. Actual checking against another version fails for all E and B comparisons and there seems to be multipop issues too, but at least the runs go through.

PR #407, opened 2019-05-15
Fixed initialisation of layers/boundaries, technical grid needs updating

ykempf:
My suggestion, posted as a PR for safety...

### DIFF
--- a/sysboundary/sysboundary.cpp
+++ b/sysboundary/sysboundary.cpp
@@ -409,6 +409,7 @@ bool SysBoundary::classifyCells(dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Ca
    // sysBoundaryLayer communicated)
    SpatialCell::set_mpi_transfer_type(Transfer::CELL_SYSBOUNDARYFLAG);
    mpiGrid.update_copies_of_remote_neighbors(SYSBOUNDARIES_NEIGHBORHOOD_ID);
+   technicalGrid.updateGhostCells();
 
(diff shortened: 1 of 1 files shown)


ykempf:
The first reviewer who is ok with this can merge it. :)

sysboundary/sysboundary.cpp
```diff
@@ -409,6 +409,7 @@ bool SysBoundary::classifyCells(dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Ca
    // sysBoundaryLayer communicated)
    SpatialCell::set_mpi_transfer_type(Transfer::CELL_SYSBOUNDARYFLAG);
    mpiGrid.update_copies_of_remote_neighbors(SYSBOUNDARIES_NEIGHBORHOOD_ID);
+   technicalGrid.updateGhostCells();
```
galfthan: My only comment is that I would move this to around line 477, so that the technical grid layer computation is in one place. Now the dccrg layer computation is inbetween

galfthan: Other than that it is fine :)

ykempf: Like so?

markusbattarbee:
Does this mean the "No closest cells found!" (ionosphere.cpp, line 553) check should be kept in?

ykempf:
Yes, it does. That was also the topic of one of yesterday's PRs, but there were more problems with this algorithm there initialising the layers and setting DO_NOT_COMPUTE cells...

PR #409, opened 2019-05-15
Fix check cfg output variables

markusbattarbee:


markusbattarbee:
I noticed that this doesn't recognize output variables which have a space after the variable name.
e.g. "output = proton/VNonBackstream" is correctly identified, whereas 
"output = proton/VNonBackstream " is not.

ykempf:
Do you want to fix this or shall I look into it? We can remove trailing spaces from all lines before further processing I guess.

PR #405, opened 2019-05-14
Cleanup grid coupling

tkoskela:
Removed obsolete grid glue code on the vlasiator side. Compiles and runs, Flowthrough diffs to 0 with current dev.
- getBgFieldsAndDerivativesFromFsGrid, getDerivativesFromFsGrid and all calls to them have been removed
- getFieldsFromFsGrid is called in initializeGrids in place of getBgFieldsAndDerivativesFromFsGrid
- List of removed variables:
-- fieldsolver namespace entirely
-- CellParams::BGBX,BGBY,BGBZ
-- CellParams::EXHALL_*
- removed derivatives member from SpatialCell
- support for total fluxes in diagnostic output is disabled
- output reducers for deleted variables are removed

PR #411, opened 2019-05-16
Precipitation dro merging dev

maximegrandin:
Added a data reduction operator to save the loss-cone population of ion species in each cell.
* The default number of energy bins is 16, and the minimum and maximum bin values are 100 eV and 100 keV. Energy bins are logarithmically spaced. Particles with energies under 100 eV (above 100 keV) are counted with the lowest (highest) energy bin.
* The loss-cone opening angle has a fixed value which is the same in the whole simulation domain. The value was set to 10 deg, which should be ok as a first approximation, since the phase-space density is averaged inside the loss cone (not integrated).
* At the moment, the loss-cone population is saved everywhere, including in regions where it is not relevant for particle precipitation.
* Differential ion number flux values are saved in units used by observers [ion/(cm2 s sr eV)], which is a deviation from the convention of manipulating SI units only.
* Magnetosphere_polar_small.cfg file was updated to include the setting of parameters related to this DRO (in [proton_precipitation]) and to write the precipitation differential fluxes in bulk files.
Suggested reviewers: @ykempf @markusbattarbee @ursg


datareduction/datareductionoperator.cpp
```diff
@@ -1436,6 +1436,136 @@ namespace DRO {
       return true;
    }
 
+   // Precipitation directional differential number flux
```
ykempf: I suggest that this comment be expanded (Doxygen style) to include a brief documentation of what's being done in this DRO.

ykempf: Apologies if I forgot to ask this in the PR by @markusbattarbee for the energy DRO. But it would be nice to have a couple of lines explaining the bins, angle and such.

datareduction/datareductionoperator.cpp
```diff
@@ -1436,6 +1436,136 @@ namespace DRO {
...
+   VariablePrecipitationDiffFlux::VariablePrecipitationDiffFlux(cuint _popID): DataReductionOperatorHasParameters(),popID(_popID) {
+      popName = getObjectWrapper().particleSpecies[popID].name;
+      cosAngle = cos(10.*M_PI/180.0); // cosine of fixed loss cone angle
```
ykempf: Should this not be a cfg parameter too? Or is 10° a Holy Value (TM) of the Ionospheric Community (TM)?

datareduction/datareductionoperator.cpp
```diff
@@ -1436,6 +1436,136 @@ namespace DRO {
...
+   }
+   
+   bool VariablePrecipitationDiffFlux::reduceData(const SpatialCell* cell,char* buffer) {
+
+      dataDiffFlux.assign(nChannels,0.0);
+
+      std::vector<Real> sumWeights(nChannels,0.0);
+
+      std::array<Real,3> B;
+      B[0] = cell->parameters[CellParams::PERBXVOL] +  cell->parameters[CellParams::BGBXVOL];
```
ykempf: Will this conflict with #405 or subsequent coding by @galfthan? In what order should we merge?

datareduction/datareductionoperator.cpp
```diff
@@ -1436,6 +1436,136 @@ namespace DRO {
...
+               int binNumber = round((log(energy) - log(emin)) / log(emax/emin) * (nChannels-1));
+               binNumber = max(binNumber,0); // anything < emin goes to the lowest channel
+               binNumber = min(binNumber,nChannels-1); // anything > emax goes to the highest channel
+
+	       thread_lossCone_sum[binNumber] += block_data[n * SIZE_VELBLOCK + cellIndex(i,j,k)] * countAndGate * normV*normV * DV3;
+	       thread_count[binNumber] += countAndGate * DV3;
+            }
+         }
+         for (int i=0; i<nChannels; i++){
+            thread_lossCone_sum[i] *= 1.0 / getObjectWrapper().particleSpecies[popID].mass * physicalconstants::CHARGE * 1.0e-4; // cm-2 s-1 sr-1 ev-1
```
ykempf: I suppose that this can be pulled out of the loop and parallel region?

datareduction/datareductionoperator.cpp
```diff
@@ -1484,13 +1614,7 @@ namespace DRO {
 	       if (ENERGY > E1limit) thread_E1_sum += block_data[n * SIZE_VELBLOCK+cellIndex(i,j,k)] * ENERGY * DV3;
 	       if (ENERGY > E2limit) thread_E2_sum += block_data[n * SIZE_VELBLOCK+cellIndex(i,j,k)] * ENERGY * DV3;
             }
-         }
```
ykempf: What do these changes have to do with the sauerkraut? That looks wrong to me, isn't it?

ykempf: (this line and the block below)

datareduction/datareductionoperator.cpp
maximegrandin: I was wondering about that too. I can certainly make it a cfg parameter with 10 deg as a default value. I chose this value based on empirical considerations :)

ykempf: Then I strongly suggest to avoid hard-coded values and make it also a parameter for the user.

datareduction/datareductionoperator.cpp
ursg: It will, since this should be changed to overall BVOL then. But I would actually suggest to merge this one first, because it's the smaller change

datareduction/datareductionoperator.cpp
markusbattarbee: And also then output as a .vlsv file parameter

ykempf: Good point!

datareduction/datareductionoperator.cpp
markusbattarbee: indeed - a merge mixup?

datareduction/datareductionoperator.cpp
markusbattarbee: indeed, after  dataDiffFlux[i] /= sumWeights[i];

datareduction/datareductionoperator.cpp
maximegrandin: Yes, l. 1489-1493 somehow disappeared in the merge with Markus's branch, and I am to blame for deleting l. 1487 to fix the resulting error...
I am adding those lines back.

maximegrandin: Yes, l. 1489-1493 somehow disappeared in the merge with Markus's branch, and I am to blame for deleting l. 1487 to fix the resulting error...
I am adding those lines back.

datareduction/datareductionoperator.cpp
maximegrandin: Yep, I moved those multiplications by constants to the suggested line.

datareduction/datareductionoperator.cpp
maximegrandin: I am adding doxygen-style documentation for both the precipitation and the energy density DROs

datareduction/datareductionoperator.cpp
maximegrandin: Thanks for bringing it up, I also add it as an output parameter in bulk files.

datareduction/datareductionoperator.cpp
maximegrandin: @markusbattarbee will make some updates in the energy density DRO tomorrow before this pull request can be merged

markusbattarbee:
Conflicting files: parameters.cpp
I split the output list to multi-line, might want to keep that formulation and pre-fix this conflict

datareduction/datareductionoperator.cpp
```diff
@@ -1436,6 +1436,153 @@ namespace DRO {
...
+      dataDiffFlux.assign(nChannels,0.0);
+
+      std::vector<Real> sumWeights(nChannels,0.0);
+
+      std::array<Real,3> B;
+      B[0] = cell->parameters[CellParams::PERBXVOL] +  cell->parameters[CellParams::BGBXVOL];
+      B[1] = cell->parameters[CellParams::PERBYVOL] +  cell->parameters[CellParams::BGBYVOL];
+      B[2] = cell->parameters[CellParams::PERBZVOL] +  cell->parameters[CellParams::BGBZVOL];
+
+      Real cosAngle = cos(lossConeAngle*M_PI/180.0); // cosine of fixed loss cone angle
```
ykempf: Je pinaille. But this could be the value computed and stored by the constructor. But never mind.

ykempf:
Looks good, up to the TBD for tomorrow. :) Congrats @maximegrandin!

datareduction/datareductionoperator.cpp
maximegrandin: Could be done, but since we are anyway writing lossConeAngle to bulk files, I think it can be left as is.

PR #413, opened 2019-05-16
Correct BVOL derivatives on DCCRG for Vlasov solver.

ykempf:
I fixed the bvol derivative issue. Any reviewer is welcome, this is a simple one I think. It is at least a temporary fix to have correct tests. We can make the definition and naming of the derivatives on fsgrid (differences, at the moment) and dccrg (proper derivatives with this merge) consistent later.


fieldsolver/gridGlue.cpp
```diff
@@ -276,13 +276,13 @@ void getFieldsFromFsGrid(
...
         sendBuffer[ii].sums[4 ] += volcell->at(fsgrids::volfields::EYVOL);
         sendBuffer[ii].sums[5 ] += volcell->at(fsgrids::volfields::EZVOL);
-        sendBuffer[ii].sums[6 ] += volcell->at(fsgrids::volfields::dPERBXVOLdy);
-        sendBuffer[ii].sums[7 ] += volcell->at(fsgrids::volfields::dPERBXVOLdz);
-        sendBuffer[ii].sums[8 ] += volcell->at(fsgrids::volfields::dPERBYVOLdx);
-        sendBuffer[ii].sums[9 ] += volcell->at(fsgrids::volfields::dPERBYVOLdz);
-        sendBuffer[ii].sums[10] += volcell->at(fsgrids::volfields::dPERBZVOLdx);
-        sendBuffer[ii].sums[11] += volcell->at(fsgrids::volfields::dPERBZVOLdy);
-        sendBuffer[ii].sums[12] += bgcell->at(fsgrids::bgbfield::BGBXVOL); 
+        sendBuffer[ii].sums[6 ] += volcell->at(fsgrids::volfields::dPERBXVOLdy) / technicalGrid.DY;
```
ursg: Hmm, so the fsgrid now contains finite difference deltas, and the DCCRG grid actually contains derivatives here, right?

Not fundamentally bad, but potentially confusing, that these two variables with the same name have different semantics.

ursg: Should this division be already be performed in the fieldsolver, when the values are written to fsgrid in the first place?

ykempf: Yes but that will introduce a zillion small changes to the field solver.

ykempf:
> Seems to make sense. Tested?

Would make sense.

fieldsolver/gridGlue.cpp
galfthan: That is a large change. If one wants to take care of it then instead rename derivative => difference or something. It is more of an naming problem. No sense to do it now for this fix

markusbattarbee: The variable name in FSgrid is indeed confusing. Would it be too much to rename them all over:
fsgrids::volfields::dPERBYVOLdz -> fsgrids::volfields::dzPERBYVOL etc?

ykempf: It might also clog the view when we debug this at some point, but it's less risky (only a sed away) than manually changing the divisions by D[XYZ] indeed.

ursg: Alright, this'll work just fine then.

We can consider *also* doing this division if we output this data though. But that's a separate construction site. :)

ykempf: This branch or with the sed suggested by @markusbattarbee?

ykempf:
I get 0 diff for proton rho, VX, VY and VZ in a Magnetosphere test, I'll leave this as is now because renaming all of fsgrid::dperb, dmoments, and some of bgbfield and volfields will be a bit much for now.

PR #416, opened 2019-05-21
Changed Vec type arrays that get allocated dynamically to std::vector…

tkoskela:
…<Vec,aligned_allocator<Vec,64>> types.

### DIFF
--- a/vlasovsolver/cpu_trans_map_amr.cpp
+++ b/vlasovsolver/cpu_trans_map_amr.cpp
@@ -521,8 +521,7 @@ void propagatePencil(Vec* dz, Vec* values, const uint dimension,
    const uint nTargetNeighborsPerPencil = 1;
 
    // Vector buffer where we write data, initialized to 0*/
-   std::vector<Vec, aligned_allocator<Vec,64>> targetValues;
-   targetValues.reserve((lengthOfPencil + 2 * nTargetNeighborsPerPencil) * WID3 / VECL);
+   std::vector<Vec, aligned_allocator<Vec,64>> targetValues((lengthOfPencil + 2 * nTargetNeighborsPerPencil) * WID3 / VECL);
    
    for (uint i = 0; i < (lengthOfPencil + 2 * nTargetNeighborsPerPencil) * WID3 / VECL; i++) {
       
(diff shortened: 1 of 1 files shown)


vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1231,7 +1236,7 @@ bool trans_map_1d_amr(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
                // Loop over cells in pencil +- 1 padded cell
                for ( uint celli = 0; celli < targetLength; ++celli ) {
                   
-                  Realv vector[VECL];
+                  Realf vector[VECL];
```
ykempf: Is this meant to have changed?

tkoskela: I noticed that I was using Realv datatypes in some places for vdf values, and I thought I'd fix that as well.

ykempf: ok

markusbattarbee:
Is this still WIP or good to test for fixing the Cray HSN problems?

tkoskela:
Good to test. Worked in my tests. Would like @ursg to check it out before merging.

markusbattarbee:
Ok, I'm running a small (12h, 64 nodes) 6D AMR test with this fix, my vector dipole and the new output metadata

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -520,7 +521,8 @@ void propagatePencil(Vec* dz, Vec* values, const uint dimension,
    const uint nTargetNeighborsPerPencil = 1;
 
    // Vector buffer where we write data, initialized to 0*/
-   Vec targetValues[(lengthOfPencil + 2 * nTargetNeighborsPerPencil) * WID3 / VECL];
+   std::vector<Vec, aligned_allocator<Vec,64>> targetValues;
+   targetValues.reserve((lengthOfPencil + 2 * nTargetNeighborsPerPencil) * WID3 / VECL);
```
ursg: These two lines can also be combined by simply giving the size as the constructor argument, right?
I'm kinda wondering if the compiler manages to optimize this away, or whether this will lead to one extraneous allocation here.

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1132,7 +1135,8 @@ bool trans_map_1d_amr(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
                computeSpatialSourceCellsForPencil(mpiGrid, pencils, pencili, dimension, sourceCells.data());
 
                // dz is the cell size in the direction of the pencil
-               Vec dz[sourceCells.size()];
+               std::vector<Vec, aligned_allocator<Vec,64>> dz;
+               dz.reserve(sourceLength);
```
ursg: Same goes for the vector here.

vlasovsolver/cpu_trans_map_amr.cpp
tkoskela: Sure, if you think that is safer. I was not sure how it treats the custom allocator when reallocating the vector.

PR #417, opened 2019-05-21
Added AMR support to Fluctuations.

ykempf:
Copied AMR function from Flowthrough as is.


tkoskela:
Let's move this to project.cpp

ykempf:
Like so? It compiles, I didn't run it yet, you could merge it with your branch and check again @tkoskela.

PR #418, opened 2019-05-21
Write mesh metadata, consolidate DROs

markusbattarbee:
- Adds mesh metadata for all variables.
 - Consolidates DROs to use lowercase variable names
 - Changes code to reference thermal / nonthermal portions of the distribution function instead of backstreaming / nonbackstreaming portions
 - Changed parameters.cpp to print both current and deprecated (yet still supported) output variable requests

still todo: parse input list for duplicate reducers?


ykempf:
Can the duplicate reducers be handled by the container?

markusbattarbee:
> Can the duplicate reducers be handled by the container?

the DRO container?
Hmm, that would require checking if a DRO with the same name already exists before creating one. That would be one way to do it though.

A dumb solution would be to have a flag for each DRO, set to "exists" or "doesn't exist". But this would be dumb and ugly.

An alternative would be to just generate them all, and at the end loop through them and remove duplicates. This would be reasonably easy to implement even with all the different types of DRO creation systems we already have.

markusbattarbee:
Also @ykempf, the changes to parameters.cpp mean the check cfg script would need to be updated to account for case insensitivity and accounting for regular and deprecated calls separately.

ykempf:
If we generate a `std::set<DRO>` instead of the vector used now and provide it with a comparison operator which would be the comparison of the string returned by `get_name`, it should work, shouldn't it? It's a bit of coding but there's not that many places where the vector of DROs is being tampered with (initial-grid IO off the top of my head) or used (initialisation and data reduction).

ykempf:
Sure, let's make sure that the checking script is up to par.

markusbattarbee:
Ok, yeah, that set solution sounds like it could work. However, that's a minor enhancement and not required for our current test/production goals, so I think it can be brought up as a separate issue.

markusbattarbee:
ok fixed the Hall field conflict and removed volE-fields.

PR #419, opened 2019-05-21
Fsgrid restarts

ursg:
This is the current work-in-progress state of fsgrid restart code. It doesn't crash, but there must still be an offset error in the reading code, since the data arrives scrambled.


iowrite.cpp
```diff
@@ -1291,6 +1300,108 @@ bool writeRestart(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
    restartReducer.addOperator(new DRO::MPIrank);
    restartReducer.addOperator(new DRO::BoundaryType);
    restartReducer.addOperator(new DRO::BoundaryLayer);
+
+   // Fsgrid Reducers
+   restartReducer.addOperator(new DRO::DataReductionOperatorFsGrid("fg_EFIELD",[](
```
ykempf: This is not needed I think.

ioread.cpp
```diff
@@ -778,6 +777,78 @@ bool readCellParamsVariable(
...
+      for(int task = 0; task < myRank; task++) {
+         std::array<int32_t,3> thatTasksSize;
+         thatTasksSize[0] = targetGrid.calcLocalSize(globalSize[0], decomposition[0], task%decomposition[0]);
+         thatTasksSize[1] = targetGrid.calcLocalSize(globalSize[1], decomposition[1], (task/decomposition[0])%decomposition[1]);
+         thatTasksSize[2] = targetGrid.calcLocalSize(globalSize[2], decomposition[2], (task/decomposition[0])/decomposition[1]);
+         localStartOffset += thatTasksSize[0] * thatTasksSize[1] * thatTasksSize[2];
+      }
+     
+      // Read into buffer
+      std::vector<double> buffer(storageSize*N);
```
ykempf: Should this and 839 not be Real?

iowrite.cpp
```diff
@@ -1291,6 +1300,108 @@ bool writeRestart(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
...
+                      FsGrid< std::array<Real, fsgrids::bfield::N_BFIELD>, 2>& perBGrid,
+                      FsGrid< std::array<Real, fsgrids::efield::N_EFIELD>, 2>& EGrid,
+                      FsGrid< std::array<Real, fsgrids::ehall::N_EHALL>, 2>& EHallGrid,
+                      FsGrid< std::array<Real, fsgrids::egradpe::N_EGRADPE>, 2>& EGradPeGrid,
+                      FsGrid< std::array<Real, fsgrids::moments::N_MOMENTS>, 2>& momentsGrid,
+                      FsGrid< std::array<Real, fsgrids::dperb::N_DPERB>, 2>& dPerBGrid,
+                      FsGrid< std::array<Real, fsgrids::dmoments::N_DMOMENTS>, 2>& dMomentsGrid,
+                      FsGrid< std::array<Real, fsgrids::bgbfield::N_BGB>, 2>& BgBGrid,
+                      FsGrid< std::array<Real, fsgrids::volfields::N_VOL>, 2>& volGrid,
+                      FsGrid< fsgrids::technical, 2>& technicalGrid)->std::vector<double> {
```
ykempf: Ah, I see there is double here, why should this not be Real?

iowrite.cpp
```diff
@@ -1291,6 +1300,108 @@ bool writeRestart(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
...
+                  for(int x=0; x<gridSize[0]; x++) {
+                     std::memcpy(&retval[index], perBGrid.get(x,y,z), sizeof(double)*fsgrids::bfield::N_BFIELD);
+                     index += fsgrids::bfield::N_BFIELD;
+                  }
+              }
+            }
+            return retval;
+         }
+   ));
+   restartReducer.addOperator(new DRO::DataReductionOperatorFsGrid("fg_BGB",[](
```
ykempf: So far this is being recomputed at restart.

iowrite.cpp
```diff
@@ -1291,6 +1300,108 @@ bool writeRestart(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
...
+                  for(int x=0; x<gridSize[0]; x++) {
+                     std::memcpy(&retval[index], BgBGrid.get(x,y,z), sizeof(double)*fsgrids::bgbfield::N_BGB);
+                     index += fsgrids::bgbfield::N_BGB;
+                  }
+              }
+            }
+            return retval;
+         }
+   ));
+   restartReducer.addOperator(new DRO::DataReductionOperatorFsGrid("fg_DPERB",[](
```
ykempf: And I think that this is not needed either.

ykempf:
So as you see from my comments, my biggest question at the moment is why you force to write in double when the fsgrids use Real (which typically is not double).

ursg:
Testpackage says this doesn't break anything, *and* the restart_write and restart_read tests sorta work with it.

Even though restarts still have diffs, I'm merging this into dev already, as it is a clear improvement over the current state there (in which restarts don't work at all).

PR #423, opened 2019-06-05
Fsgrid restarts

ykempf:
This is the current state of my restart debugging, there's mostly dead code cleanup and two minor bug fixes.


ykempf:
The DT2 is not changing anything of my recent tests but is likely to be better this way, I just stumbled upon it by chance. The faceneighbourranks thing is silent because it doesn't make a difference in non-amr if it's being run. It just saves a bit of computation.

ursg:
Well then it's an even more trivial change. Good to go!

PR #424, opened 2019-06-05
Fsgrid restarts with different number of tasks.

ursg:
This appears to compile and run now, but isn't thoroughly tested yet.

Especially, I'm not sure how good performance will be at scale, since it potentially includes lots of small reads. Will need to check whether multiread and/or bigger reads and recommunication is better here.


ursg:
I think this is good enough to go to dev now. We'll need to see how well the performance ends up being, but that can only be tested with... a bigger run!

PR #425, opened 2019-06-06
Fsgrid restarts

ykempf:
My latest 2 commits, some cleanup essentially.


tkoskela:
Uhh, was referring to initializeGrids in grid.cpp, thought my comment would get attached to the file changes but apparently not.

tkoskela:
Otherwise this looks fine.

PR #426, opened 2019-06-17
placed perturbed field initialization within restart check

markusbattarbee:


markusbattarbee:
A full dipole in BGB, yes.
PERB has the difference between a full dipole and an X-vanishing dipole.

Additionally, either BGB will have const IMF, or PERB will have inflow IMF which scales in with X.

PR #427, opened 2019-06-18
Fix outflow and ionosphere boundaries to get symmetric B evaluation.

ursg:
Before, B vectors were always copied in the sysboundary cells, which
caused the Yee-Lattice locations of these components to be asymmetric
wrt. the boundary. Now, it is checking whether the neighbour on the
corresponding components' side is a non-sysboundary and solving the
field normally if it is.

This should increase inner boundary accuracy. Hopefully. Testing in
progress.

### DIFF
--- a/sysboundary/ionosphere.cpp
+++ b/sysboundary/ionosphere.cpp
@@ -576,6 +576,7 @@ namespace SBC {
                   propagateMagneticField(perBGrid, perBDt2Grid, EGrid, EDt2Grid, i, j, k, dt, RKCase, false, false, true);
                   break;
                default:
+                  cerr << "ERROR: ionosphere boundary tried to propagate nonsensical magnetic field component " << component << endl;
                   break;
             }
             return bGrid->get(i,j,k)->at(fsgrids::bfield::PERBX + component);
--- a/sysboundary/ionosphere.cpp
+++ b/sysboundary/ionosphere.cpp
@@ -33,6 +33,7 @@
 #include "../vlasovmover.h"
 #include "../fieldsolver/fs_common.h"
 #include "../fieldsolver/fs_limiters.h"
+#include "../fieldsolver/ldz_magnetic_field.hpp"
 #include "../common.h"
 #include "../object_wrapper.h"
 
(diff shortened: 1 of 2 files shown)


ykempf:
And for reference the outflow logic was translated with a more elegant fsgrid formulation. That was identical as expected.

ykempf:
And for reference the outflow logic was translated with a more elegant fsgrid formulation. That was identical as expected.

sysboundary/ionosphere.cpp
```diff
@@ -33,6 +33,7 @@
 #include "../vlasovmover.h"
 #include "../fieldsolver/fs_common.h"
 #include "../fieldsolver/fs_limiters.h"
+#include "../fieldsolver/ldz_magnetic_field.hpp"
```
ykempf: Why did this change?

sysboundary/outflow.cpp
```diff
@@ -359,54 +359,40 @@ namespace SBC {
       determineFace(&isThisCellOnAFace[0], x, y, z, dx, dy, dz, true);
       
       cuint sysBoundaryLayer = technicalGrid.get(i,j,k)->sysBoundaryLayer;
+      if(sysBoundaryLayer == 1) {
+         cint neigh_i=i + ((component==0)?-1:0);
+         cint neigh_j=j + ((component==1)?-1:0);
+         cint neigh_k=k + ((component==2)?-1:0);
+         cuint neighborSysBoundaryFlag = technicalGrid.get(neigh_i, neigh_j, neigh_k)->sysBoundaryFlag;
```
markusbattarbee: Do we really want to ignore these flags?
this->facesToProcess[i]
!this->facesToSkipFields[i]

sysboundary/outflow.cpp
```diff
@@ -359,54 +359,40 @@ namespace SBC {
...
+            }
+            if(RKCase == RK_ORDER1 || RKCase == RK_ORDER2_STEP2) {
+               fieldValue = perBGrid.get(i,j,k)->at(fsgrids::bfield::PERBX + component);
+            } else {
+               fieldValue = perBDt2Grid.get(i,j,k)->at(fsgrids::bfield::PERBX + component);
+            }
+            return fieldValue;
          }
+      }
+
```
markusbattarbee: This is an implicit else for the neighbour sysboundarylayer check

sysboundary/outflow.cpp
```diff
@@ -359,54 +359,40 @@ namespace SBC {
...
+      if(RKCase == RK_ORDER1 || RKCase == RK_ORDER2_STEP2) {
+         fieldValue = fieldBoundaryCopyFromExistingFaceNbrMagneticField(perBGrid, technicalGrid, i, j, k, component);
       } else {
```
markusbattarbee: Instead of searching for the nearest neighbor, should we just choose the cell directly now that we know it's a sysboundarycell and we know the i,j,k to it and the component?

sysboundary/ionosphere.cpp
```diff
@@ -554,6 +555,34 @@ namespace SBC {
...
+                  propagateMagneticField(perBGrid, perBDt2Grid, EGrid, EDt2Grid, i, j, k, dt, RKCase, true, false, false);
+                  break;
+               case 1:
+                  propagateMagneticField(perBGrid, perBDt2Grid, EGrid, EDt2Grid, i, j, k, dt, RKCase, false, true, false);
+                  break;
+               case 2:
+                  propagateMagneticField(perBGrid, perBDt2Grid, EGrid, EDt2Grid, i, j, k, dt, RKCase, false, false, true);
+                  break;
+               default:
+                  break;
```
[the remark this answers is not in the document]
markusbattarbee: And why no check for RK1 RK2 stepping?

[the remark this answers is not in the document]
ykempf: That is passed to the solver, @markusbattarbee.

sysboundary/outflow.cpp
ykempf: The first I guess yes, the other should be integrated in the else part I think.

sysboundary/outflow.cpp
ykempf: Maybe some more commenting then?

sysboundary/outflow.cpp
ykempf: But we don't know where that nearest neighbour is (corner cases).

markusbattarbee: Ah, right we want to copy from the nearest neighbor *non*-sysboundarycell?

ykempf: As the code and documentation clearly state. Not.
https://github.com/fmihpc/vlasiator/blob/ea5a3345f45ee542432e10d498bf8a9973fb65b6/sysboundary/sysboundarycondition.cpp#L798

sysboundary/ionosphere.cpp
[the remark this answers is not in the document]
markusbattarbee: That's not the issue, but looking a bit upwards it was cleared. This and the outflow deal with the RK1/2 stepping differently, but both do it. These could be consolidated, though, and tbh I think the outflow way of writing it is clearer (lines 382-387).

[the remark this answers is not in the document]
ykempf: Sure, could be made more uniform. :)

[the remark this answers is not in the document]
ykempf: @ursg will love us. :)

sysboundary/ionosphere.cpp
ursg: I't needs this to find the definition of propagateMagneticField.

PR #428, opened 2019-06-19
Level 3 refinement for Magnetosphere, innermost sphere.

ykempf:
Apologies for the white space (mask it on GitHub) but tabs are a different length in my editor...


ykempf:
Nope, that's buggy.

ykempf:
Should we make this recursive right away? ;-)

markusbattarbee:
It'll work now assuming the refinement radius parameters are correct, but those bw bw2 and bw3 parameters need to be calculated better so that we don't start to refine boundary cells by accident. And I guess one could make it recursive, just need to manually add the config params.

ykempf:
I wasn't really serious about the recursive thing...

There is a check for ionosphere < L3 < L2, do you want more? Or what do you mean exactly?

PR #429, opened 2019-06-26
Amr fsaved stride

markusbattarbee:
Fixed stride calculation to work correctly at higher AMR levels.

### DIFF
--- a/iowrite.cpp
+++ b/iowrite.cpp
@@ -936,16 +936,19 @@ bool writeVelocitySpace(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpi
          // Determine cellID's 3D indices
 	 
 	 // Loop over AMR levels
+	 uint startindex=1;
+	 uint endindex=1;
 	 for (uint AMR = 0; AMR <= P::amrMaxSpatialRefLevel; AMR++) {
-	    uint AMRP = AMR+1;
-	    uint startindex = (AMR*P::xcells_ini)*(AMR*P::ycells_ini)*(AMR*P::zcells_ini)+1;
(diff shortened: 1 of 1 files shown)


markusbattarbee:
For now, I went with a simple cartesian formulation, extending the previous method to higher AMR levels. The old version had funny diagonal sets of cells selected.

Old version:

old_1
old_2
old_3

New version:
new_1
new_2
new_3

iowrite.cpp
```diff
@@ -934,28 +934,42 @@ bool writeVelocitySpace(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpi
...
+	    startindex = endindex;
+	    endindex = endindex + cellsthislevel;
+	  
+	    // If cell belongs to this AMR level, find indices
+	    if ((cells[i]>=startindex)&&(cells[i]<endindex)) {
+	       lineX =  (cells[i]-startindex) % (AMRm*P::xcells_ini);
+	       lineY = ((cells[i]-startindex) / (AMRm*P::xcells_ini)) % (AMRm*P::ycells_ini);
+	       lineZ = ((cells[i]-startindex) /((AMRm*P::xcells_ini) *  (AMRm*P::ycells_ini))) % (AMRm*P::zcells_ini);
+	       // Check that indices are in correct intersection at least in one plane
+	       if ((P::systemWriteDistributionWriteXlineStride[index] > 0 &&
```
ykempf: It's all boolean and, so you can remove the redundant lines. Maybe put the current whole set in a comment to make it clear, or add a short comment? Actually there is a comment but it's not quite what it does. Since it's all and this checks that it is an intersection in all 3 planes, which we want.

markusbattarbee: Actually, if we want to be able to run 2D runs with this code version, we still need these lines. And I think doing high-res 2D-AMR could be interesting in the future.

PR #430, opened 2019-07-02
Magnetosphere l3 refine

markusbattarbee:
Compiles, about to test


ursg:
This has been tested for the EGE run setup and seems to work well:

EGA0001

PR #433, opened 2019-07-08
Cpu translate nonuniform numerics

markusbattarbee:
I think calculating nonuniform translation coefficients would be better done with cell widths close to 1. 

@ykempf could you perhaps try this commit in the flowthrough tests where we had order 1e-3 diffs at refinement boundaries, to see if this helps things? cheers.


markusbattarbee:
update: didn't see meaningful improvement in my initial tests, though I think I did them without testpackage flags.

PR #434, opened 2019-07-12
Warnings reduction

ursg:
Since they've been nagging me for quite a while now: This pull request removes a whole lot of compilation warnings.
Mostly in the form of "Warning: unused parameter xyz" (fixed by flagging them as ```__attribute__((unused))```), and a bunch of "signed-unsigned comparisons" (fixed by changing the datatype of the loop counter, in cases where it will not cause integer promotion overflow problems).

This *also* fixes the handwritten warnings in project initialization, where per-population overall sparsity limits were employed instead of per-cell/per-population limits, and still requires testing that it didn't break anything.

Also, there are still a bunch of warnings left to fix.

### DIFF
--- a/amr_refinement_criteria.cpp
+++ b/amr_refinement_criteria.cpp
@@ -38,7 +38,7 @@ namespace amr_ref_criteria {
    
    Base* relDiffMaker() {return new RelativeDifference;}
    
-   void Base::evaluate(const Realf* velBlost,Realf* result,const uint popID) {
+   void Base::evaluate(const Realf* velBlost __attribute__((unused)),Realf* result,const uint popID __attribute__((unused))) {
       for (uint i=0; i<WID3; ++i) result[i] = 0.0;
    }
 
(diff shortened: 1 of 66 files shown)


ursg:
This branch now builds on my fridge, showing only hand-written ```#warning``` statements anymore.

ursg:
Ran the testpackage, and it does indeed have differences in it. Damnit. Namely in Magnetosphere_polar_small:
```
  variable     |     absolute diff     |     relative diff | 
proton/vg_rho_0                7164.79                 0.00462734    
proton/vg_v_0                11890                 0.015814    
proton/vg_v_1                163.288                 0.222152    
proton/vg_v_2                6727.73                 0.160726    
```

I'll try to bisect through my commits here which of them is causing it, and then report back.

ursg:
This branch has, unfortunately diverged too much from the dev branch to be of any use anymore. This will need a new attempt.

PR #437, opened 2019-08-30
Flux limiter numerics improvement

markusbattarbee:
Scales distribution function values closer to 1 before evaluating the flux limiter. Scaling factor to use is the per-population minimum value. Testing was performed with an acceleration test for electrons, for both PLM, PPM, and PQM. Results were greatly improved.

Mini-app compiles but results were not analyzed yet.


markusbattarbee:
hold it...  :) of course forgot to update the amr sections as was testing with the electron branch.

markusbattarbee:
Testpackage says that bulk values are conserved well for all tests. In the few cases where the relative error is significant, it's for a value (e.g. bulk v_z) which is practically zero.

I'll look at more at VDF diffs before confirming this pull.

markusbattarbee:
In my opinion, all tests look good, and this can be merged. It shouldn't as such change ion dynamics to a significant extent, but based on the improvement to electrons, is worth including.

[log_Freezer4.txt](https://github.com/fmihpc/vlasiator/files/3582906/log_Freezer4.txt)

markusbattarbee:
To add, I also looked at VDF plots from polar_magnetosphere_small and they all look good to me.

mini-apps/acceleration-vlasiator/map_test.cpp
```diff
@@ -5,6 +5,10 @@
 #include "vlasovsolver/cpu_1d_ppm.hpp"
 #include "vlasovsolver/cpu_1d_plm.hpp"
 
+#define cell_pop_threshold 1.e-15
+// In vlasiator, called via spatial_cell->getVelocityBlockMinValue(popID)
```
ursg: This comment is confusing, because it seems to imply that this is a parameter that can be changed at runtime, or by calling that function?

Maybe it should rather be something like "The slope limiters are numerically inaccurate for low values, thus the phase space densities are scaled by this divisor".

(This also makes me wonder, if it makes any sense to still call it "threshold" here, since it's really only used for scaling)

mini-apps/acceleration-vlasiator/map_test.cpp
```diff
@@ -5,6 +5,10 @@
 #include "vlasovsolver/cpu_1d_ppm.hpp"
 #include "vlasovsolver/cpu_1d_plm.hpp"
 
+#define cell_pop_threshold 1.e-15
```
ursg: C++ style hint:
```
const int cell_pop_threshold=1.e-15;
```
gives the same performance, and is type-safe

mini-apps/amr_reconstruction/map_test.cpp
```diff
@@ -10,6 +10,9 @@
 #include <boost/random/variate_generator.hpp>
 #include <iostream>
 
+#define cell_pop_threshold 1.e-15
+// In vlasiator, called via spatial_cell->getVelocityBlockMinValue(popID)
```
ursg: Same comment as above

vlasovsolver/cpu_1d_plm.hpp
```diff
@@ -38,8 +38,10 @@ t=(v-v_{i-0.5})/dv where v_{i-0.5} is the left face of a cell
 The factor 2.0 is in the polynom to ease integration, then integral is a[0]*t + a[1]*t**2
 */
 
-inline void compute_plm_coeff(const Vec * const values, uint k, Vec a[2]){
-  const Vec d_cv=slope_limiter(values[k - 1], values[k], values[k + 1]);
+inline void compute_plm_coeff(const Vec * const values, uint k, Vec a[2], const Realv threshold){
+   // scale values closer to 1 for more accurate slope limiter calculation
+  const Realv scale = 1./threshold;
+  const Vec d_cv=slope_limiter(values[k - 1]*scale, values[k]*scale, values[k + 1]*scale)*threshold;
```
ursg: This is the point where it really doesn't make a lot of sense to call it "threshould", IMHO. Maybe just make the scale a constant, instead?

markusbattarbee: It could be a constant yes, but if someone decides to run Vlasiator in unforeseen physical configurations, this could then cause errors to crop. The threshold value is inherently tied to the numerical scale of the VDF, and will always be set by the user, which is why I chose it.

If it's really seen as necessary, it can of course be made into an additional config file parameter with a default value of e.g. 1.e-15.

markusbattarbee:
Good now?

ursg:
Yeps. Tested, too. Let's take it!

PR #438, opened 2019-09-02
Puhti makefile and small fixes for new vectorclass version.

ursg:
This compiles, but hasn't managed to run a testpackage yet.


ursg:
This version now manages to run the testpackage with the intel compiler. No performance optimization / flag mangling done yet.

PR #439, opened 2019-09-05
Fix dev testpackage

markusbattarbee:
Added two helper python scripts for updating config files. Updated all config files for both testpackage and projects. Added extra (and missing) catches for output variable matching when reading a config file. Changed phiprof file to compare.

Diagnostic output is still oldschool, could do that as well at some point.


ursg:
This works great on puhti. No reason to keep the old (borderline disfunctional) state of the testpackage around in dev. So I'm merging this.

PR #441, opened 2019-10-03
Fixed taito makefiles... again.

ursg:
This makes the particle-post-pusher buildable (and hopefully runnable) on taito again, with the openmpi 1.10.0 module loaded.


markusbattarbee:
Seems to work - at least it ran for 25 minutes on a test queue before I killed it :)

markusbattarbee:
Taito no longer exists.

PR #442, opened 2019-10-10
Set Hall E to zero for layers 2 and 3. If it works, the layer limit w…

ykempf:
If it works, the layer limit will be set as a cfg parameter. Open for discussion, also while it is in the queue.


ykempf:
Indeed there needs to be more layers numbered.

ykempf:
Closing, this didn't help and Kostis modified this layering to suit his PML purposes.

PR #448, opened 2019-10-18
Restructured OpenACC acceleration solver to parallelize over columns.

ursg:
It's faster! And it doesn't seem to be all wrong (at first glance).

### DIFF
--- a/vlasovsolver/cpu_acc_map.cpp
+++ b/vlasovsolver/cpu_acc_map.cpp
@@ -36,19 +36,19 @@
 using namespace std;
 using namespace spatial_cell;
 /** Attempt to add the given velocity block to the given velocity mesh.
- * If the block was added to the mesh, its data is set to zero values and 
+ * If the block was added to the mesh, its data is set to zero values and
  * velocity block parameters are calculated.
  * @param blockGID Global ID of the added velocity block.
  * @param vmesh Velocity mesh where the block is added.
  * @param blockContainer Velocity block data container.
- * @return Local ID of the added block. If the block was not added, the 
+ * @return Local ID of the added block. If the block was not added, the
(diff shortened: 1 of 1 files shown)


ursg:
Actually, under closer inspection, this does still seem to have at least one bug. After acceleration, everything seems to be off by one velocity block in some direction (here are pictures of the first three timesteps):




I'll go through the old and new version of the code and check what I have messed up here.

ykempf:
Let's make a simple test like the original test_acc. That'll be more informative than Magnetosphere.

On 20 October 2019 16:03:54 EEST, Urs Ganse <notifications@github.com> wrote:
>Actually, under closer inspection, this does still seem to have at
>least one bug. After acceleration, everything seems to be off by one
>velocity block in some direction (here are pictures of the first three
>timesteps):
>
>
>
>
>I'll go through the old and new version of the code and check what I
>have messed up here.
>
>-- 
>You are receiving this because you are subscribed to this thread.
>https://github.com/fmihpc/vlasiator/pull/448#issuecomment-544250373

-- 

ursg:
Alright, I ran with a acctest_1, where I disabled all acceleration directions apart from x, and see what it does:







So this looks to me like it's having maybe two distinct problems? One being that old velocity space memory contents don't get cleared and just linger around and the other one being that *some* columns don't move the way they should.

ykempf:
T0 is already looking weird to me. Is there no test_acc anymore?

ursg:
Do you mean projects/unsupported/test_acc?

ykempf:
Apparently I do. :)

ursg:
Ok, the last commit seems to have fixed it, the columns are no longer swooshing away in velocity space:

PR #450, opened 2019-10-21
Changed method of RNG in projects to be explicitly threadsafe and to …

markusbattarbee:
…work on PGI compilers. Removed Boost use for lowercase strings. Test run is in the queue.


ursg:
Yup, this seems to work as intented

PR #452, opened 2019-10-22
Make the acceleration solver run async in multiple queues.

ursg:
This required a bit of loop reordering, as cells are now processed in parallel
by multiple queues on the GPU.

I suppose it would make sense to have the number of queues as a config parameter.

### DIFF
--- a/vlasovsolver/cpu_acc_map.cpp
+++ b/vlasovsolver/cpu_acc_map.cpp
@@ -132,7 +132,7 @@ bool map_1d(SpatialCell* spatial_cell,
    if(vmesh.size() == 0 )
       return true;
 
-   const int NUM_ASYNC_QUEUES=8;
+   const int NUM_ASYNC_QUEUES=P::openaccQueueNum;
    const int openacc_async_queue_id = (int)(spatial_cell->parameters[CellParams::CELLID]) % NUM_ASYNC_QUEUES;
 
    // Velocity grid refinement level, has no effect but is
--- a/vlasovsolver/cpu_acc_map.cpp
+++ b/vlasovsolver/cpu_acc_map.cpp
@@ -132,6 +132,8 @@ bool map_1d(SpatialCell* spatial_cell,
    if(vmesh.size() == 0 )
       return true;
 
+   const int NUM_ASYNC_QUEUES=8;
+   const int openacc_async_queue_id = (int)(spatial_cell->parameters[CellParams::CELLID]) % NUM_ASYNC_QUEUES;
 
    // Velocity grid refinement level, has no effect but is
    // needed in some vmesh::VelocityMesh function calls.
@@ -200,6 +202,9 @@ bool map_1d(SpatialCell* spatial_cell,
    std::vector<uint> setColumnOffsets;
    std::vector<uint> setNumColumns;
 
+   // Make sure any previous velocity space updates have finished for this cell
+   #pragma acc wait(openacc_async_queue_id)
+
    sortBlocklistByDimension(vmesh, dimension, blocks,
                             columnBlockOffsets, columnNumBlocks,
                             setColumnOffsets, setNumColumns);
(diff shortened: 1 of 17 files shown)


ykempf:
Yes that sounds like an optimisation parameter to be throughly investigated as well.

markusbattarbee:
Shouldn't the async queues count be equal to num_threads for optimal performance?

ursg:
> Shouldn't the async queues count be equal to num_threads for optimal performance?

No, I think those should be independent. Even when running just on a single thread (my tests so far), 8 queues seem like a good number, as the cpu thread dispatches velocity cells to the gpu and then continues working on the next one. It might even be that something like 8*num_threads turns out to be a good number, or that the queue allocation schedule needs to be thread aware there.

ursg:
This is what the kernel execution now looks like in nsight:
acc

markusbattarbee:
That looks good?

vlasovsolver/cpu_acc_map.cpp
```diff
@@ -132,6 +132,8 @@ bool map_1d(SpatialCell* spatial_cell,
    if(vmesh.size() == 0 )
       return true;
 
+   const int NUM_ASYNC_QUEUES=P::openaccQueueNum;
+   const int openacc_async_queue_id = (int)(spatial_cell->parameters[CellParams::CELLID]) % NUM_ASYNC_QUEUES;
```
markusbattarbee: Hm, there's no guarantee that our cellIDs would be continuous per single task, so this could end up with weird slowdowns...

ursg: And indeed, it does. The nsight graph shows occasional syncpoints, where subsequent cells are submitted to the same queue.

I wonder though: now that the calls to this function are wrapped in a loop over cells, do we even *need* the wait call inside map_1d anymore? Or can we just happily stuff more work into the same async queue to be evaluated at some later time? The overall wait statement at the end of each acceleration order should take care of synchronization then.

markusbattarbee: sounds reasonable

ursg: A quick test shows that the wait statement here can be tossed right out, without any negative effects. It seems faster too, but I'd want to see that quantitatively checked out before I commit anything.

vlasovsolver/vlasovmover.cpp
```diff
@@ -272,47 +272,52 @@ void calculateAcceleration(const uint popID,const uint globalMaxSubcycles,const
...
+            subcycleDt = max(dt - step * maxVdt, 0.0);
+         } else{
+            subcycleDt = maxVdt;
+         }
+
+         //generate pseudo-random order which is always the same irrespective of parallelization, restarts, etc.
+         char rngStateBuffer[256];
+         random_data rngDataBuffer;
+
+         // set seed, initialise generator and get value. The order is the same
```
markusbattarbee: This randomization, which provides just a single integer (same for all cells but depends randomly on timestep) could be outside the order_step loop. No point in re-randomizing the same number over and over again.

vlasovsolver/cpu_acc_map.cpp
markusbattarbee: Yep, I agree that if we have a pragma acc wait at each dimension order loop, we can lose the wait within map_1d. In this case, we're using the async queues in such a manner that we could just use the CellID as the queue ID. This could be tested by setting an arbitrarily large value to P::openaccQueueNum.

vlasovsolver/vlasovmover.cpp
ursg: I was wondering the same, but this random number generator is so bloody fast that it probably isn't even worth the effort moving it outside the loop. :)

ursg:
Merging this already into openacc, since it seems to work quite well

PR #453, opened 2019-10-24
Fix to vlsvextract because of erroneous loop over meshes.

ykempf:


ykempf:
Wait

ykempf:
Nyt

ursg:
I suppose that breaks the old vAMR stuff that Arto wrote a long time ago, but that's ok, since there's no intention to use that anyway. So by and large, I'd say it looks good.

markusbattarbee:
Looks ok to me too.

PR #455, opened 2019-11-13
Write volume B field into restarts for post-processing.

ykempf:
Do we want this or a dedicated DRO doing the sum of background and perturbed?

### DIFF
--- a/datareduction/datareducer.cpp
+++ b/datareduction/datareducer.cpp
@@ -634,7 +634,7 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
                   for(int z=0; z<gridSize[2]; z++) {
                      for(int y=0; y<gridSize[1]; y++) {
                         for(int x=0; x<gridSize[0]; x++) {
-                           retval[gridSize[1]*gridSize[0]*z + gridSize[0]*y + x] =  + (*EHallGrid.get(x,y,z))[index];
+                           retval[gridSize[1]*gridSize[0]*z + gridSize[0]*y + x] = (*EHallGrid.get(x,y,z))[index];
                         }
                      }
                   }
--- a/datareduction/datareducer.cpp
+++ b/datareduction/datareducer.cpp
@@ -574,6 +574,44 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
          }
          continue;
       }
+      if(lowercase == "vole" || lowercase == "vg_vole" || lowercase == "evol" || lowercase == "vg_e_vol" || lowercase == "e_vol") {
+         // Volume-averaged E field
+         outputReducer->addOperator(new DRO::DataReductionOperatorCellParams("vg_e_vol",CellParams::EXVOL,3));
+	 outputReducer->addMetadata(outputReducer->size()-1,"V/m","$\\mathrm{V}\\,\\mathrm{m}^{-1}$","$E_\\mathrm{vol,vg}$","1.0");
+         continue;
+      }
+      if(lowercase == "fg_vole" || lowercase == "fg_e_vol" || lowercase == "fg_evol") {
+         outputReducer->addOperator(new DRO::DataReductionOperatorFsGrid("fg_e_vol",[](
+                      FsGrid< std::array<Real, fsgrids::bfield::N_BFIELD>, 2>& perBGrid,
+                      FsGrid< std::array<Real, fsgrids::efield::N_EFIELD>, 2>& EGrid,
+                      FsGrid< std::array<Real, fsgrids::ehall::N_EHALL>, 2>& EHallGrid,
+                      FsGrid< std::array<Real, fsgrids::egradpe::N_EGRADPE>, 2>& EGradPeGrid,
+                      FsGrid< std::array<Real, fsgrids::moments::N_MOMENTS>, 2>& momentsGrid,
+                      FsGrid< std::array<Real, fsgrids::dperb::N_DPERB>, 2>& dPerBGrid,
+                      FsGrid< std::array<Real, fsgrids::dmoments::N_DMOMENTS>, 2>& dMomentsGrid,
+                      FsGrid< std::array<Real, fsgrids::bgbfield::N_BGB>, 2>& BgBGrid,
+                      FsGrid< std::array<Real, fsgrids::volfields::N_VOL>, 2>& volGrid,
+                      FsGrid< fsgrids::technical, 2>& technicalGrid)->std::vector<double> {
+
+               std::array<int32_t,3>& gridSize = technicalGrid.getLocalSize();
+               std::vector<double> retval(gridSize[0]*gridSize[1]*gridSize[2]*3);
+
+               // Iterate through fsgrid cells and extract EVOL
+               for(int z=0; z<gridSize[2]; z++) {
+                  for(int y=0; y<gridSize[1]; y++) {
+                     for(int x=0; x<gridSize[0]; x++) {
+                        retval[3*(gridSize[1]*gridSize[0]*z + gridSize[0]*y + x)] =  + (*volGrid.get(x,y,z))[fsgrids::EXVOL];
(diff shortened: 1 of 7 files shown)


iowrite.cpp
```diff
@@ -1321,6 +1321,10 @@ bool writeRestart(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
    restartReducer.addOperator(new DRO::DataReductionOperatorCellParams("max_v_dt",CellParams::MAXVDT,1));
    restartReducer.addOperator(new DRO::DataReductionOperatorCellParams("max_r_dt",CellParams::MAXRDT,1));
    restartReducer.addOperator(new DRO::DataReductionOperatorCellParams("max_fields_dt",CellParams::MAXFDT,1));
+   restartReducer.addOperator(new DRO::DataReductionOperatorCellParams("vg_b_background_vol",CellParams::BGBXVOL,3));
+   restartReducer.addMetadata(restartReducer.size()-1,"T","$\\mathrm{T}$","$B_\\mathrm{vol,vg,bg}$","1.0");
+   restartReducer.addOperator(new DRO::DataReductionOperatorCellParams("vg_b_perturbed_vol",CellParams::PERBXVOL,3));
+   restartReducer.addMetadata(restartReducer.size()-1,"T","$\\mathrm{T}$","$B_\\mathrm{vol,vg,per}$","1.0");
    restartReducer.addOperator(new DRO::DataReductionOperatorBVOLDerivatives("Bvolume_derivatives",0,bvolderivatives::N_BVOL_DERIVATIVES));
```
ykempf: Should this be kept or dropped?

ykempf: This question remains.

markusbattarbee: does VisIt use it for anything?

ykempf: No, I mean, we could plot them with VisIt but I don't see why they are written here if they are not read back in. Some remnant of something, unless this is also needed for some postprocessing.

ursg: I don't think it's needed for postprocessing. The particle pusher certainly doesn't read it.

iowrite.cpp
```diff
@@ -1321,6 +1321,8 @@ bool writeRestart(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
    restartReducer.addOperator(new DRO::DataReductionOperatorCellParams("max_v_dt",CellParams::MAXVDT,1));
    restartReducer.addOperator(new DRO::DataReductionOperatorCellParams("max_r_dt",CellParams::MAXRDT,1));
    restartReducer.addOperator(new DRO::DataReductionOperatorCellParams("max_fields_dt",CellParams::MAXFDT,1));
+   restartReducer.addOperator(new DRO::VariableBVol);
+   restartReducer.addMetadata(restartReducer.size()-1,"T","$\\mathrm{T}$","$B_\\mathrm{vol,vg}$","1.0");
```
ursg: Yup, this seems correct.

markusbattarbee:
Since it's for the same purpose, how about using this PR to also re-activate vg_e_vol, and add it to both bulk and restarts?

ykempf:
Well that's code we actually removed I think, a bit more work resurrecting. But sure.

ykempf:
Do you also want fg_e_vol? It's computed anyway.

ykempf:
Like so? Should be run-tested at least a tiny bit, after lunch maybe.

datareduction/datareducer.cpp
```diff
@@ -574,6 +574,44 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
+                      FsGrid< fsgrids::technical, 2>& technicalGrid)->std::vector<double> {
+
+               std::array<int32_t,3>& gridSize = technicalGrid.getLocalSize();
+               std::vector<double> retval(gridSize[0]*gridSize[1]*gridSize[2]*3);
+
+               // Iterate through fsgrid cells and extract EVOL
+               for(int z=0; z<gridSize[2]; z++) {
+                  for(int y=0; y<gridSize[1]; y++) {
+                     for(int x=0; x<gridSize[0]; x++) {
+                        retval[3*(gridSize[1]*gridSize[0]*z + gridSize[0]*y + x)] =  + (*volGrid.get(x,y,z))[fsgrids::EXVOL];
```
ykempf: - What are these dangling `+` after `=`?
- Is the indexing in the final bracket correct? Shouldn't it be `fsgrids::volfields::EXVOL`?

markusbattarbee: Well the plus shouldn't affect anything.
The indexing looks correct, as it's the 3 components. Can be EXVOL, EYVOL, EZVOL or EXVOL, EXVOL+1, EXVOL+2

ykempf: But it's missinf volfields in between, no?

parameters.cpp
```diff
@@ -261,7 +261,7 @@ bool Parameters::addParameters(){
    // NOTE Do not remove the : before the list of variable names as this is parsed by tools/check_vlasiator_cfg.sh
    Readparameters::addComposing("variables.diagnostic", std::string()+"List of data reduction operators (DROs) to add to the diagnostic runtime output. Each variable to be added has to be on a new line diagnostic = XXX. Names are case insensitive. "+
 				"Available (20190320): "+
-				"populations_blocks "+
+				"populations_vg_blocks "+
```
markusbattarbee: You may note that diagnostic variables don't have vg_ or fg_ as they are not saved on either grid, actually.

ykempf: But it is confusing when running the cfg check tool. Nad it does relate to vg cells and not fg cells.

datareduction/datareducer.cpp
markusbattarbee: I don't understand

ykempf: Seems to be a C++ feature I don't understand and seems to work also elsewhere, so it's ok here.

ykempf:
Like so?

datareduction/datareducer.cpp
markusbattarbee: Ah now I see. ok.

fieldsolver/ldz_volume.cpp
```diff
@@ -67,6 +67,95 @@ void calculateVolumeAveragedFields(
             volGrid0->at(fsgrids::volfields::PERBXVOL) = perturbedCoefficients[Rec::a_0];
             volGrid0->at(fsgrids::volfields::PERBYVOL) = perturbedCoefficients[Rec::b_0];
             volGrid0->at(fsgrids::volfields::PERBZVOL) = perturbedCoefficients[Rec::c_0];
+
+            // Calculate volume average of E (FIXME NEEDS IMPROVEMENT):
```
ursg: Öh, what does this "FIXME NEEDS IMPROVEMENT" refer to?

datareduction/datareducer.cpp
ursg: I guess that + was just a copy'n'paste leftover from some B reducer that summed up BGB and PerB here.

vlasiator.cpp
```diff
@@ -551,6 +546,11 @@ int main(int argn,char* args[]) {
       phiprof::stop("write-initial-state");
    }
 
+   phiprof::start("getFieldsFromFsGrid");
```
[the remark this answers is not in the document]
ykempf: But it breaks the indentation/parenthesis style. ;) I'll port that in then, thanks.

fieldsolver/ldz_volume.cpp
ykempf: The vol E is computed as the average of the four edge values, not as the proper Balsara reconstruction.

vlasiator.cpp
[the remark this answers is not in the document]
markusbattarbee: Since we now transfer the fields earlier on, I don't think this transfer is needed at all?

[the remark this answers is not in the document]
ykempf: Probably not.

ykempf:
Hahaa. I was certain you would want that. I did check that Evol comes out ok from the last version too.

ursg:
Yup. Tested and confirmed.

PR #457, opened 2019-11-15
Magnetosphere L4 ref

markusbattarbee:
Magnetosphere L4 refinement setup (was used for one test on sisu)


ykempf:
I suggest that we implement a dry-run mode for testing grid and boundary setups without the whole solvers. But that's not needed right away and would be a good task for a new coder in the project so let's not do it now. And Markus anyway said that just running the 0th step is good enough to test new grid setups.

ursg:
Yeps, this looks definitely good to be merged.
I suppose we should, but how do we test it? i suppose apart from looking that it doesn't crash, our tools aren't actually yet up to speed to vlsvdiff AMR data files, are they?

PR #458, opened 2019-11-15
Ip shock amr

markusbattarbee:


ursg:
Everything nicely hidden away behind if statements, so in the spirit of continuous integration, this one goes right into the dev branch!

PR #459, opened 2019-11-19
Fluffy ionosphere

ykempf:
This includes/branches from my HazelHenRevival branch which had the small changes I did to compile on HazelHen again. It also includes two `return` calls to make the corresponding non-void functions compliant.

This compiles after each of the commits but since HazelHen is down, testing is postponed. Up for discussion for now.


fieldsolver/ldz_magnetic_field.cpp
```diff
@@ -231,17 +240,48 @@ void propagateMagneticFieldSimple(
...
-            propagateMagneticField(perBGrid, perBDt2Grid, EGrid, EDt2Grid, i, j, k, dt, RKCase);
+            if(technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY) {
+               // Propagate B on all local cells on all faces:
+               propagateMagneticField(perBGrid, perBDt2Grid, EGrid, EDt2Grid, i, j, k, dt, RKCase, true, true, true);
+            } else {
+               // Easy case: in case we are neighboured by a non-sysboundary cell, we still solve the
+               // fields normally here.
+               cuint sysBoundaryLayer = technicalGrid.get(i,j,k)->sysBoundaryLayer;
+               if(sysBoundaryLayer == 1) {
+                  for (uint component = 0; component < 3; component++) {
```
markusbattarbee: The component loop here could be replaced with:
	bool prop_x = (technicalGrid.get(i-1,j,k)->sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY);
	bool prop_y = (technicalGrid.get(i,j-1,k)->sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY);
	bool prop_z = (technicalGrid.get(i,j,k-1)->sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY);
        if (prop_x || prop_y || prop_z) propagateMagneticField(perBGrid, perBDt2Grid, EGrid, EDt2Grid, i, j, k, dt, RKCase, prop_x, prop_y, prop_z);

ykempf: OK. Won't the compiler make it efficient anyway? But indeed it would be a bit more elegant.

fieldsolver/ldz_magnetic_field.cpp
```diff
@@ -191,8 +191,17 @@ void propagateSysBoundaryMagneticField(
...
+   cuint sysBoundaryLayer = technicalGrid.get(i,j,k)->sysBoundaryLayer;
+   
    for (uint component = 0; component < 3; component++) {
-      bGrid->at(fsgrids::bfield::PERBX + component) = sysBoundaries.getSysBoundary(sysBoundaryFlag)->fieldSolverBoundaryCondMagneticField(perBGrid, perBDt2Grid, EGrid, EDt2Grid, technicalGrid, i, j, k, dt, RKCase, component);
+      cint neigh_i=i + ((component==0)?-1:0);
+      cint neigh_j=j + ((component==1)?-1:0);
+      cint neigh_k=k + ((component==2)?-1:0);
+      cuint neighborSysBoundaryFlag = technicalGrid.get(neigh_i, neigh_j, neigh_k)->sysBoundaryFlag;
+      
+      if (neighborSysBoundaryFlag != sysboundarytype::NOT_SYSBOUNDARY) { // Complement to propagateMagneticFieldSimple main loop
```
markusbattarbee: Should this be also != sysboundarytype::DO_NOT_COMPUTE ?

sysboundary/ionosphere.cpp
```diff
@@ -613,14 +591,14 @@ namespace SBC {
       // Copy each face B-field from the cell on the other side of it
       switch(component) {
```
markusbattarbee: This block needs to be fixed so it copies from the simulation cell direction instead of just the -1 direction, such as:


// Copy each face B-field from the cell in the simulation cell direction
switch(component) {
	case 0:
		if (technicalGrid.get(i-1,j,k)->sysBoundaryLayer==1) {
			return bGrid->get(i-1,j,k)->at(fsgrids::bfield::PERBX + component);
		} else { // Is sysboundarylayer 2, read from opposite direction
			return bGrid->get(i+1,j,k)->at(fsgrids::bfield::PERBX + component);
		}
	case 1:
		if (technicalGrid.get(i,j-1,k)->sysBoundaryLayer==1) {
			return bGrid->get(i,j-1,k)->at(fsgrids::bfield::PERBX + component);
		} else { // Is sysboundarylayer 2, read from opposite direction
			return bGrid->get(i,j+1,k)->at(fsgrids::bfield::PERBX + component);
		}
	case 2:
		if (technicalGrid.get(i,j,k-1)->sysBoundaryLayer==1) {
			return bGrid->get(i,j,k-1)->at(fsgrids::bfield::PERBX + component);
		} else { // Is sysboundarylayer 2, read from opposite direction
			return bGrid->get(i,j,k+1)->at(fsgrids::bfield::PERBX + component);
		}
         default:
		cerr << "ERROR: ionosphere boundary tried to copy nonsensical magnetic field component " << component << endl;
		return 0.0;
      }

fieldsolver/ldz_magnetic_field.cpp
ykempf: Yep you also caught the first bug here. Line 200 segfaults at (i,j,k)=(0,0,0).

markusbattarbee: Hmm, darn. I guess we'll need to have a special check to see if i, j, or k is zero then.

ykempf: My last commit solves this.

ykempf: 5836b20 that is.

markusbattarbee: Not really, as for L2 it just directly calls the ionosphere Bfield boundary condition - and as I listed below, that needs to access the -1 and +1 neighbours anyway.

ykempf: Right, I'm adding that one below too. Interestingly it didn't segfault there... :^)

sysboundary/ionosphere.cpp
ykempf: Hmm but this will segfault if getting a -1 neighbour from a cell at coordinate 0. But I'll make it safe somehow.

ykempf: Doh we're in the ionosphere. But this means we likely break the support of having outflow crossing the ionosphere. Which we rarely used, though.

markusbattarbee: Right, and the outflow doesn't explicitly call a -1 or +1 neighbour, but instead copies from the closest internal cell.

ykempf: That's implemented now.

fieldsolver/ldz_magnetic_field.cpp
ykempf: Ah but this is actually activating more directions than intended.

fieldsolver/ldz_magnetic_field.cpp
```diff
@@ -191,8 +191,21 @@ void propagateSysBoundaryMagneticField(
       bGrid = perBDt2Grid.get(i,j,k);
    }
    cuint sysBoundaryFlag = technicalGrid.get(i,j,k)->sysBoundaryFlag;
+   cuint sysBoundaryLayer = technicalGrid.get(i,j,k)->sysBoundaryLayer;
+   
    for (uint component = 0; component < 3; component++) {
-      bGrid->at(fsgrids::bfield::PERBX + component) = sysBoundaries.getSysBoundary(sysBoundaryFlag)->fieldSolverBoundaryCondMagneticField(perBGrid, perBDt2Grid, EGrid, EDt2Grid, technicalGrid, i, j, k, dt, RKCase, component);
+      if (sysBoundaryFlag != 1) {
```
ursg: This should probably be NOT_SYSTBOUNDARY instead of a hardcoded 1. (I had to look up which it was)

fieldsolver/ldz_magnetic_field.cpp
```diff
@@ -231,17 +244,48 @@ void propagateMagneticFieldSimple(
...
    phiprof::start(timer);
    
-   #pragma omp parallel for collapse(3)
+   #pragma omp parallel for collapse(3) schedule(dynamic,1)
    for (int k=0; k<gridDims[2]; k++) {
       for (int j=0; j<gridDims[1]; j++) {
          for (int i=0; i<gridDims[0]; i++) {
 
             // Set the fsgrid rank in the technical grid
+            // FIXME Why is this done repeatedly here? Yann can't remember (20191119).
```
ursg: Doesn't quite seem to make any sense. As far as I can grep, this is the only point where that value ever gets touched. Maybe we can just do away with it altogether?

sysboundary/ionosphere.cpp
```diff
@@ -67,6 +67,7 @@ namespace SBC {
          Readparameters::add(pop + "_ionosphere.VX0", "Bulk velocity of ionospheric distribution function in X direction (m/s)", 0.0);
          Readparameters::add(pop + "_ionosphere.VY0", "Bulk velocity of ionospheric distribution function in X direction (m/s)", 0.0);
          Readparameters::add(pop + "_ionosphere.VZ0", "Bulk velocity of ionospheric distribution function in X direction (m/s)", 0.0);
+         Readparameters::add(pop + "_ionosphere.fluffiness", "Weight of boundary (0) vs. average of NOT_SYSBOUNDARY neighbor's (1) moments and velocity distribution.", 0);
```
ursg: The uninitiated would have no idea what to make of that help text. How about
"Inertia of boundary smoothing when copying neighbour's moments and velocity distributions (0=completely constant boundaries, 1=neighbours are interpolated immediately"

vlasovsolver/cpu_acc_intersections.cpp
```diff
@@ -29,7 +29,7 @@
 //#include <boost/unordered_map.hpp>
 //#include <boost/unordered_set.hpp>
 
-#include <Eigen/Geometry>
+#include <Geometry> //Eigen
```
ursg: These Eigen include changes break compilation for me on my fridge. Are you sure the makefile you're using shouldn't just have its include paths changed?

fieldsolver/ldz_magnetic_field.cpp
ykempf: Who doesn't know that. :) ;)

ykempf: NO this is a bug! Should be layer!

fieldsolver/ldz_magnetic_field.cpp
ykempf: No we want to set it once somewhere for IO purposes. But maybe we can do that in datareducer.cpp if it's only used in IO?

vlasovsolver/cpu_acc_intersections.cpp
ykempf: I had to do that when installing on HazelHen. I had the impression that there was one layer of directories too much or which had disappeared int he version of Eigen I took now. Maybe it was my mistake.

fieldsolver/ldz_magnetic_field.cpp
ykempf: Good catch, that reduced the diffs I am seeing.

MAKE/Makefile.hornet_gcc
```diff
@@ -22,53 +38,51 @@ FLAGS =
...
 
-
 LIB_PROFILE = -L$(LIBRARY_PREFIX)/mpich2/$(MPT_VERSION)/$(CC_BRAND)/$(CC_BRAND_VERSION)/phiprof/lib -lphiprof
-#LIB_PROFILE = $(LIBRARY_PREFIX)/mpich2/$(MPT_VERSION)/$(CC_BRAND)/$(CC_BRAND_VERSION)/phiprof/lib/libphiprof.a
 INC_PROFILE = -I$(LIBRARY_PREFIX)/mpich2/$(MPT_VERSION)/$(CC_BRAND)/$(CC_BRAND_VERSION)/phiprof/include
 
 #header libraries
 
-INC_EIGEN = -I$(LIBRARY_PREFIX)/eigen/
+INC_EIGEN = -I$(LIBRARY_PREFIX)/Eigen/
```
ursg: I suppose the Eigen includes can be made to work simply by removing the Eigen/ here?

ykempf: Yep. Shall I add this here or shall we untangle that in a separate branch?

ursg: I guess it would make sense to have a separate makefile update pull request.

ykempf:
Here's my commit hopefully implementing what we discussed. It's queuing now.

sysboundary/sysboundary.cpp
```diff
@@ -523,6 +525,64 @@ bool SysBoundary::classifyCells(dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Ca
    
    technicalGrid.updateGhostCells();
    
+   const std::array<int,3> fsGridDimensions = {convert<int>(P::xcells_ini) * pow(2,P::amrMaxSpatialRefLevel),
+      convert<int>(P::ycells_ini) * pow(2,P::amrMaxSpatialRefLevel),
+      convert<int>(P::zcells_ini) * pow(2,P::amrMaxSpatialRefLevel)};
```
markusbattarbee: Perhaps use fsgrid globalsize here instead?

fieldsolver/ldz_electric_field.cpp
```diff
@@ -1531,95 +1531,66 @@ void calculateElectricField(
    
    if (cellSysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE) return;
    
-   cuint cellSysBoundaryLayer = technicalGrid.get(i,j,k)->sysBoundaryLayer;
-   
-   if ((cellSysBoundaryFlag != sysboundarytype::NOT_SYSBOUNDARY) && (cellSysBoundaryLayer > 1)) {
-      // Sysboundary level 2+ cells
+   cuint mask = technicalGrid.get(i,j,k)->SOLVE;
```
markusbattarbee: Perhaps this isn't technically a mask, it's a bitfield :) We're using e.g. compute::EX as a mask on it

sysboundary/ionosphere.cpp
```diff
@@ -555,75 +554,169 @@ namespace SBC {
...
+                  return retval / nCells;
+               }
+            default:
+               cerr << "ERROR: ionosphere boundary tried to copy nonsensical magnetic field component " << component << endl;
+               return 0.0;
+         }
+      } else { // L2 cells
+         Real retval = 0.0;
+         uint nCells = 0;
+         for (int a=i-1; a<i+2; a++) {
```
markusbattarbee: I think this bit can go outside the grid, as there isn't a check first if i,j,k is zero or nx-1,ny-1,nz-1?

fieldsolver/ldz_magnetic_field.cpp
ykempf: I moved that.

sysboundary/sysboundary.cpp
ykempf: I took what was used to initialise the fsgrids, but indeed, this is a better suggestion.

sysboundary/ionosphere.cpp
ykempf: Now we kind of require the ionosphere not to touch the outer walls. I'll look through again as I solved this issue in another place using the global size of the fsgrid.

markusbattarbee: Oh right I forgot we're in ionosphere and not general sysboundaries :)

sysboundary/sysboundary.cpp
ykempf: But there is no getGlobalSize in fsgrid...

sysboundary/ionosphere.cpp
ykempf: I can add a check against the global size as in sysboundary.cpp when assigning the bit field, if you want. In practice we never had the ionosphere croos an outer wall in production, only in testing.

vlasovsolver/cpu_moments.cpp
```diff
@@ -161,7 +156,7 @@ void calculateMoments_R(
        for (size_t c=0; c<cells.size(); ++c) {
           SpatialCell* cell = mpiGrid[cells[c]];
           
-          if (cell->sysBoundaryFlag != sysboundarytype::NOT_SYSBOUNDARY && cell->sysBoundaryLayer != 1) {
+          if (cell->sysBoundaryFlag != sysboundarytype::DO_NOT_COMPUTE) {
```
ykempf: Doh!

fieldsolver/ldz_electric_field.cpp
ykempf: Better now?

sysboundary/sysboundary.cpp
ykempf: That's been implemented.

sysboundary/ionosphere.cpp
ykempf: Pasted that in.

ursg:
This looks really good. Maybe we should merge this right away to have it in dev.

One thing we discussed would be to have the fluffyness parametrized in physical units, such as a 1/e time of boundary VDF assimilation. Technically, this would mean that we specify a fluffyness parameter a in units of [a]=1/s, so that our actual numerical fluffyness can be calculated at every timestep as num_fluffyness = dt*a 

That way, a=0 would be zero fluffyness. The alternative would be to have the inverse of it as the parameter (in which case the fluffyness time would be specified in seconds), but then we'd need special case handling for "no fluffyness".

Anyway, that parametrization can be a separate pull request.

PR #464, opened 2020-02-18
AmrFiltering 

kstppd:


fieldsolver/gridGlue.cpp
```diff
@@ -179,6 +190,210 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+
+              std::array weights[i]
+    weight= ---------------------------
+                2*refLevel+1
+
+  and are calculated as shown above while iterating over each dimensions
+  The numerator is calculated filling the weights array
+  and the denominator depends only on the stencil width. 
+  */
+  float weightC, weightL1, weightL2, weightR1, weightR2;
```
ykempf: I would suggest to make these `Real`, a type we define as float or double at compile time.

grid.cpp
```diff
@@ -260,7 +262,27 @@ void initializeGrids(
...
+      #pragma omp parallel for collapse(3)
+      for (int k=0; k<localDims[2]; k++) {
+         for (int j=0; j<localDims[1]; j++) {
+            for (int i=0; i<localDims[0]; i++) {
+
+               const std::array<int, 3> mapIndices = momentsGrid.getGlobalIndices(i,j,k);
+               const dccrg::Types<3>::indices_t  indices = {{(uint64_t)mapIndices[0],(uint64_t)mapIndices[1],(uint64_t)mapIndices[2]}}; //cast to avoid warnings
+               CellID dccrgCellID2 = mpiGrid.get_existing_cell(indices, 0, mpiGrid.mapping.get_maximum_refinement_level());
+               int amrLevel= mpiGrid.get_refinement_level(dccrgCellID2);
+               technicalGrid.get(i, j, k)-> RefLevel =amrLevel ;
```
ykempf: DCCRG's/mpiGrid's refinement level is 0 (or 1? darn, I don't remember) at coarsest, 1 (or 2?) for the next finer level and so on. I saw that you skip if the level is 0 in the smoothing loops, right? So I think that it is reversed.

fieldsolver/gridGlue.cpp
```diff
@@ -179,6 +190,210 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+  int blurPasses = 3;
+  int refLevel;
+  std::array<int, 5> weights;
```
ykempf: Is 5 a constant? Or does it depend on the maximum refinement? See comments below too.

fieldsolver/gridGlue.cpp
```diff
@@ -179,6 +190,210 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+
+          // Dont filter if in max resolution region
+          refLevel = technicalGrid.get(ii, jj, kk)->RefLevel;
+          if (refLevel == 0)
+          {
+            continue;
+          }
+
+          // Calculate Numerator
+          for (int index = 0; index < refLevel; index++)
```
ykempf: The refinement level can be anything, 3 or more, and then this will overflow. See comment above too.

fieldsolver/gridGlue.cpp
```diff
@@ -367,6 +582,17 @@ void getFieldsFromFsGrid(
   
   MPI_Waitall(sendRequests.size(), sendRequests.data(), MPI_STATUSES_IGNORE);
   
+
```
ykempf: Space to take notes? ;)

projects/Fluctuations/diagnostic.txt
```diff
@@ -0,0 +1,10 @@
+# Column 1 Step
```
ykempf: Don't add these files please.

projects/Fluctuations/logfile.txt
```diff
@@ -0,0 +1,107 @@
+Tue Feb 18 11:48:24 2020
```
ykempf: Same.

vlasiator.cpp
```diff
@@ -408,6 +408,8 @@ int main(int argn,char* args[]) {
    FsGrid< std::array<Real, fsgrids::ehall::N_EHALL>, 2> EHallGrid(fsGridDimensions, comm, periodicity,gridCoupling);
    FsGrid< std::array<Real, fsgrids::egradpe::N_EGRADPE>, 2> EGradPeGrid(fsGridDimensions, comm, periodicity,gridCoupling);
    FsGrid< std::array<Real, fsgrids::moments::N_MOMENTS>, 2> momentsGrid(fsGridDimensions, comm, periodicity,gridCoupling);
+   FsGrid< std::array<Real, fsgrids::moments::N_MOMENTS>, 2> swapGrid(fsGridDimensions, comm, periodicity,gridCoupling);
```
ykempf: I don't think that we need to drag it along from here and pass it down, it can be declared with the = operator on the spot. Saves some memory.

fieldsolver/gridGlue.cpp
kstppd: I thought these would never need to be double.

ykempf:
Yes but then it's consistent with the format of the data in the fsgrid. Probably doesn't matter that much but I'm not quite sure.

On 19 February 2020 01:42:14 EET, Kostis Papadakis <notifications@github.com> wrote:
>kstppd commented on this pull request.
>
>
>
>> +  std::array<Real,fsgrids::moments::N_MOMENTS> *GridR1;
>+  std::array<Real,fsgrids::moments::N_MOMENTS> *GridR2;
>+  std::array<Real,fsgrids::moments::N_MOMENTS> *swap;
>+
>+
>+  /*Kernel Weights are fractional numbers 
>+
>+              std::array weights[i]
>+    weight= ---------------------------
>+                2*refLevel+1
>+
>+  and are calculated as shown above while iterating over each
>dimensions
>+  The numerator is calculated filling the weights array
>+  and the denominator depends only on the stencil width. 
>+  */
>+  float weightC, weightL1, weightL2, weightR1, weightR2;
>
>I thought these would never need to be double. 
>
>-- 
>https://github.com/fmihpc/vlasiator/pull/464#discussion_r381000890

-- 

fieldsolver/gridGlue.cpp
```diff
@@ -199,28 +198,34 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
-
+  auto swapGrid= momentsGrid;
+ 
 
   /*Kernel Weights are fractional numbers 
 
               std::array weights[i]
     weight= ---------------------------
-                2*refLevel+1
+                2*halfStencilWidht+1
```
ykempf: typo ;)

ykempf: In a comment, never mind.

fieldsolver/gridGlue.cpp
```diff
@@ -231,34 +236,43 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
           {
             weights.at(2 + index) = 1;
             weights.at(2 - index) = 1;
           }
           // Calculate Denominator
-          int weightDenominator = 2 * refLevel + 1;
+          int weightDenominator = 2 * halfStencilWidth + 1;
 
           // Calculate Weights
           weightL2 = weights.at(0) / weightDenominator;
```
ykempf: Beware of integer division.

common.h
```diff
@@ -367,6 +367,8 @@ namespace fsgrids {
       int sysBoundaryLayer; /*!< System boundary layer index. */
       Real maxFsDt;         /*!< maximum timestep allowed in ordinary space by fieldsolver for this cell**/
       int fsGridRank;       /*!< Rank in the fsGrids cartesian coordinator */
+      uint SOLVE;           /*!< Bit mask to determine whether a given cell should solve E or B components. */
```
ursg: Is this one a leftover from the PML code?
I guess this is not relevant for the filtering itself, right?

common.h
```diff
@@ -386,6 +388,15 @@ namespace sysboundarytype {
...
 
+namespace compute {
+   const uint BX = (1 << 0); // 1
+   const uint BY = (1 << 1); // 2
+   const uint BZ = (1 << 2); // 4
+   const uint EX = (1 << 3); // 8
+   const uint EY = (1 << 4); // 16
+   const uint EZ = (1 << 5); // 32
+}
+
```
ursg: Likewise, I suppose these ones are PML defines?

fieldsolver/gridGlue.cpp
```diff
@@ -179,6 +189,244 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
   MPI_Waitall(sendRequests.size(), sendRequests.data(), MPI_STATUSES_IGNORE);
 
+  // ---------------------------------------------------------------------------------
+  // Boxcar Smoothening
+  const int *mntDims= &momentsGrid.getLocalSize()[0];
+  std::array<Real,fsgrids::moments::N_MOMENTS> *Grid;
+  std::array<Real,fsgrids::moments::N_MOMENTS> *GridL1;
+  std::array<Real,fsgrids::moments::N_MOMENTS> *GridL2;
+  std::array<Real,fsgrids::moments::N_MOMENTS> *GridR1;
+  std::array<Real,fsgrids::moments::N_MOMENTS> *GridR2;
```
ursg: I would actually call these "cell" instead of "grid" since they are pointers to the data of a single grid cell each.
(Also: in the Vlasiator coding style, variable names start with lower case and datatype names with upper case!
See https://github.com/fmihpc/vlasiator/wiki/Coding-Style
)

fieldsolver/gridGlue.cpp
```diff
@@ -179,6 +189,244 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+  std::array<Real,fsgrids::moments::N_MOMENTS> *swap;
+  auto swapGrid= momentsGrid;
```
ykempf: @ursg would `FsGrid< std::array<Real, fsgrids::moments::N_MOMENTS>, 2> swapGrid = momentsGrid;` here work? Or with `= momentsGrid` in the next line? Is there any difference or prefernece vs. `auto`?

fieldsolver/gridGlue.cpp
```diff
@@ -179,6 +189,244 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+  const int maxStencilWidth=5;
+  int blurPasses = 3;
+  int refLevel;
+  int halfStencilWidth;
+  std::array<int, maxStencilWidth> weights;
+  weights.fill(0);
+
+
+  for (int blurPass = 1; blurPass <= blurPasses; blurPass++)
+  {
```
ursg: Another coding style comment: opening brace on the same line as the for loop!

fieldsolver/gridGlue.cpp
```diff
@@ -179,6 +189,244 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+          }
+          // Calculate Denominator
+          int weightDenominator = 2 * halfStencilWidth + 1;
+
+          // Calculate Weights
+          weightL2 = weights.at(0) / weightDenominator;
+          weightL1 = weights.at(1) / weightDenominator;
+          weightC  = weights.at(2) / weightDenominator;
+          weightR1 = weights.at(3) / weightDenominator;
+          weightR2 = weights.at(4) / weightDenominator;
```
ursg: Since we only have a small number (4) of possible refinement levels here, the weight values only have 4 different outcomes, right?
Instead of recalculating them for every cell, we could also prebuild the 4 possible weights arrays before the loop and then just look them up.
(This is just a performance enhancement, let's get it to work in principle fist)

fieldsolver/gridGlue.cpp
```diff
@@ -179,6 +189,244 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+          swap->at(fsgrids::moments::VY) = weightC * Grid->at(fsgrids::moments::VY) + weightL1 * GridL1->at(fsgrids::moments::VY) + weightL2 * GridL2->at(fsgrids::moments::VY) + weightR1 * GridR1->at(fsgrids::moments::VY) + weightR2 * GridR2->at(fsgrids::moments::VY);
+          swap->at(fsgrids::moments::P_11) = weightC * Grid->at(fsgrids::moments::P_11) + weightL1 * GridL1->at(fsgrids::moments::P_11) + weightL2 * GridL2->at(fsgrids::moments::P_11) + weightR1 * GridR1->at(fsgrids::moments::P_11) + weightR2 * GridR2->at(fsgrids::moments::P_11);
+          swap->at(fsgrids::moments::P_22) = weightC * Grid->at(fsgrids::moments::P_22) + weightL1 * GridL1->at(fsgrids::moments::P_22) + weightL2 * GridL2->at(fsgrids::moments::P_22) + weightR1 * GridR1->at(fsgrids::moments::P_22) + weightR2 * GridR2->at(fsgrids::moments::P_22);
+          swap->at(fsgrids::moments::P_33) = weightC * Grid->at(fsgrids::moments::P_33) + weightL1 * GridL1->at(fsgrids::moments::P_33) + weightL2 * GridL2->at(fsgrids::moments::P_33) + weightR1 * GridR1->at(fsgrids::moments::P_33) + weightR2 * GridR2->at(fsgrids::moments::P_33);
+
+          // Reset Weights
+          weights.fill(0);
```
ursg: This isn't necessary, since they anyway get overwritten in the next loop iteration

fieldsolver/gridGlue.cpp
```diff
@@ -179,6 +189,244 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+        }
+      }
+    }
+    momentsGrid.updateGhostCells();
+    swapGrid.updateGhostCells();
+
+    // Y Dimension Pass
```
ursg: Now you have 3 identical loops that only differ in their ii, jj and kk offsets. Maybe combine them into one, where you just have a
```
for(int dimension=0; dimension <3; dimension++) {
   std::array<int,3> coord_offset={0,0,0};
   coord_offset[dimension]=1;
   for( ... inner loops ... ) {
      //...
      GridL1 = momentsGrid.get(ii - coord_offset[0], jj - coord_offset[1], kk - coord_offset[2]);
      // and so on
   }
}
```

projects/Fluctuations/phiprof_0.txt
```diff
@@ -0,0 +1,48 @@
...
+ | 107 | 5   |     |         semilag-trans               | 5     | 0.4232  | 100    | 0     | 1  |       |       |                          | 
+ | 109 | 6   |     |           translate proton          | 5     | 0.4213  | 99.54  | 0     | 1  |       |       |                          | 
+ | 114 | 7   | D   |             update_remote-z         | 5     | 0.1298  | 30.81  | 0     | 1  |       |       |                          | 
+ | 119 | 7   | D   |             update_remote-x         | 5     | 0.1307  | 31.02  | 0     | 1  |       |       |                          | 
+ | 124 | 7   | D   |             update_remote-y         | 5     | 0.1291  | 30.64  | 0     | 1  |       |       |                          | 
+ | 128 | 4   |     |       Propagate Fields              | 5     | 0.4488  | 51.01  | 0     | 1  |       |       | 2.447e+04 SpatialCells/s | 
+ | 129 | 5   |     |         fsgrid-coupling-in          | 5     | 0.4112  | 91.61  | 0     | 1  |       |       |                          | 
+----------------------------------------------------------------------------------------------------------------------------------------------
```
ursg: I guess there's no real reason for this file to be in the pullrequest, right?

projects/Shock/logfile.txt
```diff
@@ -0,0 +1,53 @@
...
+(IO): .... done!
+
+Thu Feb 13 13:34:56 2020 
+(MAIN): All timesteps calculated.
+	 (TIME) total run time 0.0350031 s, total simulated time 8.08962e+307 s
+	 (TIME) seconds per timestep 0.0350031, seconds per simulated second 4.32691e-310
+
+Thu Feb 13 13:34:56 2020 
+(MAIN): Exiting.
+
```
ursg: Likewise, these other extra log files shouldn't be pulled with it

fieldsolver/gridGlue.cpp
ursg: In principle, both should work just fine.
But for readability, I guess it's better to actually write the whole type (```FsGrid< std::array<Real, fsgrids::moments::N_MOMENTS>, 2>```) there, since otherwise people reading the code might have no idea what's going on. :)

kstppd: Ok thanks!

common.h
kstppd: These  is some conflict for common.h which has confused me. This is actually not from the PML code.

common.h
kstppd: Neither are these

fieldsolver/gridGlue.cpp
kstppd: Ok!

fieldsolver/gridGlue.cpp
```diff
@@ -179,6 +189,244 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+  and the denominator depends only on the stencil width. 
+  */
+  Real weightC, weightL1, weightL2, weightR1, weightR2;
+  int maxRefLevel = mpiGrid.mapping.get_maximum_refinement_level();
+  int minRefLevel = 0;
+  const int maxStencilWidth=5;
+  int blurPasses = 3;
+  int refLevel;
+  int halfStencilWidth;
+  int switchPoint=1;
```
kstppd: Declared a switchPoint for specifying the switch between 3 and 5 point stencils.

common.h
ykempf: That's our new way of doing @ursg, since the fluffiness-related Yee-fixing of boundaries.

common.h
ykempf: ditto

fieldsolver/gridGlue.cpp
kstppd: I think it should be there. If I am right,switching from a 5 point stencil area to a 3 point stencil  area the first and last elements should be zero . The 3 point stencil will overwrite the middle values only though so I will end up with a wrong stencil.

fieldsolver/gridGlue.cpp
kstppd: That's nice! Will implement.

common.h
ursg: Oh, alright, true. Then this is all good.

fieldsolver/gridGlue.cpp
ursg: Ah, I see. True.!

fieldsolver/gridGlue.cpp
kstppd: Yes I thought of that too but skipped it till now.  It is actually easier as the weight values only depend on the stencil width. So since we are using a 3 and 5 point stencil the weights have 2 possible outcomes. Will leave this for a bit later though. Trying to merge the loops first

grid.cpp
kstppd: Guess resolved?

fieldsolver/gridGlue.cpp
kstppd: Hmm actually I had something wrong in the previous commit. I updated it. The three loops are not identical because of the array swapping. Thus I am not sure this can be implemented.

fieldsolver/gridGlue.cpp
```diff
@@ -189,256 +189,116 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
-          // Reset Weights
-          weights.fill(0);
         }
       }
     }
-    momentsGrid.updateGhostCells();
+    phiprof::stop("BlurPass");
+
+    phiprof::start("GhostUpdate");
     swapGrid.updateGhostCells();
```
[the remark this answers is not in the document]
kstppd: Don't we have to update the swapGrid ghost cells at least before copying back to momentsGrid?

fieldsolver/gridGlue.cpp
```diff
@@ -189,256 +189,116 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
 
 
-          // Dont filter if in max resolution region
-          refLevel = technicalGrid.get(ii, jj, kk)->RefLevel;
-          if (refLevel == maxRefLevel || technicalGrid.get(ii, jj, kk)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE){
-            continue;
-          }
+                int xn=ii+a-kernelOffset;
+                int yn=jj+b-kernelOffset;
+                int zn=kk+c-kernelOffset;
```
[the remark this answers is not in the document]
kstppd: That would make clearer which elements are being filtered in the moments grid. Ok will implement now.

fieldsolver/gridGlue.cpp
```diff
@@ -189,256 +189,116 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
-    momentsGrid.updateGhostCells();
+    phiprof::stop("BlurPass");
+
+    phiprof::start("GhostUpdate");
     swapGrid.updateGhostCells();
+    phiprof::stop("GhostUpdate");
 
     //  Copy swapGrid back to momentsGrid
-    // #pragma omp parallel for collapse(3)
+    phiprof::start("Swap->Moments");
```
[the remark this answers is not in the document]
kstppd: This has been mentioned before and indeed it would make it more readable but it fails to compile when trying to swap pointers.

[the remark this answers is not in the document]
markusbattarbee: @ursg, do your magic? :)

fieldsolver/gridGlue.cpp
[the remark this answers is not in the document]
markusbattarbee: Wouldn't think so since we do the ghost cell upgrade after replacing the data (l296). But Others can verify.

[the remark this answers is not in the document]
ykempf: True, we always only loop on the local domain (excluding ghost cells) so it should be safe. It would be different if we were accessing neighbours which could then potentially be ghosts.

fieldsolver/gridGlue.cpp
```diff
@@ -179,6 +189,117 @@ void feedMomentsIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+      for (int jj = 0; jj < mntDims[1]; jj++){
+        for (int kk = 0; kk < mntDims[2]; kk++){
+          for (int e = 0; e < fsgrids::moments::N_MOMENTS; ++e){
+
+            // Get Arrays
+            cell = momentsGrid.get(ii, jj, kk);
+            swap = swapGrid.get(ii, jj, kk);
+
+            //Copy
+            cell->at(e) = swap->at(e);
```
ykempf, line 287: Really this could be a one-liner, less hassle for me to read but I don't mind if you prefer like this. The compiled code will be identical I assume.

projects/Shock/phiprof_0.txt
```diff
@@ -0,0 +1,30 @@
+
```
ykempf, line 1: There's a bunch of files still lying around that are not needed in the repo.

fieldsolver/gridGlue.cpp
```diff
@@ -63,7 +69,8 @@ template <typename T, int stencil> void computeCoupling(dccrg::Dccrg<SpatialCell
 						      (uint64_t)globalIndices[1],
 						      (uint64_t)globalIndices[2]}}; //cast to avoid warnings
 	CellID dccrgCell = mpiGrid.get_existing_cell(indices, 0, mpiGrid.mapping.get_maximum_refinement_level());
-	int process = mpiGrid.get_process(dccrgCell);
+  
+  int process = mpiGrid.get_process(dccrgCell);
```
ykempf, line 73: And let's wrap up this short review with a new complaint about weird indentations. ;)

But, still: please use 3-space indentation levels as we have throughout the code. Let's not mix that or it will become a pain later. Some of the files I poked at had 2-space indentation levels.

kstppd: Ok updating shortly!

fieldsolver/gridGlue.cpp
kstppd: This has been requested many times I know, however the compiler doesn't let me swap pointers between the two of them. I must be doing something wrong I guess. I would also prefer a one liner here for readability purposes

markusbattarbee:
With the adaptive filter passes activated the parameter isn't needed anymore

fieldsolver/gridGlue.cpp
ykempf: :)

parameters.cpp
```diff
@@ -84,6 +84,8 @@ vector<int> P::systemWriteDistributionWriteStride;
 vector<int> P::systemWriteDistributionWriteXlineStride;
 vector<int> P::systemWriteDistributionWriteYlineStride;
 vector<int> P::systemWriteDistributionWriteZlineStride;
+vector<Real> P::systemWriteDistributionWriteShellRadius;
```
ykempf, line 87: Why does this show here? Interesting.

fieldsolver/gridGlue.cpp
ykempf: !

ykempf:
2/3rd majority is good enough. :) All discussions were resolved.

PR #465, opened 2020-02-24
Pml

kstppd:


sysboundary/pml.cpp
```diff
@@ -0,0 +1,275 @@
...
+                pmlValue->at(fsgrids::pml::PGK3) = 1.0;
+                pmlValue->at(fsgrids::pml::PFK1) = 1.0;
+                pmlValue->at(fsgrids::pml::PFK2) = 1.0;
+                pmlValue->at(fsgrids::pml::PFK3) = 1.0;
+            }
+        }
+    }
+
+
+   Real xnum, xd;
```
ykempf: This is now not thread-safe as these variables will be accessed by all threads. If all the loops are otherwise thread-safe it would anyway be more efficient to start the threaded region `#pragma omp parallel` before the first threaded loop, then have `#pragma omp for collapse(3)` for the loops. This means that the threads are being forked only once and not every loop. Also, then each thread would have its own private copy of these four variables, as you don't want other threads messing with them in between.

sysboundary/pml.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+    /*-----Initially set all arrays to one-----*/
+    /*Iterate over domain and set the PML arrays to 1.0 thus not affecting the fieldsolver*/ 
+    #pragma omp parallel for collapse(3)
+    for (int kk = 0; kk < pmlDims[2]; kk++){
+        for (int jj = 0; jj < pmlDims[1]; jj++){
+            for (int ii = 0; ii < pmlDims[0]; ii++){
+
+                pmlValue = pmlGrid.get(ii, jj, kk);
+                pmlValue->at(fsgrids::pml::PGI2) = 1.0;
```
ykempf: Since `fsgrids::pml` is an enum, you could have avoided typing pain and done a loop from 0 to N_PML. This would actually reduce the maintenance issues if this array changes later.

sysboundary/pml.cpp
```diff
@@ -0,0 +1,275 @@
...
+bool buildPMLGrids(
+    FsGrid<std::array<Real, fsgrids::pml::N_PML>, 2> &pmlGrid)
+{
+
+   typedef Parameters P;
+
+
+   /*----Get PML Array and Domain Size-----*/
+   std::array<Real, fsgrids::pml::N_PML> *pmlValue;
+   int start = 2;   /*don't judge will change it */
```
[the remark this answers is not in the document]
ykempf: Indeed this will not work in AMR setups.

sysboundary/pml.cpp
```diff
@@ -0,0 +1,275 @@
...
+   #pragma omp parallel for collapse(3)
+   for (int kk = 0; kk < pmlDims[2] ; kk++){
+      for (int jj = 0; jj < pmlDims[1]; jj++){
+         for (int ii = 0; ii <pmlDims[0]; ii++){
+
+         // Get Global Arrays
+         pos=pmlGrid.getGlobalIndices(ii,jj,kk);
+
+         if (Xm==true){
+            if (pos[0]>=start && pos[0]<=P::pmlWidthX+start){
```
ykempf: These two ifs can be combined but the compiler will probably do it anyway.

This `pos` also needs to be thread-private. Do check all the variables. Or remove threading for now and we can do it at a later stage.

sysboundary/pml.cpp
```diff
@@ -0,0 +1,275 @@
...
+               } 
+            }
+         }
+      }
+   }
+
+
+// Attentuation Along the Y-Dimension
+   #pragma omp parallel for collapse(3)
+   for (int kk = 0; kk < pmlDims[2] ; kk++){
```
ykempf: There is no reason to change the order of the loops so keep them k, j, i as this is more efficient for memory accesses. But then if it's anyway a threaded collapsed set of iterations it makes no difference until we run without threading, isn't it? Ping @ursg.

sysboundary/pml.h
```diff
@@ -0,0 +1,37 @@
+/*
+ * This file is part of Vlasiator.
+ * Copyright 2010-2016 Finnish Meteorological Institute
```
ykempf: Also UH 2020. :)

sysboundary/sysboundary.cpp
```diff
@@ -432,7 +432,7 @@ bool SysBoundary::classifyCells(dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Ca
    mpiGrid.update_copies_of_remote_neighbors(SYSBOUNDARIES_NEIGHBORHOOD_ID);
 
    /*Compute distances*/
-   uint maxLayers=3;//max(max(P::xcells_ini, P::ycells_ini), P::zcells_ini);
+   uint maxLayers=22;//max(max(P::xcells_ini, P::ycells_ini), P::zcells_ini);
```
ykempf: This is now a function of the cfg parameter, isn't it? Actually there should be a min of 3, to catch the cases where we'd set all the PML widths to zero.

vlasiator.cpp
```diff
@@ -489,7 +500,9 @@ int main(int argn,char* args[]) {
 		   BgBGrid,
 		   volGrid,
 		   technicalGrid,
-		   sysBoundaries, 0.0, 1.0
```
ykempf: Greetings from the random indentation service. ;)

Makefile
```diff
@@ -26,7 +26,8 @@ AR ?= ar
 FIELDSOLVER ?= londrillo_delzanna
 #Add -DFS_1ST_ORDER_SPACE or -DFS_1ST_ORDER_TIME to make the field solver first-order in space or time
 # COMPFLAGS += -DFS_1ST_ORDER_SPACE
-# COMPFLAGS += -DFS_1ST_ORDER_TIME
+# COMPFLAGS += -DFS_1ST_ORDER_TIME 
+COMPFLAGS += -DFS_1ST_ORDER_TIME
```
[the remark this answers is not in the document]
ykempf: Although in practice we have subcycling which will go to 1st order time stepping anyway...

fieldsolver/ldz_magnetic_field.cpp
```diff
@@ -70,13 +71,25 @@ void propagateMagneticField(
...
             EGrid2 = EGrid.get(i,j,k+1);
-            perBGrid0->at(fsgrids::bfield::PERBX) += dt/dz*(EGrid2->at(fsgrids::efield::EY) - EGrid0->at(fsgrids::efield::EY)) + dt/dy*(EGrid0->at(fsgrids::efield::EZ) - EGrid1->at(fsgrids::efield::EZ));
+            
+            // GetPMLarrays for Bx Component   
+            pmlGrid0=pmlGrid.get(i,j,k);
+
+            // Update  Bx
+            // perBGrid0->at(fsgrids::bfield::PERBX) += dt/dz*(EGrid2->at(fsgrids::efield::EY) - EGrid0->at(fsgrids::efield::EY)) + dt/dy*(EGrid0->at(fsgrids::efield::EZ) - EGrid1->at(fsgrids::efield::EZ));
+
+            perBGrid0->at(fsgrids::bfield::PERBX) = pmlGrid0->at(fsgrids::pml::PFJ3) * pmlGrid0->at(fsgrids::pml::PFK3) * perBGrid0->at(fsgrids::bfield::PERBX) + pmlGrid0->at(fsgrids::pml::PFJ2) * pmlGrid0->at(fsgrids::pml::PFK2) * (dt / dz) * (EGrid2->at(fsgrids::efield::EY) - EGrid0->at(fsgrids::efield::EY)) + dt / dy * (EGrid0->at(fsgrids::efield::EZ) - EGrid1->at(fsgrids::efield::EZ));
```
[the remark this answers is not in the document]
kstppd: Yes it is intended.

fieldsolver/ldz_magnetic_field.cpp
```diff
@@ -70,13 +71,25 @@ void propagateMagneticField(
    std::array<Real, fsgrids::efield::N_EFIELD> * EGrid2;
    std::array<Real, fsgrids::bfield::N_BFIELD> * perBDt2Grid0;
    
+   // Get PML Array
+   std::array<Real, fsgrids::pml::N_PML> * pmlGrid0;
+   
+
    if (doX == true) {
       switch (RKCase) {
          case RK_ORDER1:
```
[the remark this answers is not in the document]
kstppd: No problem with RK second order. I am swapping everything to these new branch but haven't yet copied over the second order RK

Makefile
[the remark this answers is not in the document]
markusbattarbee: I thought our current version allowed substepping of 2nd order as well (that being the most common parameter setup we've used)

sysboundary/pml.cpp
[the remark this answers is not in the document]
ykempf: Suggestion: instead of this `start` and the repeated calculation of the x/y/z positions that factor into the calculation at each iteration, these positions could be stored as extra fields in the pmlGrid and computed at initialisation. This would have the added benefit that you could there make use of the determination of the layers, as this is a similar process.

sysboundary/pml.cpp
kstppd: Well I need to switch the order as in each loop I want to move along one dimension. I guess that when they get collapsed ,it will be fine.

Makefile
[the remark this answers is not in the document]
ykempf: https://github.com/fmihpc/vlasiator/blob/dev/fieldsolver/ldz_main.cpp#L252
Heh, I should stop trusting my memory when quoting the code then.

vlasiator.cpp
kstppd: VScode does't show this indentation. But it is there I can see it in Vim.

sysboundary/pml.cpp
ykempf: Two ways of answering here:
1. When you collapse it just makes a stack of single iterations (or groups thereof) without any ordering.
2. The code doesn't seem to have any dependence on the order (which would kill threadability anyway) so I don't understand our point or didnät see the dependence.

Makefile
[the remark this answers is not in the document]
markusbattarbee: Ok, so we need to confirm this PML boundary has been tested with 2nd order FS as well.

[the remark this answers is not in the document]
ykempf: It has only been tested with test_fp so far, don't worry, "actual" testing is pending...

sysboundary/pml.cpp
kstppd: I was wrong. There is no dependence. My bad. I will keep the same order .

ykempf: Good. :) Actually then it could be that it would be most efficient to have one single k-j-i threaded-collapsed loop, no?

kstppd: Sounds good. I will give it some more thought a bit later and do it however I don't think it is that important as this function is only executed once. I will give it some thought though, and implement it later today.

sysboundary/sysboundary.cpp
kstppd: @ykempf . So since maxLayers is defined for every dimension, the pml cannot have different widths per dimension. That's a problem to me. I think in the regular case where solar wind flows in, we would not want the pml applied to the inflow boundary. Unless you set the inflow cells in front of the pml (essentially moving them inwards) but I am not sure this is a good idea.

ykempf: At some point we had discussed this parameter as you were possibly going to use the layer numbering for the PML. Indeed maybe the layers are not needed for PML business, but then here 3 can be put back and the PML magic happens independently of this parameter.

PR #466, opened 2020-03-09
Translation peformance

ykempf:
This is the result of performance optimisations tested on HazelHen before its shutdown.

One aspect (2-3x faster) was balancing load by translation mapping time, as timed per process then divided per cell. This introduces some non-locality of the LB weight which can have nasty side-effects, thus will have to be improved with further testing.

Another aspect thanks to  @galfthan's suggestions was optimising (aligned) allocations, most notably pulling them out of the intensive loops, and threading the pencil building.

This is now still a version with barriers in translation after each step, should I remove those?


ykempf:
This does the trick! Sadly I don't have access to the restart I was using at HLRS to test the performance gain in a proper evolved case.

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1206,7 +1247,7 @@ bool trans_map_1d_amr(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
 
                         // Unpack the vector data
                         Realf vector[VECL];
```
ykempf: Question @ursg @galfthan: do you expect a performance gain if this declaration was moved outside? SImilarly, there's a bunch of ints that get redeclared inside loops at every iteration. Is the compiler clever enough for these? Or is it less of an issue for small data types?

ursg: Yeah, the compiler should be clever enough. And even if not: declaration of an automatic variable like that on the stack does not actually go through the heap allocator, but just shifts the stack pointer by a couple of bytes. So this is literally only a single integer addition and does not even perform a memory access. Don't worry about it.

ykempf:
Urs please have a look at the last commit, to be sure.

MAKE/Makefile.hawk_gcc_mpt
```diff
@@ -0,0 +1,90 @@
...
+CXXFLAGS +=  -DPAPI_MEM
+
+
+#======== Allocator =========
+#Use jemalloc instead of system malloc to reduce memory fragmentation? https://github.com/jemalloc/jemalloc
+#Configure jemalloc with  --with-jemalloc-prefix=je_ when installing it
+CXXFLAGS += -DUSE_JEMALLOC -DJEMALLOC_NO_DEMANGLE
+
+
+#CXXFLAGS = -DMPICH_IGNORE_CXX_SEEK
```
markusbattarbee: I know this is a comment line but still: = vs +=

MAKE/Makefile.hawk_gcc_mpt
```diff
@@ -0,0 +1,90 @@
...
+# MATHFLAGS are for special math etc. flags, these are only applied on solver functions
+# LDFLAGS flags for linker
+
+#GNU flags:
+CC_BRAND = gcc
+CC_BRAND_VERSION = 9.2.0
+CXXFLAGS += -g -O3 -fopenmp -funroll-loops -std=c++17 -W -Wall -Wno-unused -fabi-version=0 -Wall -Wpedantic -mfma -march=native -mavx2
+#CXXFLAGS += -O3 -fopenmp -funroll-loops -std=c++17 -W -Wall -Wno-unused -fabi-version=0 -mavx
+not_parallel_tools: CXXFLAGS += -march=native -mno-avx2 -mavx 
+testpackage: CXXFLAGS = -g -O2 -fopenmp -funroll-loops -std=c++17 -fabi-version=0
```
markusbattarbee: here too

MAKE/Makefile.hawk_gcc_openmpi
```diff
@@ -0,0 +1,94 @@
...
+CXXFLAGS +=  -DPAPI_MEM
+
+
+#======== Allocator =========
+#Use jemalloc instead of system malloc to reduce memory fragmentation? https://github.com/jemalloc/jemalloc
+#Configure jemalloc with  --with-jemalloc-prefix=je_ when installing it
+CXXFLAGS += -DUSE_JEMALLOC -DJEMALLOC_NO_DEMANGLE
+
+
+#CXXFLAGS = -DMPICH_IGNORE_CXX_SEEK
```
markusbattarbee: ditto

MAKE/Makefile.hawk_gcc_openmpi
```diff
@@ -0,0 +1,94 @@
...
+# MATHFLAGS are for special math etc. flags, these are only applied on solver functions
+# LDFLAGS flags for linker
+
+#GNU flags:
+CC_BRAND = gcc
+CC_BRAND_VERSION = 9.2.0
+CXXFLAGS += -g -O3 -fopenmp -funroll-loops -std=c++17 -W -Wall -Wno-unused -fabi-version=0 -Wall -Wpedantic -mfma -march=native -mavx2
+#CXXFLAGS += -O3 -fopenmp -funroll-loops -std=c++17 -W -Wall -Wno-unused -fabi-version=0 -mavx
+not_parallel_tools: CXXFLAGS += -march=native -mno-avx2 -mavx 
+testpackage: CXXFLAGS = -g -O2 -fopenmp -funroll-loops -std=c++17 -fabi-version=0
```
markusbattarbee: ditto

MAKE/Makefile.hawk_intel_mpt
```diff
@@ -0,0 +1,88 @@
...
+CXXFLAGS +=  -DPAPI_MEM
+
+
+#======== Allocator =========
+#Use jemalloc instead of system malloc to reduce memory fragmentation? https://github.com/jemalloc/jemalloc
+#Configure jemalloc with  --with-jemalloc-prefix=je_ when installing it
+CXXFLAGS += -DUSE_JEMALLOC -DJEMALLOC_NO_DEMANGLE
+
+
+#CXXFLAGS = -DMPICH_IGNORE_CXX_SEEK
```
markusbattarbee: you guessed it

MAKE/Makefile.hawk_intel_openmpi
```diff
@@ -0,0 +1,88 @@
...
+CXXFLAGS +=  -DPAPI_MEM
+
+
+#======== Allocator =========
+#Use jemalloc instead of system malloc to reduce memory fragmentation? https://github.com/jemalloc/jemalloc
+#Configure jemalloc with  --with-jemalloc-prefix=je_ when installing it
+CXXFLAGS += -DUSE_JEMALLOC -DJEMALLOC_NO_DEMANGLE
+
+
+#CXXFLAGS = -DMPICH_IGNORE_CXX_SEEK
```
markusbattarbee: once more, with feeling

MAKE/Makefile.puhti_gcc
```diff
@@ -35,19 +35,14 @@ endif
...
 
 FLAGS = 
 
 #GNU flags:
 CC_BRAND = gcc
 CC_BRAND_VERSION = 9.1.0
-CXXFLAGS += -g -O3 -fopenmp -funroll-loops -std=c++17 -W -Wall -Wno-unused -fabi-version=0 -march=native -mavx2 -mfma
-testpackage: CXXFLAGS = -O2 -fopenmp -funroll-loops -std=c++17 -fabi-version=0  -mavx2 -mfma
+CXXFLAGS += -g -O3 -fopenmp -funroll-loops -std=c++17 -W -Wall -Wno-unused -fabi-version=0 -march=native -mfma -mavx512f
+testpackage: CXXFLAGS = -O2 -fopenmp -funroll-loops -std=c++17 -fabi-version=0 -mfma
```
markusbattarbee: here too

MAKE/Makefile.hawk_intel_mpt
ykempf: No, what?

MAKE/Makefile.hawk_intel_openmpi
ykempf: And each of these came as a separate email... ;)

MAKE/Makefile.puhti_gcc
ykempf: #metoo

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -884,13 +902,12 @@ bool checkPencils(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpi
    }
 
    for (int ipencil = 0; ipencil < pencils.N; ++ipencil) {
```
markusbattarbee: This loop goes over every pencil, and for each pencil takes every cell, and for each of those, checks over all id's within all pencils. Just for a sanity check. Sounds super-wasteful!

Plus, we actually already checked the count for each cell separately above.

Would it work to:
    for (auto id : cells) {
      if (mpiGrid[id]->sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY )  {
         int myCount = std::count(pencils.ids.begin(), pencils.ids.end(), id);
 use std::find to get first pencil in which this cell is used
 then use cint nPencilsThroughThisCell = pow(pow(2,pencils.path[ipencil].size()),2); on that pencil and perform the check

Of course this is a valid question only if this takes a non-negligible time to calculate, but it does seem like a fair bit of looping.

markusbattarbee: Also, this is called in a non-threaded section and could surely be threaded

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1097,103 +1147,94 @@ bool trans_map_1d_amr(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+      // declarations for variables needed by the threads
+      std::vector<Realf, aligned_allocator<Realf, 64>> targetBlockData((pencils.sumOfLengths + 2 * pencils.N) * WID3);
+      std::vector<std::vector<SpatialCell*>> pencilSourceCells;
+      
+      // Allocate aligned vectors which are needed once per pencil to avoid reallocating once per block loop + pencil loop iteration
+      std::vector<std::vector<Vec, aligned_allocator<Vec,64>>> pencilTargetValues;
+      std::vector<std::vector<Vec, aligned_allocator<Vec,64>>> pencilSourceVecData;
+      std::vector<std::vector<Vec, aligned_allocator<Vec,64>>> pencildz;
+      
+      for(uint pencili = 0; pencili < pencils.N; ++pencili) {
```
markusbattarbee: Do we allocate too much memory here, as this loop isn't threaded? I.e. enough memory to cover all threads for each thread separately? I guess not as each thread below is for a separate block, and these allocators are for one block each?

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1043,23 +1063,47 @@ bool trans_map_1d_amr(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+      }
+      
+      // accumulate thread results in global set of pencils
+#pragma omp critical
+      {
+         for (uint i=0; i<thread_pencils.N; i++) {
+            // Use vector range constructor
+            ibeg = thread_pencils.ids.begin() + thread_pencils.idsStart[i];
+            iend = ibeg + thread_pencils.lengthOfPencils[i];
+            std::vector<CellID> pencilIds(ibeg, iend);
```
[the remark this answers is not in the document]
markusbattarbee: Ah, probably could use something like
for (auto& idv : pencils.idsVectors)
    idsCount += std::count(idv.begin(), idv.end(), localPropagatedCells[i]);

vlasovsolver/vlasovmover.cpp
```diff
@@ -233,12 +305,44 @@ void calculateSpatialTranslation(
...
       phiprof::stop(profName);
    }
    
+   if (Parameters::prepareForRebalance == true) {
+      if(P::amrMaxSpatialRefLevel == 0) {
+//          const double deltat = (MPI_Wtime() - t1) / local_propagated_cells.size();
+         for (size_t c=0; c<localCells.size(); ++c) {
+//            mpiGrid[localCells[c]]->parameters[CellParams::LBWEIGHTCOUNTER] += time / localCells.size();
+            for (uint popID=0; popID<getObjectWrapper().particleSpecies.size(); ++popID) {
+               mpiGrid[localCells[c]]->parameters[CellParams::LBWEIGHTCOUNTER] += mpiGrid[localCells[c]]->get_number_of_velocity_blocks(popID);
```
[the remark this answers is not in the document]
ykempf: So let's discuss this. I added the value knowing full well it would swamp the (few tens of) millisecond timings of acceleration. Do we want to make a compound? We're dominated by translation by quite some margin at the moment. Anything based on (time spent in trans) is prone to the non-locality issues of Lb we had been theorising.

vlasovsolver/cpu_trans_map_amr.cpp
ykempf: The whole "buildPencils" group which includes the pencil-building and this check takes 0.5% of translation. Do we really want to fiddle here?

vlasovsolver/vlasovmover.cpp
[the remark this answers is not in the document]
markusbattarbee: The acc LBweight addition is commented out now - but if we're able to improve performance with MPI changes in the future, we might want to aggregate these two weights as they might be 50-50.

But you're right in that as long as we have MPI waiting time inside trans, it might end up being a bad metric. So when I get to looking at local ghost translation, I'll do the change then if it works out.

vlasovsolver/cpu_trans_map_amr.cpp
markusbattarbee: perhaps not.

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1043,23 +1063,47 @@ bool trans_map_1d_amr(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
```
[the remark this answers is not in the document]
ykempf: I move this to the next parallel region to avoid unnecessary overhead.

vlasovsolver/cpu_trans_map_amr.cpp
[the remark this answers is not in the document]
ykempf: I don't see what the issue is here?

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1071,13 +1115,19 @@ bool trans_map_1d_amr(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
       abort();
    }
    
-   // Add the final set of pencils to the pencilSets - vector.
-   // Only one set is created for now but we retain support for multiple sets
-   pencilSets.push_back(pencils);
+   if (Parameters::prepareForRebalance == true) {
+      for (uint i=0; i<localPropagatedCells.size(); i++) {
+         cuint myPencilCount = std::count(pencils.ids.begin(), pencils.ids.end(), localPropagatedCells[i]);
```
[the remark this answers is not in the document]
ykempf: This is done once every LB, so probably not worth storing and dragging along at all other times?

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1097,103 +1147,94 @@ bool trans_map_1d_amr(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
-//                cout << endl;
-               
-               // Allocate source data: sourcedata<length of pencil * WID3)
-               // Add padding by 2 * VLASOV_STENCIL_WIDTH
-               std::vector<Vec, aligned_allocator<Vec,64>> sourceVecData(sourceLength * WID3 / VECL);
-
+                              
                // load data(=> sourcedata) / (proper xy reconstruction in future)
-               copy_trans_block_data_amr(sourceCells.data(), blockGID, L, sourceVecData.data(),
+               bool pencil_has_data = copy_trans_block_data_amr(pencilSourceCells[pencili].data(), blockGID, L, pencilSourceVecData[pencili].data(),
```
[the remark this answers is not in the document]
ykempf: No we loop over the union of blocks existing on this process, and skip here if in a given pencil (e.g. solar wind) a block (e.g. hot sheath halo) is not present at all.

vlasovsolver/cpu_trans_map_amr.cpp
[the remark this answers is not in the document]
markusbattarbee: Hm, looks to me like this is done every single trans_map_amr call

vlasovsolver/cpu_trans_map_amr.cpp
ykempf: We are in the parallel region so each thread allocates its own of the three guys above.

ykempf: And they only need one each.

vlasovsolver/cpu_trans_map_amr.cpp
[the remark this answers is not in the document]
markusbattarbee: No real issue. Was thinking of performance but it's probably best left as-is.

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1097,103 +1147,94 @@ bool trans_map_1d_amr(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+   
+   phiprof::stop("setup");
+   
    int t1 = phiprof::initializeTimer("mapping");
    int t2 = phiprof::initializeTimer("store");
    
 #pragma omp parallel
    {
+      // declarations for variables needed by the threads
+      std::vector<Realf, aligned_allocator<Realf, 64>> targetBlockData((pencils.sumOfLengths + 2 * pencils.N) * WID3);
```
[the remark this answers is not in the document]
ykempf: WID3 agree.

vlasovsolver/vlasovmover.cpp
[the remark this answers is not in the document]
ykempf: So? Keep as is for now?

vlasovsolver/cpu_trans_map_amr.cpp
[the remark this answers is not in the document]
ykempf: `if (Parameters::prepareForRebalance == true) {` is only true, well, a step before LB when we are preparing for rebalancing. No?

vlasovsolver/cpu_trans_map_amr.cpp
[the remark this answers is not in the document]
markusbattarbee: Ok yes my bad.

vlasovsolver/cpu_trans_map_amr.cpp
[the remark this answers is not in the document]
markusbattarbee: Durr true, was looking too far behind!

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1097,103 +1147,94 @@ bool trans_map_1d_amr(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
    for(const auto blockGID:  unionOfBlocksSet) {
       unionOfBlocks.push_back(blockGID);
    }
+   phiprof::stop("buildBlockList");
    // ****************************************************************************
    
+   // Compute spatial neighbors for target cells.
+   // For targets we need the local cells, plus a padding of 1 cell at both ends
+   phiprof::start("computeSpatialTargetCellsForPencils");
+   std::vector<SpatialCell*> targetCells(pencils.sumOfLengths + pencils.N * 2 );
```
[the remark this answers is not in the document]
ykempf: Aah, now I see what you mean. Instead of 2 at the end? Yes it looks like that.

vlasovsolver/vlasovmover.cpp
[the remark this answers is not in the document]
markusbattarbee: Yep might as well.

ykempf:
So are we all resolved @markusbattarbee?

ykempf:
Or did I miss more? I got a single email for all your comments in cpu_trans_map_amr.cpp, now I saw there wqas more to it than the Makefile lines. :)

markusbattarbee:
Hm so on my behalf it's all ok. @ursg might want to chime in on that fsgrid restart thing still.

ykempf:
Great. I'll check on Hawk with a restart that I didn't do any last-minute blunders as the last pushes were all without compiling...

MAKE/Makefile.hawk_gcc_mpt
ykempf: No actually, I was suspecting that, this negates the testpackage flags as it only adds them on top of the ones before. We'll have to think this through more carefully.

MAKE/Makefile.hawk_gcc_openmpi
ykempf: re

MAKE/Makefile.puhti_gcc
ykempf: re

PR #467, opened 2020-03-11
Re-implemented check for cubical cells

markusbattarbee:
It compiles!


ursg:
However: this pull request is set up to merge into master. Shouldn't it go to dev instead?

PR #472, opened 2020-04-02
Bring dev up to date with master

ursg:
This includes recent translation performance updates and vorna makefiles that had accidentally been commited to master instead of dev.


ykempf:
Well I suppose this is more of a formality or do you want formal reviews?

PR #473, opened 2020-04-02
Shells at selected radii where VDFs should be saved

maximegrandin:
This gives the possibility to add a new parameter to the cfg file
     _io.system_write_distribution_shell_radius_
which defines one or several spheres of selected radii (in m) where all VDFs should be saved in output files.
This has been tested with two shells at 5e7 m and 2e8 m in a test-run with 1 level of refinement. The results are shown in the attached screenshots (threeSlice, and three slices).

threeSlice_fsaved_with_SpatialGrid_mesh
slice_XY
slice_YZ
slice_XZ

### DIFF
--- a/iowrite.cpp
+++ b/iowrite.cpp
@@ -934,6 +934,9 @@ bool writeVelocitySpace(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpi
       //Compute which cells will write out their velocity space
       vector<uint64_t> velSpaceCells;
       int lineX, lineY, lineZ;
+      Real shellRadiusSquare;
+      Real cellX, cellY, cellZ, halfDX, halfDY, halfDZ;
+      Real minRCornerSquare,maxRCornerSquare,rCornerSquare;
       for (uint i = 0; i < cells.size(); i++) {
          mpiGrid[cells[i]]->parameters[CellParams::ISCELLSAVINGF] = 0.0;
          // CellID stride selection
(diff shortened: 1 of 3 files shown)


ykempf:
Yes, it should skip since the vector size would be zero. Good to go for me.

markusbattarbee:
How many cells does this add to the bulk files? For comparison, in EGE, we had 2258 cells with saved VDFs. I'm thinking if the value goes up a lot, it may get expensive. Just print the sum of vg_f_saved to find out.

You won't see it in a fresh simulation from the bulk file size but hundreds of hot magnetosheath VDFs  later in the simulation adds up. You could do a quick restart of EGE / EGA with this branch to verify.

A way around this would be to add an optional modulo-stride parameter (scalar or vector) so that for each spherical shell, all / every 2nd / every 3rd / etc cells are saved. Just make a check which checks if cellid % stride == 0.

maximegrandin:
> How many cells does this add to the bulk files? For comparison, in EGE, we had 2258 cells with saved VDFs. I'm thinking if the value goes up a lot, it may get expensive. Just print the sum of vg_f_saved to find out.
The count gives 3684 in this example, which is a +63% increase.

> A way around this would be to add an optional modulo-stride parameter (scalar or vector) so that for each spherical shell, all / every 2nd / every 3rd / etc cells are saved. Just make a check which checks if cellid % stride == 0.
I can add this, yes, this would indeed give extra flexibility in the scheme.

iowrite.cpp
```diff
@@ -978,9 +981,56 @@ bool writeVelocitySpace(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpi
...
+            if (rCornerSquare < minRCornerSquare) {minRCornerSquare = rCornerSquare;}
+            // Eighth corner
+            rCornerSquare = (cellX + DX) * (cellX + DX) + (cellY + DY) * (cellY + DY) + (cellZ + DZ) * (cellZ + DZ);
+            if (rCornerSquare > maxRCornerSquare) {maxRCornerSquare = rCornerSquare;}
+            if (rCornerSquare < minRCornerSquare) {minRCornerSquare = rCornerSquare;}
+            if (minRCornerSquare <= shellRadiusSquare && maxRCornerSquare > shellRadiusSquare) {
+               velSpaceCells.push_back(cells[i]);
+               mpiGrid[cells[i]]->parameters[CellParams::ISCELLSAVINGF] = 1.0;
+            }
+         }
```
markusbattarbee, line 1033: Lots of ifs. How about:
dx_rm = cellX < 0 ? DX : 0; 
dx_rp = cellX < 0 ? 0 : DX; 
dy_rm = cellY < 0 ? DY : 0; 
dy_rp = cellY < 0 ? 0 : DY; 
dz_rm = cellZ < 0 ? DZ : 0; 
dz_rp = cellZ < 0 ? 0 : DZ;
rsquare_minus = (cellX+dx_rm)*(cellX+dx_rm) + (cellY+dy_rm)*(cellY+dy_rm) + (cellZ+dz_rm)*(cellZ+dz_rm);
rsquare_plus = (cellX+dx_rp)*(cellX+dx_rp) + (cellY+dy_rp)*(cellY+dy_rp) + (cellZ+dz_rp)*(cellZ+dz_rp);
if (rsquare_minus <= shellRadiusSquare && rsquare_plus > shellRadiusSquare) {
               velSpaceCells.push_back(cells[i]);
               mpiGrid[cells[i]]->parameters[CellParams::ISCELLSAVINGF] = 1.0;
            }

maximegrandin: Implemented as suggested.

parameters.cpp
```diff
@@ -374,6 +380,13 @@ bool Parameters::getParameters(){
       }
       return false;
    }
+   if ( P::systemWriteDistributionWriteShellStride.size() != P::systemWriteDistributionWriteShellRadius.size()) {
+      if(myRank == MASTER_RANK) {
+         cerr << "ERROR io.system_write_distribution_shell_stride should have the same size as " <<
```
ykempf, line 385: This is nitpicking but it would be clearer to the user if it said something like "You should set the same number of io.system_write_distribution_shell_stride and io.system_write_distribution_shell_radius".

ykempf:
That's the idea. We'll have to be careful not to make the files too massive but that's operational wisdom. :)

PR #476, opened 2020-04-07
Outflow fix

ykempf:
It compiles and runs at least. Any comments?

### DIFF
--- a/fieldsolver/ldz_magnetic_field.cpp
+++ b/fieldsolver/ldz_magnetic_field.cpp
@@ -185,13 +185,12 @@ void propagateSysBoundaryMagneticField(
    cint& RKCase,
    cuint component
 ) {
-   std::array<Real, fsgrids::bfield::N_BFIELD> * bGrid;
+   FsGrid< std::array<Real, fsgrids::bfield::N_BFIELD>, 2> * bGrid;
    if (RKCase == RK_ORDER1 || RKCase == RK_ORDER2_STEP2) {
-      bGrid = perBGrid.get(i,j,k);
+      perBGrid.get(i,j,k)->at(fsgrids::bfield::PERBX + component) = sysBoundaries.getSysBoundary(technicalGrid.get(i,j,k)->sysBoundaryFlag)->fieldSolverBoundaryCondMagneticField(perBGrid, technicalGrid, i, j, k, dt, component);
    } else {
-      bGrid = perBDt2Grid.get(i,j,k);
+      perBDt2Grid.get(i,j,k)->at(fsgrids::bfield::PERBX + component) = sysBoundaries.getSysBoundary(technicalGrid.get(i,j,k)->sysBoundaryFlag)->fieldSolverBoundaryCondMagneticField(perBDt2Grid, technicalGrid, i, j, k, dt, component);
    }
-   bGrid->at(fsgrids::bfield::PERBX + component) = sysBoundaries.getSysBoundary(technicalGrid.get(i,j,k)->sysBoundaryFlag)->fieldSolverBoundaryCondMagneticField(perBGrid, perBDt2Grid, EGrid, EDt2Grid, technicalGrid, i, j, k, dt, RKCase, component);
 }
 
 /*! \brief High-level magnetic field propagation function.
--- a/sysboundary/ionosphere.cpp
+++ b/sysboundary/ionosphere.cpp
@@ -535,53 +535,41 @@ namespace SBC {
     */
    Real Ionosphere::fieldSolverBoundaryCondMagneticField(
       FsGrid< std::array<Real, fsgrids::bfield::N_BFIELD>, 2> & perBGrid,
-      FsGrid< std::array<Real, fsgrids::bfield::N_BFIELD>, 2> & perBDt2Grid,
-      FsGrid< std::array<Real, fsgrids::efield::N_EFIELD>, 2> & EGrid,
-      FsGrid< std::array<Real, fsgrids::efield::N_EFIELD>, 2> & EDt2Grid,
       FsGrid< fsgrids::technical, 2> & technicalGrid,
       cint i,
       cint j,
       cint k,
       creal& dt,
-      cuint& RKCase,
       cuint& component
    ) {
-      FsGrid< std::array<Real, fsgrids::bfield::N_BFIELD>, 2> * bGrid;
-      
-      if(RKCase == RK_ORDER1 || RKCase == RK_ORDER2_STEP2) {
-         bGrid = &perBGrid;
-      } else {
-         bGrid = &perBDt2Grid;
-      }
-      
       if (technicalGrid.get(i,j,k)->sysBoundaryLayer == 1) {
          switch(component) {
             case 0:
                if (  ((technicalGrid.get(i-1,j,k)->SOLVE & compute::BX) == compute::BX)
                   && ((technicalGrid.get(i+1,j,k)->SOLVE & compute::BX) == compute::BX)
                ) {
-                  return 0.5 * (bGrid->get(i-1,j,k)->at(fsgrids::bfield::PERBX) + bGrid->get(i+1,j,k)->at(fsgrids::bfield::PERBX));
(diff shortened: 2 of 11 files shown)


ykempf:
Not sure why f9e026d shows up here, it doesn't in the file diffs.

sysboundary/outflow.cpp
```diff
@@ -334,25 +334,27 @@ namespace SBC {
...
       cuint& component
    ) {
-      Real fieldValue;
-      
-      if(RKCase == RK_ORDER1 || RKCase == RK_ORDER2_STEP2) {
-         fieldValue = fieldBoundaryCopyFromExistingFaceNbrMagneticField(perBGrid, technicalGrid, i, j, k, component);
-      } else {
-         fieldValue = fieldBoundaryCopyFromExistingFaceNbrMagneticField(perBDt2Grid, technicalGrid, i, j, k, component);
+      switch(component) {
+      case compute::BX:
```
ykempf, line 345: Wrong way of doing!

sysboundary/sysboundarycondition.cpp
```diff
@@ -785,48 +785,54 @@ namespace SBC {
...
-      if (closestCell[0] == std::numeric_limits<int>::min()) {
-         //cerr << "(" << gid[0] << "," << gid[1] << "," << gid[2] << ")" << __FILE__ << ":" << __LINE__ << ": No closest cell found!" << endl;
-         //abort();
+      int distance = std::numeric_limits<int>::max();
+      std::vector< std::array<int,3> > closestCells;
+
+      for (int kk=-2; kk<3; kk++) {
+         for (int jj=-2; jj<3; jj++) {
+            for (int ii=-2; ii<3 ; ii++) {
+               if( technicalGrid.get(i+ii,j+jj,k+kk) && (technicalGrid.get(i+ii,j+jj,k+kk)->SOLVE & mask) == mask) {
```
ykempf, line 804: Not enough. Skip oneself as well.

sysboundary/sysboundarycondition.cpp
```diff
@@ -785,48 +785,60 @@ namespace SBC {
...
+      int distance = std::numeric_limits<int>::max();
+      std::vector< std::array<int,3> > closestCells;
 
+      for (int kk=-2; kk<3; kk++) {
+         for (int jj=-2; jj<3; jj++) {
+            for (int ii=-2; ii<3 ; ii++) {
+               if( technicalGrid.get(i+ii,j+jj,k+kk) // skip invalid cells returning NULL
+                   && (technicalGrid.get(i+ii,j+jj,k+kk)->SOLVE & mask) == mask // Did that guy solve this component?
+                   && technicalGrid.get(i+ii,j+jj,k+kk)->sysBoundaryFlag != sysboundarytype::DO_NOT_COMPUTE // Do not copy from there
+                   && ii != 0 && jj != 0 && kk != 0 // skip oneself!
```
ykempf, line 807: Of this logic, I added also the last line but it is unnecessary as a cell calling this function wouldn't match the mask condition. But something (here or elsewhere) is flawed as I don't get the correct copied values e.g. for fg_per_b_x at the -z edge.

sysboundary/ionosphere.cpp
```diff
@@ -535,61 +535,49 @@ namespace SBC {
...
-      }
-      
       if (technicalGrid.get(i,j,k)->sysBoundaryLayer == 1) {
          switch(component) {
             case 0:
                if (  ((technicalGrid.get(i-1,j,k)->SOLVE & compute::BX) == compute::BX)
                   && ((technicalGrid.get(i+1,j,k)->SOLVE & compute::BX) == compute::BX)
                ) {
-                  return 0.5 * (bGrid->get(i-1,j,k)->at(fsgrids::bfield::PERBX) + bGrid->get(i+1,j,k)->at(fsgrids::bfield::PERBX));
+                  return 0.5 * (perBGrid.get(i-1,j,k)->at(fsgrids::bfield::PERBX) + perBGrid.get(i+1,j,k)->at(fsgrids::bfield::PERBX));
```
markusbattarbee, line 551: Here and on lines below it'd actually be better to keep it as bGrid, as it can in fact be perBGrid or perBDt2Grid, right?

And make the grid received in the function call be bGrid, of course.

sysboundary/sysboundarycondition.cpp
```diff
@@ -785,48 +785,60 @@ namespace SBC {
...
-   Real SysBoundaryCondition::fieldBoundaryCopyFromExistingFaceNbrMagneticField(
+   Real SysBoundaryCondition::fieldBoundaryCopyFromSolvingNbrMagneticField(
       FsGrid< std::array<Real, fsgrids::bfield::N_BFIELD>, 2> & perBGrid,
       FsGrid< fsgrids::technical, 2> & technicalGrid,
       cint i,
       cint j,
       cint k,
-      cuint component
+      cuint component,
+      cuint mask
```
markusbattarbee, line 795: Since mask depends directly on component we don't need it as an input parameter, it can be extracted in this function.

ykempf:
This latest version I think is close to being good. I'll do some last checks later tonight. Feel free to review already.

sysboundary/sysboundarycondition.cpp
markusbattarbee: So what was the thing here? Make clean? :)

fieldsolver/ldz_magnetic_field.cpp
```diff
@@ -185,13 +185,12 @@ void propagateSysBoundaryMagneticField(
    cint& RKCase,
    cuint component
 ) {
-   std::array<Real, fsgrids::bfield::N_BFIELD> * bGrid;
+   FsGrid< std::array<Real, fsgrids::bfield::N_BFIELD>, 2> * bGrid;
```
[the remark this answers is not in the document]
ykempf: Leftover, will be removed, thanks.

sysboundary/sysboundarycondition.cpp
```diff
@@ -785,48 +785,60 @@ namespace SBC {
...
+      for (int kk=-2; kk<3; kk++) {
+         for (int jj=-2; jj<3; jj++) {
+            for (int ii=-2; ii<3 ; ii++) {
+               if( technicalGrid.get(i+ii,j+jj,k+kk) // skip invalid cells returning NULL
+                   && (technicalGrid.get(i+ii,j+jj,k+kk)->SOLVE & mask) == mask // Did that guy solve this component?
+                   && technicalGrid.get(i+ii,j+jj,k+kk)->sysBoundaryFlag != sysboundarytype::DO_NOT_COMPUTE // Do not copy from there
+                   && ii != 0 && jj != 0 && kk != 0 // skip oneself!
+               ) {
+                  int d = ii*ii + jj*jj + kk*kk;
+                  if( d == distance ) {
```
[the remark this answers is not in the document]
ykempf: - Right.
- No, it's not a vector it's a struct.

sysboundary/ionosphere.cpp
ykempf: That'd make this diff easier to follow indeed. Thanks.

sysboundary/sysboundarycondition.cpp
ykempf: I'll look that up. Possibly. But we have more than B in compute::, so we have to be careful.

ykempf:
Thread-safety hasn't been changed by this. So, yes, it is (still) safe, or at least as safe as it was already.

sysboundary/sysboundarycondition.cpp
markusbattarbee: ping @ykempf

sysboundary/sysboundarycondition.cpp
ykempf: Which was incorrect.

sysboundary/sysboundarycondition.cpp
ykempf: No, removing the last line of checks as that would trigger whenever ii, jj or kk was zero, so copying from any but face neighbours.

markusbattarbee: ahh right :)

markusbattarbee:
Nitpick reminder but you're using ykempf:dev for this instead of a feature branch

PR #479, opened 2020-04-15
Added zeroing of inflow cells for perB with vector dipole (type 4)

markusbattarbee:
Now it sets type maxwellian cell perB-fields to zero if starting a fresh run. I guess this could be done for all dipole types (for all others it should already be zero) but playing it safe. Due to the way SOLVE::BXYZ works I think the fields should not be reset at a restart so added that check.

### DIFF
--- a/projects/Magnetosphere/Magnetosphere.cpp
+++ b/projects/Magnetosphere/Magnetosphere.cpp
@@ -464,6 +464,9 @@ namespace projects {
                      if(technicalGrid.get(x, y, z)->sysBoundaryFlag == sysboundarytype::SET_MAXWELLIAN ) {
                         for (int i = 0; i < fsgrids::bgbfield::N_BGB; ++i) {
                            BgBGrid.get(x,y,z)->at(i) = 0;
+			   if ( (this->dipoleType==4) && (P::isRestart == false) ) {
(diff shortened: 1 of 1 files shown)


projects/Magnetosphere/Magnetosphere.cpp
```diff
@@ -464,6 +464,9 @@ namespace projects {
                      if(technicalGrid.get(x, y, z)->sysBoundaryFlag == sysboundarytype::SET_MAXWELLIAN ) {
                         for (int i = 0; i < fsgrids::bgbfield::N_BGB; ++i) {
                            BgBGrid.get(x,y,z)->at(i) = 0;
+			   if ( (this->dipoleType==4) && (P::isRestart == false) ) {
+			      perBGrid.get(x,y,z)->at(i) = 0;
+			   }
```
ykempf, line 469: No, this will break spectacularly. You are accessing perBGrid with indices over fsgrid::bgbfield. Unless they have the same size?

ykempf: This will go haywire, perBGrid has 3 entries, bgBGrid has 17 or so.

markusbattarbee: Ah whoops, too coding fingers were too quick for my brain

PR #483, opened 2020-04-27
Make writeAsFloat more flexible for fsgrids, don't float them in restart

ursg:
Before, their behaviour was globally defined in the P::writeAsFloat
variable, now it can be changed on a call-by-call manner.

### DIFF
--- a/datareduction/datareducer.cpp
+++ b/datareduction/datareducer.cpp
@@ -1196,8 +1196,8 @@ bool DataReducer::writeFsGridData(
                       FsGrid< std::array<Real, fsgrids::dmoments::N_DMOMENTS>, 2>& dMomentsGrid,
                       FsGrid< std::array<Real, fsgrids::bgbfield::N_BGB>, 2>& BgBGrid,
                       FsGrid< std::array<Real, fsgrids::volfields::N_VOL>, 2>& volGrid,
-                      FsGrid< fsgrids::technical, 2>& technicalGrid, const
-                      std::string& meshName, const unsigned int operatorID,
+                      FsGrid< fsgrids::technical, 2>& technicalGrid,
+                      const std::string& meshName, const unsigned int operatorID,
                       vlsv::Writer& vlsvWriter,
                       const bool writeAsFloat) {
    
--- a/datareduction/datareducer.cpp
+++ b/datareduction/datareducer.cpp
@@ -1196,13 +1196,16 @@ bool DataReducer::writeFsGridData(
                       FsGrid< std::array<Real, fsgrids::dmoments::N_DMOMENTS>, 2>& dMomentsGrid,
                       FsGrid< std::array<Real, fsgrids::bgbfield::N_BGB>, 2>& BgBGrid,
                       FsGrid< std::array<Real, fsgrids::volfields::N_VOL>, 2>& volGrid,
-                      FsGrid< fsgrids::technical, 2>& technicalGrid, const std::string& meshName, const unsigned int operatorID, vlsv::Writer& vlsvWriter) {
+                      FsGrid< fsgrids::technical, 2>& technicalGrid, const
+                      std::string& meshName, const unsigned int operatorID,
+                      vlsv::Writer& vlsvWriter,
+                      const bool writeAsFloat) {
    
    if (operatorID >= operators.size()) return false;
    DRO::DataReductionOperatorFsGrid* DROf = dynamic_cast<DRO::DataReductionOperatorFsGrid*>(operators[operatorID]);
    if(!DROf) {
       return false;
    } else {
-      return DROf->writeFsGridData(perBGrid, EGrid, EHallGrid, EGradPeGrid, momentsGrid, dPerBGrid, dMomentsGrid, BgBGrid, volGrid, technicalGrid, meshName, vlsvWriter);
+      return DROf->writeFsGridData(perBGrid, EGrid, EHallGrid, EGradPeGrid, momentsGrid, dPerBGrid, dMomentsGrid, BgBGrid, volGrid, technicalGrid, meshName, vlsvWriter, writeAsFloat);
    }
 }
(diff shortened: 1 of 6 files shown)


datareduction/datareducer.cpp
```diff
@@ -1196,13 +1196,16 @@ bool DataReducer::writeFsGridData(
                       FsGrid< std::array<Real, fsgrids::dmoments::N_DMOMENTS>, 2>& dMomentsGrid,
                       FsGrid< std::array<Real, fsgrids::bgbfield::N_BGB>, 2>& BgBGrid,
                       FsGrid< std::array<Real, fsgrids::volfields::N_VOL>, 2>& volGrid,
-                      FsGrid< fsgrids::technical, 2>& technicalGrid, const std::string& meshName, const unsigned int operatorID, vlsv::Writer& vlsvWriter) {
+                      FsGrid< fsgrids::technical, 2>& technicalGrid, const
+                      std::string& meshName, const unsigned int operatorID,
+                      vlsv::Writer& vlsvWriter,
```
ykempf, line 1201: Interesting line break pattern...

datareduction/datareducer.h
```diff
@@ -68,7 +68,10 @@ class DataReducer {
                       FsGrid< std::array<Real, fsgrids::dmoments::N_DMOMENTS>, 2>& dMomentsGrid,
                       FsGrid< std::array<Real, fsgrids::bgbfield::N_BGB>, 2>& BgBGrid,
                       FsGrid< std::array<Real, fsgrids::volfields::N_VOL>, 2>& volGrid,
-                      FsGrid< fsgrids::technical, 2>& technicalGrid, const std::string& meshName, const unsigned int operatorID, vlsv::Writer& vlsvWriter);
+                      FsGrid< fsgrids::technical, 2>& technicalGrid, const
```
ykempf, line 71: Also, why is const up there?

datareduction/datareductionoperator.cpp
```diff
@@ -143,7 +143,9 @@ namespace DRO {
                       FsGrid< std::array<Real, fsgrids::dmoments::N_DMOMENTS>, 2>& dMomentsGrid,
                       FsGrid< std::array<Real, fsgrids::bgbfield::N_BGB>, 2>& BgBGrid,
                       FsGrid< std::array<Real, fsgrids::volfields::N_VOL>, 2>& volGrid,
-                      FsGrid< fsgrids::technical, 2>& technicalGrid, const std::string& meshName, vlsv::Writer& vlsvWriter) {
+                      FsGrid< fsgrids::technical, 2>& technicalGrid, const
```
ykempf, line 146: I see a pattern there.

datareduction/datareductionoperator.h
```diff
@@ -134,7 +134,9 @@ namespace DRO {
                       FsGrid< std::array<Real, fsgrids::dmoments::N_DMOMENTS>, 2>& dMomentsGrid,
                       FsGrid< std::array<Real, fsgrids::bgbfield::N_BGB>, 2>& BgBGrid,
                       FsGrid< std::array<Real, fsgrids::volfields::N_VOL>, 2>& volGrid,
-                      FsGrid< fsgrids::technical, 2>& technicalGrid, const std::string& meshName, vlsv::Writer& vlsvWriter);
+                      FsGrid< fsgrids::technical, 2>& technicalGrid, const
```
ykempf, line 137: Check

markusbattarbee:
datareductionoperator.cpp line 117 still uses the P:: and I guess it shouldn't as restart writing also uses DROs?

ykempf:
https://github.com/fmihpc/vlasiator/blob/662b7610f888dbb3358a0583cb71bc727418b325/iowrite.cpp#L1416
For restarts this is set here, correctly, but Markus' comment on datareductionoperator.cpp line 117 holds. We have to set that variable differently. Pass the float or double to the constructor? Or reset the class' datasize variable at write time somehow?

ursg:
> datareductionoperator.cpp line 117 still uses the P:: and I guess it shouldn't as restart writing also uses DROs?

Oh indeed, it does. This value ends up never being *used* afterall (as the reducer writes its data itself, so the getVectorDataInfo information simply gets discarded) but this is simply confusing code.

datareduction/datareducer.cpp
ursg: Whoopsie. I screwed that up consistently it seems. Fixed.

ykempf:
The simplest is the testpackage restart test I supose.

ykempf:
Btw I suggest to abort the run if the reading of any restart field fails, instead of just printing to logfile. As I missed that for several restarts...

ursg:
> Where is P::writeAsFloat checked now? I can't see it.

The writeFsGrid method of the reducer gets it passed as a parameter. The actual testing of the parameter, and passing down into the function happens here: https://github.com/fmihpc/vlasiator/blob/e21380482e7e5c4ac9c4984015d0935bcff27c90/iowrite.cpp#L1189

Whereas fore restarts, it is not being checked:
https://github.com/fmihpc/vlasiator/blob/e21380482e7e5c4ac9c4984015d0935bcff27c90/iowrite.cpp#L1416

ykempf:
Well then 2 testpackage restart tests with a general writeAsFloat true and false are needed next.

ursg:
> Btw I suggest to abort the run if the reading of any restart field fails, instead of just printing to logfile. As I missed that for several restarts...

Indeed. The code to test for it was there, but the status variable was overwritten right after... A check has now been addad.

ioread.cpp
```diff
@@ -1180,8 +1180,9 @@ bool exec_readGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
       exitOnError(false, "(RESTART) FSGrid writing rank number not found in restart file", MPI_COMM_WORLD);
    }
    
-   success = readFsGridVariable(file, "fg_PERB", fsgridInputRanks, perBGrid);
-   success = readFsGridVariable(file, "fg_E", fsgridInputRanks, EGrid);
+   if(success) { success = readFsGridVariable(file, "fg_PERB", fsgridInputRanks, perBGrid); }
+   if(success) { success = readFsGridVariable(file, "fg_E", fsgridInputRanks, EGrid); }
+   exitOnError(success,"(RESTART) Failure reading fsgrid restart variables",MPI_COMM_WORLD);
```
ykempf, line 1185: Yes, that's how to do it.

ursg:
I tested this with the restart_write / restart_write testpackage combination now, with both io.write_as_float set and unset. Restart files are now written as double in all cases. Bulk files honour the parameter.

ykempf:
No diffs downstream?

ykempf:
Ah but float vs double diffs...

PR #484, opened 2020-05-06
Profile fsgrid read

ykempf:


ykempf:
I have no idea why it shows a pile of commits wrongly but 1 file changed correctly.

ursg:
That is indeed weird. Was that based on some occasionally-updated feature branch on your side that had commits in different order?

Anyway, the commits themselves seem reasonable!

PR #485, opened 2020-05-11
Hawk threading

ykempf:
I threaded some bottlenecks of translation initialisation. With testpackage flags and the translation barriers active, the gain is 1% in translation.

image

What I did:
- Thread a loop in `check_ghost_cells()`;
- Thread `pencils.split()` as that is called in `check_ghost_cells()`; these two were the largest factor in setting up translation;
- Thread `buildBlockList` as that was the other slow part of trans initialisation but not as bad as the above.

Remind me to disable the trans barriers again before merging this.


markusbattarbee:
This PR also needs to be carefully checked so that all necessary variables (including structs!) are properly threadprivate.

markusbattarbee:
vlasovsolver/cpu_trans_map_amr.cpp:1224:7: error: invalid controlling predicate
       for (vector<SpatialCell*>::iterator cell = allCellsPointer.begin(); cell != allCellsPointer.end(); cell++) {
       ^~~
Makefile:408: recipe for target 'cpu_trans_map_amr.o' failed
make: *** [cpu_trans_map_amr.o] Error 1

markusbattarbee:
apparently != is not a supported predicate in openMP loops

ykempf:
But... this did compile at some point.

ursg:
In that case, just replace the loop by a simpler array iteration?
```
for(unsigned int i=0; i<allCellsPointer.size(); i++) {
   auto cell = &allCellsPointer[i];
   // ...
```

ykempf:
I guess so, yes.

ykempf:
Mildly worrisome if changing "something" brings errors which it didn't appear earlier – was it wrong allalong and doing BS anyway?

ykempf:
So do we fix it in this branch still or are you fixing that while merging that too into dccrg_update2?

markusbattarbee:
Can merge it into dccrg_update2 once I've tested it. Dunno why it compiled for you earlier, but Urs' version was indeed palatable.

ykempf:
So are you committing it?

PR #487, opened 2020-05-22
Flipped construction of target cells in cpu_trans_map_amrp.cpp to wor…

markusbattarbee:
…k for long pencils


ykempf:
First test gives negligible diffs, this looks quite good!

markusbattarbee:
So this is wrong, I got confused with the order in which pencils were built and in which vectors were handled.

PR #488, opened 2020-05-26
Dccrg update

markusbattarbee:
Updates Vlasiator to work with DCCRG master (as of 26th May 2020)

Ran testpackage
Mostly zero diffs

Magnetosphere_polar_small
proton/vg_v_nonthermal_0                4.65661e-10                 5.82336e-16    
proton/vg_v_nonthermal_1                1.09139e-11                 1.86547e-16    
proton/vg_v_nonthermal_2                1.16415e-10                 2.44615e-16    
proton/vg_ptensor_nonthermal_diagonal_0                2.22143e-27                 8.72321e-16    
proton/vg_ptensor_nonthermal_diagonal_1                1.61559e-27                 3.21758e-16

### DIFF
--- a/vlasovsolver/cpu_trans_map_amr.cpp
+++ b/vlasovsolver/cpu_trans_map_amr.cpp
@@ -216,59 +216,109 @@ void computeSpatialSourceCellsForPencilWithFaces(const dccrg::Dccrg<SpatialCell,
    ngh_front.push_back(ids.front());
    ngh_back.push_back(ids.back());
    for (int ngh_i = 0; ngh_i < VLASOV_STENCIL_WIDTH; ++ngh_i) {
-     const auto frontNeighbors = mpiGrid.get_face_neighbors_of(ngh_front[0]);
-     refLvl = mpiGrid.get_refinement_level(ngh_front[0]);
-     if (frontNeighbors.size()>0) {
+     const auto frontNeighbors = mpiGrid.get_face_neighbors_of(ngh_front.front());
+     refLvl = mpiGrid.get_refinement_level(ngh_front.front());
+     if (frontNeighbors.size()!=NULL) {
        for (const auto nbr: frontNeighbors) {
 	 if(nbr.second == ((int)dimension + 1)) {
 	   neighbors.push_back(nbr.first);
 	 }
        }
-       if (neighbors.size() == 0) {
-	 std::cerr<<"abort front neighbors.size() == 0 at "<<ngh_front[0]<<std::endl;
-	 for( const auto nbrPair: frontNeighbors ) {
-	   std::cerr<<ngh_front[0]<<" dim "<<dimension<<" "<<nbrPair.first<<" "<<nbrPair.second<<std::endl;
-	 }
-	}
        if (neighbors.size() == 1) {
-	 sourceCells[VLASOV_STENCIL_WIDTH+L+ngh_i] = mpiGrid[neighbors.at(0)];
-	 ngh_front.push_back(neighbors.at(0));
+	 if (neighbors.at(0)!=NULL) {
+	   sourceCells[VLASOV_STENCIL_WIDTH+L+ngh_i] = mpiGrid[neighbors.at(0)];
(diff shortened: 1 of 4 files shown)


ykempf:
Surprisingly little changes then. Any insight about how stable this is now in DCCRG? Will it change again any time soon?

ykempf:
Little birds confirm this will be stable, good.
But: should there not be some more elaborate AMR test, as this is used only in AMR code (except the () addition)?

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1364,8 +1364,11 @@ int get_sibling_index(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGr
    if (parent == INVALID_CELLID) {
       return ERROR;
    }
-   
-   vector<CellID> siblings = mpiGrid.get_all_children(parent);
+
+   // get_all_children returns an array instead of a vector now, need to map it to a vector for find and distance
+   std::array<uint64_t, 8> siblingarr = mpiGrid.mapping.get_all_children(parent);
+   vector<CellID> siblings(&siblingarr[0], &siblingarr[0] + sizeof(siblingarr) / sizeof(siblingarr[0]));
```
ursg, line 1370: That took me a moment to parse. Wouldn't ```siblangarr.begin(), siblingarr.end()``` fulfill the same function and be more understandable?

markusbattarbee: Well, perhaps &siblingarr[0] can be replaced with siblingarr.begin()
but I tried
    vector<CellID> siblings(siblingarr.begin(),siblingarr.end());
and it gives major diffs. All the examples of generating a vector from an array used the method I used, apparently for a good reason.

markusbattarbee: replacing &siblingarr[0] with siblingarr.begin() also gave diffs.

markusbattarbee: > Little birds confirm this will be stable, good.
> But: should there not be some more elaborate AMR test, as this is used only in AMR code (except the () addition)?

Yeah, AMR testing underway and causing funny stuff. Don't merge yet.

ursg: Bizarre. I guess my C++ kungfoo is weak today. Well, leave it as it is then.

markusbattarbee:
In AMR tests, there are some diffs. My theory:

I took regular Vlasiator dev and added a error check output here:
https://github.com/fmihpc/vlasiator/compare/dev...markusbattarbee:dev_translation_printouts

and got lots of warnings. I think this is triggered in this case:

---------------------------------------------------------------
|      L0        |      *L0*      |  *L1*  |   L1   |    L1   |
---------------------------------------------------------------

The starred L1 cell looks stencils cells from a 2-cell neighborhood, which includes the non-starred L1-cells and the starred L0 cell. However, if I understand correctly, it doesn't include the non-starred L0 cell. Thus, in the negative direction, it finds only one cell to put in the vlasov stencil source cells array, and replaces the NULL value (which should point to the non-starred L0) with a copy of the starred L0.

In my test to make Vlasiator work with DCCRG master, I switched to an iterative method for going through face neighbors, which should activate the non-starred source cell L0 correctly. In this one, I did not get warnings for a lack of neighbors for non-sysboundary cells. This would explain the diffs I was getting, and would classify as a bugfix.

Thoughts?

Still working on redoing some testpackage runs to be absolutely sure of what I'm seeing, though.

ursg:
Interesting. Do we still have the mini-app that was doing a 1-d propagation? I remember dimly from back then, that is tended to cause some artefacts lagging behind the gauss peak that it transported through, and now I'm wondering if this would fix those.

ykempf:
I believe the miniapps are stored in that directory in the repo.

markusbattarbee:
Yep vlasiator/mini-apps/translation_pencils/trans_test.cpp  could be used to check this.

PR #491, opened 2020-06-23
Egi filtering- Using addComposing to parse number of passes

kstppd:


kstppd:
I am thinking that I should use a multimap for the number of passes for each individual ref level. It would make it more foolproof and safer. Opinions? @ykempf  @markusbattarbee @ursg

markusbattarbee:
As per Flowdock discussion, I think a vector would be good. I thought boost program options would have had simple support for a vector of ints as a single value, but not so sure of that...? Found some results on stackoverflow that show convoluted methods to achieve that.

Suggestions:
Do a reverse sort of the blurpass levels so that if the user puts them in L3->L0 order it gets corrected:
    sort(a.begin(), a.end(), greater<int>()); 
This would also fix the potential issues of ordering if using addComposing.

replace P::maxNumPasses with the largest value (i.e. after sorting; P::numPasses.at(0))

Add a verification that P::numPasses.size() is equal to P::amrMaxSpatialRefLevel+1

kstppd:
I think it is good to go now.

parameters.cpp
```diff
@@ -467,7 +471,43 @@ bool Parameters::getParameters(){
...
+   if (!isEmpty){
+
+      for (auto i : blurPassString){
+         P::numPasses.push_back(std::stoi(i));
+      }
+
+      
+      // Reverse Sort and Get the maximum number of filter passes
+      sort(numPasses.begin(), numPasses.end(), greater<int>());
+      P::maxNumPasses = *max_element(P::numPasses.begin(), P::numPasses.end());
```
kstppd, line 489: @markusbattarbee It belongs to Parameteres and it is also calculated as the max element from numPasses above.

markusbattarbee: right, well it should be either or. A parameter is an input thing, not something calculated in the code.

kstppd: Fixed!

PR #493, opened 2020-07-07
Meridional6D

kstppd:


ykempf:
Well if you give the three parameters to the user, they should all three be implemented or this will bite us one day – setting in the cfg but not seeing the effect.

ursg:
Agreed. Otherwise this looks sane.

kstppd:
@ykempf I am not sure I follow. By default they are set to 1 so they do not affect anything.

projects/Magnetosphere/Magnetosphere.cpp
```diff
@@ -390,8 +408,11 @@ namespace projects {
       
 #pragma omp parallel
       {
+         bool doZeroOut;
          //Force field to zero in the perpendicular direction for 2D (1D) simulations. Otherwise we have unphysical components.
-         if(P::xcells_ini==1) {
+         doZeroOut = P::xcells_ini ==1 && this->zeroOutComponents[0]==1;
+      
+         if(doZeroOut) {
```
ykempf, line 415: I would prefer this in one `if(P::xcells_ini ==1 && this->zeroOutComponents[0]==1)`

ykempf:
If you allow the user to set it to zero, they will one day. And they might wonder why setting to zero did not have the expected effect (or might not because it's hard to see "by eye").
So either do not provide the other two options as they would have no effect, or provide them including the corresponding implementation.

kstppd:
Ok so is it ok if I only keep a zeroOutDerivatives option instead of 3 ?

projects/Magnetosphere/Magnetosphere.cpp
kstppd: Done

ykempf:
It's more OK than having 3 out of which 2 do nothing... But now it is also confusing as the option name is not telling which plane is concerned...

ykempf:
Sorry for confusing you!

kstppd:
No problem!

PR #501, opened 2020-09-07
dz indexing in translation

ykempf:
The second-shortest bug possible...

Any reviewer will do, this is tested empirically...


markusbattarbee (approved):

ursg:
I would have actually suggested that the (now even more confusing) source code comment above that change be updated as well.

ykempf:
Well the comment is correct, we pass `dz + i` and deeper down `+ VLASOV_STENCIL_WIDTH` is included in the scheme...

PR #506, opened 2020-10-19
Dccrg update2 3 wide stencils

ykempf:
This is an uncontroversial bugfix. #teamshelob

The default stencil for DCCRG as well as the translation stencils need to be wider than 2 because of the way DCCRG handles neighbours at AMR interfaces, namely in terms of the local cell width, where Vlasiator needs 2 neighbours, i.e. 2 coarse ones even if the calling cell is refined.

Tests with a width of 4 didn't have significant differences.


ykempf:
Oops

PR #507, opened 2020-10-19
Polytropic gradpe

markusbattarbee:
Set electron solar wind driving conditions (temperature, density) and behaviour using polytropic index:

fieldsolver.electronTemperature: Upstream electron temperature to be used for the electron pressure gradient term (K)
fieldsolver.electronDensity: Upstream electron density to be used for the electron pressure gradient term (m^-3).
fieldsolver.electronPTindex: Polytropic index for electron pressure gradient term. 0 is isobaric, 1 is isothermal, 1.667 is adiabatic electrons


ykempf:
I now this is superseded, but I was wondering in the monster merge whether the density or pressure should be scaled before the polytropic index is applied? Or how are the units doing in lines like (pseudocode) `Pe = pow(P, polytropicIndex)`?

markusbattarbee:
Superceded by #512

Issue #510, opened 2020-11-12
Configure file for installation


henry2004y:
Hi everyone,

I have been playing with the code this week after the 101 tutorial. IMHO, the installation process has room for improvements.

Several suggestions:
1. All the source codes may be put into `src/` folder, but not under the same level as the main folder.
2. Making direct modifications in the main Makefile for setting parameters and track it with git may not be a good idea. I learned from the SWMF code at Michigan and Athena code at Princeton that we can use Makefile templates together with a configuration file to set the parameters before compilation.
3. After compilation, the obj and executable may be locate in separate folders under the main folder.

As a proof of concept, I forked the repo [here](https://github.com/henry2004y/vlasiator/tree/dev) and committed some changes to the dev branch. It demonstrate several features:
1. Source codes are now all under the `src/` directory.
2. A new [configure.py](https://github.com/henry2004y/vlasiator/blob/dev/configure.py) is used to set the parameters. It will use [Makefile.in](https://github.com/henry2004y/vlasiator/blob/dev/Makefile.in) as a template to generate the actual Makefile for compilation. For example, on my laptop I use the following command:
```
./configure.py -debug -mpi -omp -distfloat \
    --include=src \
    --include=src/projects \
    --include=/usr/include/boost \
    --include=$HOME/lib/zoltan/include \
    --include=$HOME/include/vlsv \
    --fsgrid_path=$HOME/include \
    --dccrg_path=$HOME/lib/dccrg \
    --include=$HOME/include/fsgrid \
    --include=$HOME/include/jemalloc/include \
    --include=$HOME/include/phiprof/include \
    --include=$HOME/include \
    --lib=boost_program_options \
    --lib=zoltan \
    --lib=vlsv \
    --lib=jemalloc \
    --lib=phiprof \
    --boost_path=/usr/lib/x86_64-linux-gnu \
    --zoltan_path=$HOME/lib/zoltan/lib \
    --vlsv_path=$HOME/include/vlsv \
    --jemalloc_path=$HOME/include/jemalloc/lib \
    --phiprof_path=$HOME/include/phiprof/lib
```
This will set all the required/optional library dependencies and generate `Makefile` in the main folder.
3. Then if you type `make`, it will first generate two folders `obj` and `bin`, create `version.cpp` with `generate_version.sh` and compile the code.

[...]

PR #511, opened 2020-11-13
Simplify the installation and organized source

henry2004y:
As I mentioned in the (Configure file for installation)[https://github.com/fmihpc/vlasiator/issues/510#issue-741505081], this process will make it easier to install the code on a new machine. The only prerequisites are `python`, `git`, `wget` and a MPI library.


ursg:
Wow, this is a massive effort, and a bunch of different things mashed together, that I'd like to comment on separately:

- Creating a separate src/ subdirectory indeed makes things a bit cleaner. Then we'd have src/, doc/, testpackage/, tools/ lying next to each other on an equal. Of course, the current situation is simply how it has organically grown, but this is no reason to leave it as such. So I like that!
- Likewise, a separate dir for the object files make sense. I would though suggest that the resulting vlasiator binary still comes out in the root directory
- I'm not a big fan of that configure script, in part because I don't see the problem it would be solving. The current mechanism, with the VLASIATOR_ARCH environment variable and architecture-dependent makefiles, makes it easy to build on different architectures. The reality of moving to a new machine is, that it's environment is typically very similar to an existing one, so in most cases we have been able to simply copy an existing file over and slightly modify. Since most of our target platforms are HPC machines in some supercomputing centres, at least some of the libraries will be supplied by the vendor there, and we'll have to first locate their directories and put them into the configure command line -- in the end, we'll end up writing a platform-specific script that ends up doing the same that our system-specific makefiles are doing now.
- I also don't quite get your point of direct modifications in the main makefile. The actual toplevel "Makefile" has been changing quite seldomly, with most of the changes and updates going to the individual system setups. If you specifically are talking about orders of the solvers, velocity space block sizes etc., then I agree we might consider moving *those* out into separate per-user config files. So far, we haven't been touching them very often either.
- What are the "gr_user" coordinates you are referencing there? :D
- I *do* like the effort you put into building a library auto-installation script, and I'm strongly in favour of keeping at least that part around. For installing on a smaller machine, that is absolutely helpful! On HPC machines, expect us to require lots of fiddling anyway in the near future still.

So, overall comment: awesome work! I'm looking forward to you convincing me that the configure script is superior to our current setup.

henry2004y:
I agree that we won't gain much by replacing the customized Makefiles in the MAKE directory with scripts. So this part can be improved to utilize the existing available Makefile format, and make changes accordingly. However, given that C/C++ handles dependencies already in the header-include way, we can simply replace all the explicit dependency list.

For me, setting parameters like order of solvers, turn on debugging/profiling etc. is ok, but I think it would be more natural to set them with the config file. Since we are controlling most of those through preprocessors, one would less likely make a mistake if he or she can set these with numbers and names instead of turning on and off the `-D` macros.

This config file originates from the [Athena++ code](https://github.com/PrincetonUniversity/athena-public-version), which is one of the most famous MHD code in astrophysics. I followed their template and removed many stuffs that we wouldn't need now.  But there are still leftovers, just as `gr_user` as an option for user-defined grid.

markusbattarbee:
My input:
 having a configure-style system as an option makes sense :) but I also think that keeping machine/compiler-specific Makefiles like we used to makes a lot of sense. On a supercomputer, often one person gets and compiles the required libraries and either stores them in his/her home directory or in a project directory, and then all of us can use those - and system-specific Makefiles make that a breeze.

Agree with object, source, and binary locations as per Urs' reply. A bit worried how easy current development branch comparisons with a dev version updated in this drastic manner would be, but at least my test merge into my electron branch appears to be mostly ok - although it deleted my own system-specific makefile?!?

Our old top-level Makefile is quite huge so if that can be shortened, I'm all up for that. But indeed it's very very rare that anyone needs to touch anything in there. Likewise, we really don't often need to edit solver orders - it's just either make, make testpackage, or going into the system-makefile to activate fallback vectorclass and that's it. (having fallback and testpackage_fallback as targets might be a nice change).

BTW, I would recommend that you read up on the forking workflow:
https://www.atlassian.com/git/tutorials/comparing-workflows/forking-workflow
we've tried to keep all changes in new separate branches, doing PRs into dev and then sometimes at milestones merging dev onto master. Doing changes in your own dev means that you may be unable to keep your own dev up to date with fmihpc/dev. :)

henry2004y:
I am working on both maintaining the availability of our previous approach and adding some new features for installation. Originally I thought the main Makefile in the top level should not be tracked by git since the config file will generate one for each setup. I may also happen to delete some Makefiles during the moving of files.

Not really familiar with git development especially in a large project, but hopefully I can learn from mistakes and don't accidentally create issues for working codes.

ursg:
My suggestion would be: the config script could be used to easily *create* such a machine specific makefile. Setting cflags, paths, libraries etc all happens in that situation anyway.

In the scenario where we are moving to a new machine, the first person to build the code there could then simply run the configure script with the --install option, take the resulting machine-specific makefile, commit it to the repository, and let the other know it's now ready to go.

henry2004y:
Now with some modifications, the process looks like the following:

1. For a fresh new machine,
```
./configure.py -h
```
shows the available options.

```
./configure.py -install
```
tries to install all the dependencies in the new `lib` folder.

To access an existing customized Makefile in `MAKE` folder with library paths and compiler flags, e.g.
```
./configure.py --machine=yann
```
then it will incorporate those into generating the main Makefile.

To save the configured paths into a new customized Makefile (by default Makefile.new)
```
./configure.py -install -save
```
to specify a name
```
./configure.py -install -save --save_machine=mymachine
```
or something like
```
python3.6 configure.py -debug -save -mpi -omp -distfloat \
    --include=src/projects \
    --include=/usr/include/boost \
    --include=lib/zoltan/include \
    --include=lib/vlsv \
    --include=lib \
    --vectorclass_path=lib/vectorclass \
    --fsgrid_path=lib/fsgrid \
    --dccrg_path=lib/dccrg \
    --include=lib/fsgrid \
    --include=lib/jemalloc/include \
    --include=lib/phiprof/include \
    --include=lib/vectorclass \
    --lib=boost_program_options \
    --lib=zoltan \
    --lib=vlsv \
    --lib=jemalloc \
    --lib=phiprof \
    --boost_path=/usr/lib/x86_64-linux-gnu \
    --zoltan_path=lib/zoltan/lib \
    --vlsv_path=lib/vlsv \
    --jemalloc_path=lib/jemalloc/lib \
    --profile_path=lib/phiprof/lib
```

If you really want to set everything manually, just remove the `-save` above and it will generate the main Makefile.

By default options with one `-` require no arguments, and options with two `--` require one arguments. I do not provide support for python2, since the official support has also ended by the end of last year.

ykempf:
So what is the status of this? Shall I test it on my laptop or Hawk and report or is that not needed?

It would be nice to get this in to make reviewing #521 easier.

henry2004y:
I resolved the conflicts manually today in one commits d666bac after a bunch of commits in my own branch but I am not sure how this git workflow will do here.

ykempf:
It seems not to like the shortcuts in machine Makefiles like `$(HOME)`, any way these can be used?

ykempf:
Order of semilog velocity:  5
  Order of semilog spatial:   3

That should be semilag or so. :)

ykempf:
This will take some learning to reconfigure before make clean and make...

henry2004y:
The conflicts are yet again resolved manually in f289ddf. If it makes reviewing easier, I can resubmit the pull requests from scratch, one purely for directory changes, one for the boundary conditions and related stuff. Let me know.

henry2004y:
I have decided to resubmit the pull requests. Let me know what is a proper time!

PR #508, opened 2020-10-28
Egi monstermerge

ursg:
This is the amalgamation of a big number of loose branches that we pull together for the next big 3D simulation ("EGI"):

- [x] https://github.com/markusbattarbee/vlasiator-public/tree/dccrg_update2
- [x] https://github.com/fmihpc/vlasiator/pull/507
- [x] https://github.com/fmihpc/vlasiator/pull/491 (plus a small dirty fix for crashes, that might still be overriden)
- [x] https://github.com/fmihpc/vlasiator/pull/485
- [x] https://github.com/fmihpc/vlasiator/pull/484
- [x] https://github.com/markusbattarbee/vlasiator-public/pull/9
- [x] https://github.com/ursg/vlasiator/tree/ionosphere
- [x] Vlsvdiff updates from https://github.com/kstppd/vlasiator/tree/vlsvTools

Still Missing:
- [ ] Bulirsch-Stoer fixes for ionosphere stepping

Current state of this is that it compiles, and will still need to be tested before it gets green light to go.

PR #512, opened 2020-11-19
Monster minus iono backport

markusbattarbee:
This PR replaces the EGI monstermerge #508 which had an ionosphere causing crashes.
Originally started as a dccrg update  https://github.com/markusbattarbee/vlasiator-public/tree/dccrg_update2 and various small AMR bugfixes, but ended up reverting to using the Nov 8th 2018 DCCRG version due to memory and startup bloat.

Includes PRs:
 #507
 #491
 #485
 #484
 markusbattarbee#9
 Vlsvdiff updates from https://github.com/kstppd/vlasiator/tree/vlsvTools

Used to run EGI on Hawk, November 2020 onwards
(should this be also then merged into master as Vlasiator 5.1?)


markusbattarbee:
> This should be moved before feedMomentsIntoFsGrid above.

why? These lines
   momentsGrid.updateGhostCells();
   momentsDt2Grid.updateGhostCells();
happen on init, and I would think we want to have the updated ghosts for FSgrid *after* fsgrid gets some values :)

ykempf:
This I think was referring to the technicalGrid update. That grid should be updated once and for all at the beginning, then we're good.

markusbattarbee:
OK, yes now I remember discussion about that. Apparently parsing comments to source code lines went awry here :) It's all good then.

ykempf:
I had to rethink too how it went, whether a comment went before or after the relevant line...

ursg:
I would reccomend retargeting this pull request to merge into fmihpc:dev instead of master though.

Otherwise, I agree, this should be merged.

ioread.cpp
```diff
@@ -900,12 +916,26 @@ template<unsigned long int N> bool readFsGridVariable(
...
+            if(file.readArray("VARIABLE",attribs, fileOffset, thatTasksSize[0]*thatTasksSize[1]*thatTasksSize[2], (char*)readBuffer.data()) == false) {
+               logFile << "(RESTART)  ERROR: Failed to read fsgrid variable " << variableName << endl << write;
+               return false;
+            }
+
+            for(int i=0; i< thatTasksSize[0]*thatTasksSize[1]*thatTasksSize[2]*N; i++) {
+               buffer[i]=readBuffer[i];
+            }
          }
+         phiprof::stop("readArray");
```
ursg, line 936: Just tested this:
This is an unmatched phiprof::stop, which causes this branch to crash on restart reading.

MAKE/Makefile.vorna_gcc
```diff
@@ -0,0 +1,94 @@
...
+INC_PROFILE = -I$(LIBRARY_PREFIX)/$(CC_BRAND)/$(CC_BRAND_VERSION)/phiprof/include
+
+LDFLAGS += -Wl,-rpath=$(LIBRARY_PREFIX)/$(CC_BRAND)/$(CC_BRAND_VERSION)/phiprof/lib:$(LIBRARY_PREFIX)/$(CC_BRAND)/$(CC_BRAND_VERSION)/boost/lib:$(LIBRARY_PREFIX)/$(CC_BRAND)/$(CC_BRAND_VERSION)/vlsv/lib:$(LIBRARY_PREFIX)/$(CC_BRAND)/$(CC_BRAND_VERSION)/jemalloc/lib:$(LIBRARY_PREFIX)/$(CC_BRAND)/$(CC_BRAND_VERSION)/zoltan/lib   
+
+LIB_PAPI = -L$(LIBRARY_PREFIX)/$(CC_BRAND)/$(CC_BRAND_VERSION)/papi/lib -lpapi -Wl,-rpath=$(LIBRARY_PREFIX)/$(CC_BRAND)/$(CC_BRAND_VERSION)/papi/lib
+INC_PAPI = -I$(LIBRARY_PREFIX)/$(CC_BRAND)/$(CC_BRAND_VERSION)/papi/include
+
+#header libraries
+INC_FSGRID = -I$(LIBRARY_PREFIX)/fsgrid/
+INC_DCCRG = -I$(LIBRARY_PREFIX)/dccrg-master/
```
ursg, line 89: I suppose this should also point to ```/dccrg/``` instead of ```/dccrg-master/```?

markusbattarbee: yes, correct. Fixed.

ioread.cpp
markusbattarbee: Interesting. I ran a full testpackage of this version including read_restart and it ran ok - but of course that is inside the FSgrid part which only runs if the number of ranks has changed. And also running this brach for EGI on Hawk has not encountered that issue as unlike in usual run situations, we've been running right from the start at high node counts.

markusbattarbee: so apparently that was in the profile fsgrid read branch but had been lost in the merging of float reading. Whoops. Fixed now.

markusbattarbee:
@ursg Thanks for the bug comments, fixed those two and re-targeted to dev.

ursg:
I confirm that this is running (and successfully restarting, both on the same number of tasks and a different one) now, both in small tests and "at scale".

parameters.cpp
```diff
@@ -296,6 +317,7 @@ bool Parameters::addParameters(){
    Readparameters::add("AMR.box_center_x","x coordinate of the center of the box that is refined (for testing)",0.0);
    Readparameters::add("AMR.box_center_y","y coordinate of the center of the box that is refined (for testing)",0.0);
    Readparameters::add("AMR.box_center_z","z coordinate of the center of the box that is refined (for testing)",0.0);
+   Readparameters::addComposing("AMR.filterpasses", std::string()+"AMR filter passes for each individual refinement level");
```
ykempf, line 320: Why the `std::string()+` here @kstppd?

mini-apps/build_pencils/grid_test_neighbors.cpp
```diff
@@ -132,8 +132,10 @@ setOfPencils buildPencilsWithNeighbors( dccrg::Dccrg<grid_data> &grid,
    // Maybe you could use physical coordinates here?
    if( startingRefLvl > path.size() ) {
       for ( uint i = path.size(); i < startingRefLvl; i++) {
-         auto parent = grid.get_parent(id);
-         auto children = grid.get_all_children(parent);
+         auto parent = grid.mapping.get_parent(id);
+
+	 std::array<uint64_t, 8> childrenarr = mpiGrid.mapping.get_all_children(parent);
```
ykempf, line 137: Dodgy indentation?

projects/projects_common.h
```diff
@@ -72,23 +72,24 @@ namespace projects {
...
+/*     auto neighbors = mpiGrid.get_neighbors_of_at_offset(cellID, i, j, k); */
+/*     // Warning: DCCRG currently uses indexing based on greatest possible refinement level, not the current cell size. */
+    
+/*     //FIXME: support refined grids */
+/*     if(neighbors.size() > 0) { */
+/*         return neighbors[0].first; */
+/*     } else { */
+/*         return INVALID_CELLID; */
+/*     } */
+/* } */
```
ykempf, line 92: How about just ditching these?

sysboundary/sysboundarycondition.cpp
```diff
@@ -613,46 +613,98 @@ namespace SBC {
...
+         //          if(cell != INVALID_CELLID) {
+         //             if(mpiGrid[cell]->sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY) {
+         //                cuint d2 = i*i+j*j+k*k;
+         //                if(d2 == dist) {
+         //                   closestCells.push_back(cell);
+         //                }
+         //             }
+         //          }
+         //       }
+
```
ykempf, line 707: Also here I am in favour of removing lengthy commented blocks instead.

testpackage/tests/restart_write/restart_write.cfg
```diff
@@ -7,6 +7,7 @@ dynamic_timestep = 1
 ParticlePopulations = proton
 
 [io]
+#write_as_float = 1
```
ykempf, line 10: Should this be tested separately too?

testpackage/tests/restart_write/restart_write.cfg
```diff
@@ -17,6 +18,10 @@ system_write_distribution_xline_stride = 0
 system_write_distribution_yline_stride = 0
 system_write_distribution_zline_stride = 0
 
+[restart]
+#write_as_float = 1
```
ykempf, line 22: Same as above?

testpackage/tests/transtest_amr/transtest_amr.cfg
```diff
@@ -49,10 +49,11 @@ vy_min = -2.0e6
 vy_max = +2.0e6
 vz_min = -2.0e6
 vz_max = +2.0e6
-vx_length = 1
-vy_length = 1
-vz_length = 1
-max_refinement_level = 1
+vx_length = 30
```
ykempf, line 52: Is this supposed to be now so much bigger?

tools/vlsvdiff.cpp
```diff
@@ -40,6 +40,7 @@
```
ykempf, line 40: So I am not now going through vlsvdiff as I hope/trust it works. If it is working, I suggest that field variables be included in the output variable and to-be-diffed lists of the testpackage, where relevant (i.e. not the trans and acc dedicated tests of course).

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -7,6 +7,8 @@
 #include "cpu_trans_map_amr.hpp"
 #include "cpu_trans_map.hpp"
 
+// use DCCRG version Nov 8th 2018
```
ykempf, line 10: Add the hash here too then?

ykempf: 01482cfba8

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -92,48 +93,95 @@ void computeSpatialSourceCellsForPencil(const dccrg::Dccrg<SpatialCell,dccrg::Ca
...
 
       int refLvl = mpiGrid.get_refinement_level(ids.front());
-      
       if (neighbors.size() == 1) {
+         if (sourceCells[iSrc+1] == mpiGrid[neighbors.at(0)]) continue; // already found this cell for different distance         
          sourceCells[iSrc--] = mpiGrid[neighbors.at(0)];
       } else if ( pencils.path[iPencil][refLvl] < neighbors.size() ) {
+         if (sourceCells[iSrc+1] == mpiGrid[neighbors.at(pencils.path[iPencil][refLvl])]) continue; // already found this cell for different distance (should not happen)
          sourceCells[iSrc--] = mpiGrid[neighbors.at(pencils.path[iPencil][refLvl])];
+         // Code to verify that multiple neighbors are in correct ordering (z-y-x)
```
ykempf, line 142: Add this block behind a `if(debug)` or so? We have a few of those in this file anyway.

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -604,21 +698,16 @@ void propagatePencil(
...
-         
-         for (uint planeVector = 0; planeVector < VEC_PER_PLANE; planeVector++) {            
-            int im1 = i - 1; // doing this to shut up compiler warnings
-            values[i_trans_ps_blockv_pencil(planeVector, k, im1, lengthOfPencil)] =
-               targetValues[i_trans_pt_blockv(planeVector, k, im1)];
-            
-         }
-      }
-   }  
+   // Previously we wrote target data into source data, but this is unnecessary
```
ykempf, line 701: Ditch this block?

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -856,15 +998,16 @@ void check_ghost_cells(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>
          dy = mpiGrid[ids[0]]->SpatialCell::parameters[CellParams::DZ];
          break;
       case 1:
-         dx = mpiGrid[ids[0]]->SpatialCell::parameters[CellParams::DZ];
-         dy = mpiGrid[ids[0]]->SpatialCell::parameters[CellParams::DX];
+         dx = mpiGrid[ids[0]]->SpatialCell::parameters[CellParams::DX];
+         dy = mpiGrid[ids[0]]->SpatialCell::parameters[CellParams::DZ];
```
ykempf, line 1002: I prefer the previous ordering. But this is not used anyway. ;)

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1245,8 +1402,9 @@ bool trans_map_1d_amr(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
 
                         // Unpack the vector data
                         Realf vector[VECL];
-                        pencilSourceVecData[pencili][i_trans_ps_blockv_pencil(planeVector, k, icell - 1, L)].store(vector);
-                        
+                        //pencilSourceVecData[pencili][i_trans_ps_blockv_pencil(planeVector, k, icell - 1, L)].store(vector);
+			pencilTargetValues[pencili][i_trans_pt_blockv(planeVector, k, icell - 1)].store(vector);
```
ykempf, line 1406: Indentation?

ykempf (changes requested):
Here's my review.

ykempf:
Look at me, I wrote a review and not a stack of individual comments. :)

vlasovsolver/cpu_trans_map_amr.cpp
markusbattarbee: It's true that the dx andy values are not currently used for anything, but I want to point out that this is not a cosmetic change but actually fixes the orderings to match with what dccrg assumes.

ykempf: All right then, that reason primes over my esthetic preferences.

parameters.cpp
ursg: I would suggest this should better be ```std::string("text")```.

I understand what's happening here: if you *just* put a ```const char *``` there, in the form of a C string literal, that gets typeconverted into an integer type at compile time, so you end up with a integer-type parameter. Making sure that the datatype is ```std::string``` avoids that. But constructing a ```std::string``` by using the ```+``` operator is a bit odd. :)

ykempf: Oh, I see that that's been done for other calls to `addComposing`, I never noticed that (or when it changed).

ykempf: But not for the IO parameters at the top of this function. So I don't see why this is needed here.

ursg:
> Here's my review.

Hmm, while I in principle agree with all of the points in your review, I'd actually suggest we merge this pull request first, as I functionally approve it. The code runs in production and clearly improves the state of the dev branch there. The fixes you suggest are mostly about code aesthetics

In other words: "If this is all you can complain about, that is actually a very positive sign of the code quality" :)

So I suggest we merge and fix these then.

ykempf:
How's keeping track of those going to work though? I wasn't planning on copying the comments to an issue each now...

fieldsolver/derivatives.cpp
```diff
@@ -58,7 +58,12 @@ void calculateDerivatives(
    // Get boundary flag for the cell:
    cuint sysBoundaryFlag  = technicalGrid.get(i,j,k)->sysBoundaryFlag;
    cuint sysBoundaryLayer = technicalGrid.get(i,j,k)->sysBoundaryLayer;
-   
+
+   // Constants for electron pressure derivatives
+   // Upstream pressure
+   Real Peupstream = Parameters::electronTemperature * Parameters::electronDensity * physicalconstants::K_B;
+   Real Peconst = Peupstream * pow(Parameters::electronDensity, -Parameters::electronPTindex);
```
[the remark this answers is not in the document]
markusbattarbee: I calculated that they work out exactly as they should, but please do verify my math. :)

fieldsolver/ldz_gradpe.cpp
```diff
@@ -47,8 +47,9 @@ void calculateEdgeGradPeTermXComponents(
       case 1:
          rhoq = momentsGrid.get(i,j,k)->at(fsgrids::moments::RHOQ);
          hallRhoq = (rhoq <= Parameters::hallMinimumRhoq ) ? Parameters::hallMinimumRhoq : rhoq ;
-         EGradPeGrid.get(i,j,k)->at(fsgrids::egradpe::EXGRADPE) = -physicalconstants::K_B*Parameters::electronTemperature*dMomentsGrid.get(i,j,k)->at(fsgrids::dmoments::drhoqdx) / (hallRhoq*EGradPeGrid.DX);
-         break;
+         //EGradPeGrid.get(i,j,k)->at(fsgrids::egradpe::EXGRADPE) = -physicalconstants::K_B*Parameters::electronTemperature*dMomentsGrid.get(i,j,k)->at(fsgrids::dmoments::drhoqdx) / (hallRhoq*EGradPeGrid.DX);
+         EGradPeGrid.get(i,j,k)->at(fsgrids::egradpe::EXGRADPE) = - dMomentsGrid.get(i,j,k)->at(fsgrids::dmoments::dPedx) / (hallRhoq*EGradPeGrid.DX);
```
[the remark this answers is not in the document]
markusbattarbee: Hmm, that is an interesting point which I hadn't considered. Indeed we might want to use a non-limited RHOQ here as it's the non-limited version which goes into the limiter.

testpackage/tests/transtest_amr/transtest_amr.cfg
markusbattarbee: Reverted this - although to note that transtest_amr is designed for testing velocity AMR, not spatial AMR.

markusbattarbee (commented):
Ok, I implemented the cosmectic changes.
Leaving the EgradPe limiting as is to err on the side of caution - it should only affect low-density regions anyway. Did not implement the neighbor order checking as a debug because it was in fact an alternative approach, not a check per se - but did leave the code in in case someone wants to use it in the future.

fieldsolver/ldz_gradpe.cpp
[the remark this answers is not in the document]
markusbattarbee: On second thought, I think we want to keep it this way. We are calculating the gradient of the electron pressure tensor (non-limited), but then in the Ohm's law term denominator, limit the electron density.

mini-apps/build_pencils/grid_test_neighbors.cpp
```diff
@@ -132,8 +132,10 @@ setOfPencils buildPencilsWithNeighbors( dccrg::Dccrg<grid_data> &grid,
    // Maybe you could use physical coordinates here?
    if( startingRefLvl > path.size() ) {
       for ( uint i = path.size(); i < startingRefLvl; i++) {
-         auto parent = grid.get_parent(id);
-         auto children = grid.get_all_children(parent);
+         auto parent = grid.mapping.get_parent(id);
+
+         std::array<uint64_t, 8> childrenarr = mpiGrid.mapping.get_all_children(parent);
+	 vector<CellID> children(childrenarr.begin(), childrenarr.end());
```
ykempf, line 138: This still has indentation issues, looks like a tab?

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -148,116 +194,129 @@ void computeSpatialSourceCellsForPencil(const dccrg::Dccrg<SpatialCell,dccrg::Ca
...
-            backNeighborIds.push_back(nbrPair.first);
-         }
-      }
-
-      if (backNeighborIds.size() == 0) {
-         abort();
+      const auto frontNeighbors = mpiGrid.get_face_neighbors_of(ids.front());
+      if (frontNeighbors.size() > 0) {
+	for (const auto nbr: frontNeighbors) {
+	  if(nbr.second == (-((int)dimension + 1))) {
```
ykempf, line 274: 2-space indentation?

ykempf (approved):
I scrolled through, looks ok, up to the 2 peanuts, no need to worry about those...

PR #513, opened 2020-11-23
Added ghost cell updates for volumetric field calculation

markusbattarbee:
At the end of ldz_main.cpp, after fields have been propagated, volumetric fields and bvol derivatives are updated, but the volumetric field calculation uses non-updated ghost information. This fixes it. Tested to compile but didn't check further yet.


markusbattarbee:
Unnecessary after all as the updates take place in previous sub-functions of the fieldsolver.

PR #514, opened 2020-11-26
Initial grid stripe

### LINKED ISSUE #509: initial-grid writing should be striped too
initial-grid is very slow writing in a 3D production run. I suspect it's not striped by default like restarts. We might want to include that to the initial-grid IO as it is also a big file.

markusbattarbee:
This is to fix #509 
Quite possible though that bulk files already are written with the system default striping setting, but now it can be manually set.
(not tested yet)


markusbattarbee:
testing this now on Vorna

markusbattarbee:
Well, on Vorna, setting either stripe setting (bulk or restart) did not affect the actual striping of the file. I tried setting them to 8 and 12 respectively, with the directory stripe value set to 16 (maximum) or -1, and either way around, the files report 16 with `lfs getstripe -c`. Hmm. Lustre-dependent settings preventing an override?

ursg:
My suspicion is that the MPI implementation on Vorna is simply not aware of lustre, and thus ignores this MPI hint. In these cases, it probably makes sense to set the striping factor on the working directory with the lfs tool before submitting, or adjust it in the job script.

markusbattarbee:
Anyhow can confirm testpackage zero diffs.

markusbattarbee:
Tested this on Mahti, there also the settings didn't propagate into an actual file striping. Will need to test on Puhti and Hawk as well.

ykempf:
Can you at the same time check whether restart striping hints propagate to the files or not?

markusbattarbee:
Yeah, testing both initial, restart and bulk, all of those had the directory striping value set to them. Also checked  with directory striping values of -1 (default) and a manually set value (10).

markusbattarbee:
Well, on Hawk this PR works as intended.

ursg:
I'm wondering: can we somehow query from the MPI implementation, whether it supports and implements this? So that we could give a friendly warning message in those situations?

ursg:
[This slide set](http://www.cenapad-rj.lncc.br/tutoriais/materiais-hpc/semana-sdumont/mpi-io-lncc-2018-jean-bez.pdf) is listing some code (on page 101-109), to query all predifined mpi info values for a newly opened file. We could do that, and check if a striping parameter even exists there, but I'm not sure this is sufficient to test for it.

[This Nasa document](https://www.nas.nasa.gov/hecc/support/kb/lustre-best-practices_226.html) on the other hand warns that "The hints are only advisory and may not be honored. For example, SGI MPT 2.12r26 honors these hints, but MPT 2.14r19 does not. Intel MPI 5.0x honors these hints when the I_MPI_EXTRA_FILESYSTEM and I_MPI_EXTRA_FILESYSTEM_LIST variables are set appropriately", so this might be completely arbitrary.

markusbattarbee:
I think verifying if striping is set to the file is a further code improvement which can be potentially tackled later, separate from this PR.

ursg:
Fair enough. Merge early, merge often. :+1:

PR #515, opened 2020-11-26
Only write bulk file after restart if not first timestep

### LINKED ISSUE #384: bulk writing at a restart uninitialised
When restarting, if a bulk should be written right  away it overwrites the one written just before when stopping. This is an expected and probably desired behaviour. However some fields, in particular populations/moments, are not recomputed yet when that IO is performed. This is not very nice in regard to post-processing/analysis and I suggest to fix that, even though this comes at the cost of an ugly workaround.

markusbattarbee:
Fixes #384

### DIFF
--- a/MAKE/Makefile.Freezer
+++ b/MAKE/Makefile.Freezer
@@ -1,0 +1,112 @@
+CMP = mpic++
+LNK = mpic++
+
+# Markus' desktop computer
+
+#======== Vectorization ==========
+#Set vector backend type for vlasov solvers, sets precision and length. 
+#Options: 
+# AVX:	    VEC4D_AGNER, VEC4F_AGNER, VEC8F_AGNER
+# AVX512:   VEC8D_AGNER, VEC16F_AGNER
+# Fallback: VEC4D_FALLBACK, VEC4F_FALLBACK, VEC8F_FALLBACK
+
+ifeq ($(DISTRIBUTION_FP_PRECISION),SPF)
+#Single-precision        
+	VECTORCLASS = VEC8F_AGNER
+	INC_VECTORCLASS = -I$(LIBRARY_PREFIX)/vectorclass
+# 	VECTORCLASS = VEC8F_FALLBACK
+#	INC_VECTORCLASS = -I$(LIBRARY_PREFIX)/../vlasiator/vlasovsolver
+else
+#Double-precision
+#	VECTORCLASS = VEC4D_AGNER
+#	INC_VECTORCLASS = -I$(LIBRARY_PREFIX)/vectorclass
+	VECTORCLASS = VEC4D_FALLBACK
+	INC_VECTORCLASS = -I$(LIBRARY_PREFIX)/../vlasiator/vlasovsolver
+endif
+
+#======== PAPI ==========
+#Add PAPI_MEM define to use papi to report memory consumption?
+#CXXFLAGS +=  -DPAPI_MEM
+
+
+#======== Allocator =========
+#Use jemalloc instead of system malloc to reduce memory fragmentation? https://github.com/jemalloc/jemalloc
+#Configure jemalloc with  --with-jemalloc-prefix=je_ when installing it
+#CXXFLAGS += -DUSE_JEMALLOC -DJEMALLOC_NO_DEMANGLE
+
+
+#======= Compiler and compilation flags =========
+# NOTES on compiler flags:
+# CXXFLAGS is for compiler flags, they are always used
+# MATHFLAGS are for special math etc. flags, these are only applied on solver functions
+# LDFLAGS flags for linker
+
+#-DNO_WRITE_AT_ALL:  Define to disable write at all to 
+#                    avoid memleak (much slower IO)
+#-DMPICH_IGNORE_CXX_SEEK: Ignores some multiple definition 
+#                         errors that come up when using 
+#                         mpi.h in c++ on Cray
+
+#CXXFLAGS += -DMPICH_IGNORE_CXX_SEEK
+
+#FLAGS = -ggdb
+
+#GNU flags:
+CC_BRAND = gcc
+CC_BRAND_VERSION = 6.2.0
+CXXFLAGS += -O3 -fopenmp -funroll-loops -std=c++17 -W -Wall -Wno-unused -fabi-version=0 -mavx2
+#CXXFLAGS += -ggdb -O0 -fopenmp -funroll-loops -std=c++17 -W -Wall -Wno-unused -fabi-version=0 -mavx2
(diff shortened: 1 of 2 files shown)


ursg:
In the complete set of bulk files of a simulation in temporal order, this would cause the bulk written after restart to be a bit more out-of-order than the other ones, right? They are anyway not perfectly equidistant in time, as the bulks are written whenever their writing interval gets exceeded, but this one will be off by a bit more.

As most of our analyses aren't sensitive to bulk data being exactly time-equidistant, this should be quite ok. For temporal variability studies (such as wave Fourier transforms), we have to resample anyway already, so this fix doesn't change a lot there, anymore.

markusbattarbee:
The idea is to not write the bulk file after a restart as that one was already written by the previous job. 

However, you are correct that P::systemwrites[i] was incorrectly left un-incremented. So I made a fix to that. Will need to verify that this works as well...

ursg:
Ah, yep, that looks like a reasonable fix. Before merging this, let's make sure it is properly tested.

markusbattarbee:
Ok, I was able to verify that this works. I did however at the same time find out that the current method of just incrementing P::systemWrites[i] for each bulk file write does not work correctly for the (perhaps pathological) case of a time step which is greater than an output step (and we should end up writing a non-continuous list of bulk files)

markusbattarbee:
Directory listing
```
-rw-rw---- 1 markusb hyad-all    11769 Nov 30 16:52 logfile.txt
-rw-rw---- 1 markusb hyad-all     9112 Nov 30 16:52 phiprof_0.txt
-rw-rw---- 1 markusb hyad-all 70882445 Nov 30 16:52 restart.0000302.2020-11-30_16-52-30.vlsv
-rw-rw---- 1 markusb hyad-all 70776876 Nov 30 16:52 bulk.0000015.vlsv
-rw-rw---- 1 markusb hyad-all 68407404 Nov 30 16:52 bulk.0000014.vlsv
-rw-rw---- 1 markusb hyad-all 66628236 Nov 30 16:52 bulk.0000013.vlsv
-rw-rw---- 1 markusb hyad-all 64593132 Nov 30 16:52 bulk.0000012.vlsv
-rw-rw---- 1 markusb hyad-all 62578973 Nov 30 16:52 restart.0000221.2020-11-30_16-52-29.vlsv
-rw-rw---- 1 markusb hyad-all 62473404 Nov 30 16:52 bulk.0000011.vlsv
-rw-rw---- 1 markusb hyad-all 60093540 Nov 30 16:52 bulk.0000010.vlsv
-rw-rw---- 1 markusb hyad-all 57575460 Nov 30 16:52 bulk.0000009.vlsv
-rw-rw---- 1 markusb hyad-all 54442308 Nov 30 16:52 bulk.0000008.vlsv
-rw-rw---- 1 markusb hyad-all 51765300 Nov 30 16:52 bulk.0000007.vlsv
-rw-rw---- 1 markusb hyad-all       86 Nov 30 16:52 loki_read.txt
drwxrwx--- 2 markusb hyad-all     4096 Nov 30 16:48 write
-rw-rw---- 1 markusb hyad-all 70882445 Nov 30 16:48 restart.0000302.2020-11-30_16-48-32.vlsv
-rw-rw---- 1 markusb hyad-all 64698701 Nov 30 16:48 restart.0000241.2020-11-30_16-48-32.vlsv
-rw-rw---- 1 markusb hyad-all 48485979 Nov 30 16:48 restart.0000120.2020-11-30_16-48-31.vlsv
-rw-rw---- 1 markusb hyad-all 48380506 Nov 30 16:48 bulk.0000006.vlsv
-rw-rw---- 1 markusb hyad-all 43951162 Nov 30 16:48 bulk.0000005.vlsv
-rw-rw---- 1 markusb hyad-all 39106954 Nov 30 16:48 bulk.0000004.vlsv
-rw-rw---- 1 markusb hyad-all 33895354 Nov 30 16:48 bulk.0000003.vlsv
-rw-rw---- 1 markusb hyad-all 28316362 Nov 30 16:48 bulk.0000002.vlsv
-rw-rw---- 1 markusb hyad-all 22369978 Nov 30 16:48 bulk.0000001.vlsv
```



First run log:
```
Mon Nov 30 16:48:31 2020 
(IO): Writing spatial cell and reduced system data to disk, tstep = 5 t = 100.736

Mon Nov 30 16:48:31 2020 
(writeGrid) Wrote 43.9442 MB in 0.00379069 seconds, approximate data rate is 11.5927 GB/s
(IO): .... done!

Mon Nov 30 16:48:31 2020 
---------- tstep = 6 t = 120.883 dt = 20.1472 FS cycles = 1 ----------

Mon Nov 30 16:48:31 2020 
(IO): Writing spatial cell and reduced system data to disk, tstep = 6 t = 120.883

Mon Nov 30 16:48:31 2020 
(writeGrid) Wrote 48.3735 MB in 0.00390334 seconds, approximate data rate is 12.3929 GB/s
(IO): .... done!

Mon Nov 30 16:48:31 2020 
(IO): Writing restart data to disk, tstep = 6 t = 120.883

Mon Nov 30 16:48:31 2020 
(writeGrid) Wrote 48.486 MB in 0.00337558 seconds, approximate data rate is 14.3637 GB/s
(IO): .... done!
```



Restart log:
```
Velocity meshes and data read, approximate data rate is 5.68786 GB/s
Mon Nov 30 16:52:29 2020 
(MAIN): Starting main simulation loop.

Mon Nov 30 16:52:29 2020 
---------- tstep = 6 t = 120.883 dt = 20.1472 FS cycles = 1 ----------

Mon Nov 30 16:52:29 2020 
---------- tstep = 7 t = 141.03 dt = 20.1472 FS cycles = 1 ----------

Mon Nov 30 16:52:29 2020 
(IO): Writing spatial cell and reduced system data to disk, tstep = 7 t = 141.03
```

markusbattarbee:
I guess we could change the bulk file output to actually use 
`      P::systemWrites[i] =(int)(P::t/P::systemWriteTimeInterval[i]); 
`?

ursg:
This has yet another corner case, in the situation where we change the output frequency during runtime (I'm specifically thinking of our switch between 5 and 1 second intervals).
Of course, that can simply be avoided by changing the bulk name at the same time, like we did.

markusbattarbee:
> This has yet another corner case, in the situation where we change the output frequency during runtime (I'm specifically thinking of our switch between 5 and 1 second intervals).
> Of course, that can simply be avoided by changing the bulk name at the same time, like we did.

Yep, and I don't think there's any elegant solution to that. I'll implement and test the suggested change.

markusbattarbee:
Ok, tested this, still works as intended for a regular case and if the timestep is larger than the output interval, it skips the indices where no data was available.

ursg (approved):
Excellent, let's merge it!

PR #516, opened 2020-12-08
Recent fix to bulk files after restart prevented output of bulk.0000000.vlsv

markusbattarbee:
...this fixes that :)


ykempf:
The email scared me! It has like 200 commits!

ykempf:
Looks good to me.

ursg:
Yup, seems reasonable.

PR #521, opened 2021-01-18
Time-dependent boundary condition

henry2004y:
Check if it can show conflicts.


src/boundary/boundarycondition.cpp
```diff
@@ -197,7 +197,7 @@ void BoundaryCondition::setCellBVOLDerivativesToZero(FsGrid<std::array<Real, fsg
  * closest boundarytype::NOT_BOUNDARY cell.
  * \param copyMomentsOnly If true, do not touch velocity space.
  */
-void BoundaryCondition::vlasovBoundaryCopyFromTheClosestNbr(
+void BoundaryCondition::vlasovBoundaryCopyFromClosestNbr(
```
ykempf, line 200: This was because we have two functions, one for "the closest" i.e. one cell, one for all "closest neighbors", and to make it more obvious "The" was included in the name.

src/boundary/maxwellian.cpp
```diff
@@ -243,10 +243,9 @@ void Maxwellian::generateTemplateCell(spatial_cell::SpatialCell &templateCell, R
                      for (uint vi = 0; vi < speciesParams[popID].nVelocitySamples; ++vi)
                         for (uint vj = 0; vj < speciesParams[popID].nVelocitySamples; ++vj)
                            for (uint vk = 0; vk < speciesParams[popID].nVelocitySamples; ++vk)
-                           {
                               avr += maxwellianDistribution(popID, rho, T, vxCell + vi * d_vx - Vx,
                                                             vyCell + vj * d_vy - Vy, vzCell + vk * d_vz - Vz);
-                           }
```
ykempf: Actually in principle for loops without a curly bracket are frowned upon in the Vlasiator style, only a few of such nested loops were accepted with a single pair of curly brackets. If something changes here, I'd rather have the three pairs than none at all.

src/vlasovsolver/cpu_moments.cpp
```diff
@@ -156,9 +155,7 @@ void calculateMoments_R(
        for (size_t c=0; c<cells.size(); ++c) {
           SpatialCell* cell = mpiGrid[cells[c]];
           
-          if (cell->boundaryFlag == boundarytype::NO_COMPUTE) {
-             continue;
-          }
+          if (cell->boundaryFlag == boundarytype::NO_COMPUTE) continue;
```
ykempf: Same here, in principle we don't want if's without a curly bracket pair.

src/vlasovsolver/vlasovmover.cpp
```diff
@@ -22,77 +22,71 @@
 
 #include <cstdlib>
 #include <iostream>
-#include <vector>
 #include <stdint.h>
+#include <vector>
```
ykempf: The whole file seems to have only white-space and indentation diffs, I probably missed whatever was significant. Also, quite a few related to having the curly brackets after instead of before the line break are against the style of the rest of the code.

ykempf (commented):
I gave up, there's so much white-space and file renaming diffs due to the other PR that it'll take me a day to look in detail.

src/vlasovsolver/vlasovmover.cpp
ykempf: I leave it up to @ursg about the changes in e.g. indentation and curly bracket style, but if it's not done in the whole code at once and with everyone knowing about it, the code will become a lot more difficult to read if there's different styles in different places in the code. Also, such massive white-space changes are rather done in a separate pass to make sure there's no hidden bug in the middle of thousands of lines of stuff.

If/when we merge the configure/repo sorting PR this will become a lot easier to read.

My four comments above only pertain to the last commit as I was looking only at those diffs by mistake.

As mentioned orally, the reason why I dragged throughout "sysboundary" instead of just "boundary" when the system boundaries were coded at the time is because there was places in the code where a clear distinction between system and process boundaries is (was?) needed. If this is no longer seen as a potential pitfall I don't mind if it gets changed to just "boundary".

src/boundary/maxwellian.cpp
henry2004y: I will change this style accordingly.

src/vlasovsolver/vlasovmover.cpp
henry2004y: I was worried about some of these style changes. I used some auto-styling tools to help formatting the source code, so it will potentially generate many differences like this. Sometimes I just couldn't control myself clicking the tool buttons...

But I also think instead of manually going over each line that changes, we need to have some auto CI checks and tests that can help us quickly identifying if anything changes for the existing verification tests. Urs mentioned to me there is one written in shell script that can run tests and compare with another git version. It might be helpful to introduce some easy-to-use standard tests with known results already in the repo and run it with, e.g., `make test flowthrough`.

henry2004y:
Currently this cannot change the magnetic field values in boundary cells. This part is still WIP. Once it can run, I'll think about how to speed things up, for example, by not generating a template cell at every required time step, but just storing the values.

src/boundary/boundarycondition.cpp
henry2004y: I noticed that, but since the other one is named `vlasovBoundaryCopyFromAllClosestNbrs` with a plural form, I would hope this is ok. There are other place where I see, e.g. `SysBoundaryConditions` vs `SysBoundaryCondition`, `SysBoundaries` vs `SysBoundary` etc.

ykempf: I think this one is OK. I picked up on it because I was seeing a single commit's worth of changes.

Places with e.g. sysBoundaries (a vector containing all of them) vs. the class name SysBoundary should be correct and consistent, but do point out if there's unclear things for sure.

src/boundary/maxwellian.cpp
ykempf: Thanks. :)

src/vlasovsolver/vlasovmover.cpp
ykempf: I think we need to automatise that indeed quite soon.

ykempf:
We started discussing the magnetic field too today, don't hesitate to ask. I was thinking that we could use the hook function in the project that can be called prior to or after each time step to update e.g. the interpolated magnetic field values, and then in the (system) boundary magnetic filed update function use that stored value.

Also, I think you can now already implement a user parameter setting at what intervals the values should be updated, as I don't think it is sensible to update every step. Any time dt changes (which is more a numerical than a physical thing) it would change the rate of update at the boundary too, which I think is dodgy. Setting an interval to a tiny value would force the update at every step e.g. for testing, and in production that could go to 0.5 o5 1 s in my opinion.

src/vlasovsolver/vlasovmover.cpp
ursg: I guess it's time we produce some preset files for the common editors (Vim? Emacs? VS Code?), so that indentation comes out right for everybody. Also, I guess an automatic a preset for an automatic code formatter like ClangFormat might be good.
I made a [separate Issue](https://github.com/fmihpc/vlasiator/issues/530) out of this.

If your commits already make these files more conforming to the [Vlasiator Coding Style](https://github.com/fmihpc/vlasiator/wiki/Coding-Style), then that's great. I wonder if we should pull that in as a separate pull request first though, just so that things are a little less confusing.

henry2004y:
This is now ready for review. The capability of setting a time interval for updating the dynamic boundary condition is also contained.

For a local 2D test run, the new top level function call `updateState` which updates the values in the boundary cells spend 13% of the total time in Simulation if called every time step, and less than 1% if only every 20s for a 1MPI 100s run with 20*20 spatial cells and 15^3 velocity cells. The performance may be improved further by not setting template cells as Yann suggested, but currently it is still doing so.

Initial tests are performed on 1D/2D nonAMR grid. My next goal is to come up with some concise tests that can be used as verification and also benchmarks.

src/vlasovsolver/vlasovmover.cpp
henry2004y: My latest commits followed the coding style, except some places where there might be missing curly braces around a single if statement.

henry2004y:
Because there are several things mixed together here, I've decided to resubmit the PRs. Let me know what is a good time!

PR #525, opened 2021-01-27
Fix slope limiter scaling to work if threshold is set to non-positive…

markusbattarbee:
… values. Removed trailing whitespace.

This will allow tests where a full non-sparse vspace is requested.


ursg:
Looks like this will keep the "normal" usecase completely unchanged and lead to zero diffs, right? In that case, :+1:

markusbattarbee:
Verified, testpackage gives zero diffs. (also included a fix to tp definitions in this PR)

ursg:
Great! Let's merge it.

PR #526, opened 2021-01-27
Flowthrough shapes

markusbattarbee:
Added possibility of initializing Flowthrough with an x-directional signal inside the box: square, sinewave or triangle.


ursg:
Looks good!

PR #527, opened 2021-01-27
Switch to fs stencil width

markusbattarbee:
Changed FSgrid creation to use a stencil width parameter instead of a hard-coded value


ursg:
This makes sense. Hardcoded numbers in datatypes are always bad style.

I would guess this can be merged right away if it doesn't change the default behaviour in any way.
For a sanity check, have you ever set this value to 3 and looked what happens? Does the code even still compile in that case?

markusbattarbee:
Changing it to 3 showed I had missed updating the one new datareducer (fg_amr_level). 
Compiles, runs, and testpackage gives zero diffs, both with stencils 2 and 3!

ursg:
Excellent, good that we tested. :)

PR #528, opened 2021-02-01
Merge dev into master, for Vlasiator 5.1 release

ursg:
This pull requests collects a big number of changes over the last year and bumps the Vlasiator version number to 5.1

Major changes include:
 - A long debugging campagin of the AMR solvers, which found many small bugs, so the AMR simulation results are more reliable
 - Fixed makefiles for a couple of new machines we have gotten access to
 - Scalability and memory usage improvements at scale.

This code has run some major large-scale runs on hawk and a bunch of small ones in the local clusters of the university of helsinki, so confidence is high.

We should still run some basic sanity tests on this, but otherwise I suppose it's good to be merged!


markusbattarbee:
As this may be our final milestone using the current build system, we might want to go through the included Makefiles and ensure they are up to date? That could include e.g. the changes to agner / fallback vectorclass selection, the C++ standard, and purging some old Makefiles for systems no longer in use.

What sanity tests would you suggest @ursg ?

ursg:
Good idea to clean up that makefile dir. I started #533, by simply removing a bunch of horribly outdated config files, that nobody had bothered to clean up yet.

For all the machines in our "current working set" (that is hawk, vorna, puhti, sisu and "home"), I suppose these are up to date and working?

ursg:
As for further sanity tests, I suppose our production runs on hawk and vorna pretty much qualify this code version as "well tested".

markusbattarbee:
Sanity tests are indeed done. 

As for the Makefiles, I think the vectorclass vs fallback include dir changes have not propagated to all makefiles - that should be fixed. And we probably want to ensure that all the files request a suitably up-to-date C++ standard (17?)

PR #534, opened 2021-02-11
Makefile cleanup part 2: Remove one more makefile, add VECTORCLASS path to two others.

ursg:
This should bring the MAKE directory into a reasonable state for merging into master.


markusbattarbee:
Hmm, we recently did a change where INC_VECTORCLASS was defined differently for AGNER and FALLBACK, due to some kind of contamination of loaded headers, I think. Yet looking at the source code, vlasovsolver/vec.h is the only place where vectorclass is included, and both types are inside vector IFDEF-checks. Thus, I don't see how enunciation like
```
ifeq ($(DISTRIBUTION_FP_PRECISION),SPF)
#Single-precision
        VECTORCLASS = VEC8F_AGNER
        INC_VECTORCLASS = -I$(LIBRARY_PREFIX)/vectorclass-version1
#       VECTORCLASS = VEC8F_FALLBACK
#       INC_VECTORCLASS = -I$(HOME)/vlasiator/vlasovsolver
else
#Double-precision
        VECTORCLASS = VEC4D_AGNER
        INC_VECTORCLASS = -I$(LIBRARY_PREFIX)/vectorclass-version1
#       VECTORCLASS = VEC4D_FALLBACK
#       INC_VECTORCLASS = -I$(HOME)/vlasiator/vlasovsolver
endif

```
would actually be even needed?

ursg:
The changing VECTORCLASS - define at least comes from architectural constraints: what is the maximum *width* of a vector of the given floating point type that fits into a register? As this varies from subarchitecture to subarchitecture, that construct makes sense.

But the different paths don't really. Or is there *any* platform, where we are using vectorclass in single precision and fallback for the testpackage? (Wouldn't make a lot of sense, would it?)

Anyway, since this seems to *work* everywhere at the moment, shouldn't we just leave it as it is?

markusbattarbee:
I didn't mean to comment on the vector lengths, just the include directories.

I think it might have been that with the PGI compiler we had to exclude the agner-vectorclass directory from being included, otherwise it tried to still pull that path in and caused compilation errors. And then we decided to make separate INC_VECTORCLASS values for AGNER and FALLBACK. But indeed that's probably a special case and doesn't need to be propagated throughout.

I'll pull this branch and do some more cleanup (like going to stdc++17 everywhere).

ykempf:
Sorry I missed the comment from 19 days ago. I had to do that or it wouldn't compile correctly with the version I wanted. And yes, I had to test fallback with testpackage to somewhat ascertain it was fixed properly.

markusbattarbee (approved):
Fixed the merge conflict - Merge away!

PR #535, opened 2021-02-11
Fsgrid restart debug, when restarting with different number of tasks.

ursg:
We are seeing big simulations crashing, when we try to restart them with a different number of mpi tasks.
The offending codepath is exactly in fsgrid reading, where data has to be reordered when restarting from a different task number.

This crash is not reproduced in small runs.

First attempt: could this be a simple signed 32bit vs. unsigned 64bit mismatch issue?

### DIFF
--- a/ioread.cpp
+++ b/ioread.cpp
@@ -917,28 +917,30 @@ template<unsigned long int N> bool readFsGridVariable(
          std::vector<Real> buffer(thatTasksSize[0]*thatTasksSize[1]*thatTasksSize[2]*N);
 
          phiprof::start("readArray");
-         if(!convertFloatType) {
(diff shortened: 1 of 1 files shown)


ursg:
I tested this with EGL production restart files on 128 nodes of hawk: The restart crashes are apparently fixed by this, and restarting is faster, too!

markusbattarbee:
I guess the question is - should we merge early and merge often this (as it fixes a bug!) and then try to remember to do the code improvement later, or will the new method be coded, tested, and merged within a reasonable timeframe?

Methings merge this and make a new issue about the restart speed.

ursg:
Agreed. It is definitely better with this fix included than without.

PR #538, opened 2021-02-25
Ensure all process boundary cells at sysboundaries are updated only a…

markusbattarbee:
…fter remote cell information is up-to-date. In the current version, it is possible that some calculation of vlasov system boundaries for cells on process boundaries happens before that cell has had time to transmit it's information, leading to conflicting timesteps being transmitted.

### DIFF
--- a/sysboundary/sysboundary.cpp
+++ b/sysboundary/sysboundary.cpp
@@ -693,7 +693,7 @@ void SysBoundary::applySysBoundaryVlasovConditions(
    
       timer=phiprof::initializeTimer("Wait for receives","MPI","Wait");
       phiprof::start(timer);
-      mpiGrid.wait_remote_neighbor_copy_update_receives(SYSBOUNDARIES_EXTENDED_NEIGHBORHOOD_ID);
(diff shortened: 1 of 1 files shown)


ykempf:
Have you run some test? I suppose a "single process dev" vs. a "multi-process branch" testpackage should return clean?

markusbattarbee:
Yeah didn't run a test yet as am expecting diffs but that's a good idea, will set it up.

markusbattarbee:
I guess this is a non-issue, in that transfers all have time to complete whilst the non-process-boundary sysboundary cells are being processes. That is, even by trying to construct a worst-case-scenario, there were no diffs between the original or the new version, regardless of process counts. So it's safe to merge! If there was some MPI slowdowns, it theoretically could cause those issues - plus the new code is neater (and indeed more correct).

ursg (approved):
I would have given some critical comment about how pre-posted MPI recieve give a performance benefit, and this code is instead doing the whole communication in one block, but then I looked at the code just before and after it. Since those are *anyway* synchronising neighbours through the remotve velocity block list communication, there's no hurt in doing it this way. So thumbs up from me!

PR #540, opened 2021-02-25
Add 2nd and 3rd order terms to volumetric average

lkotipal:
Add 2nd and 3rd order terms to volumetric average. The additional terms were calculated by plugging the expression from Balsara's paper to Mathematica, notebook [here](https://github.com/fmihpc/vlasiator/files/6043295/Balsara.zip).

Not tested or verified yet beyond running, do not merge.

### DIFF
--- a/fieldsolver/ldz_volume.cpp
+++ b/fieldsolver/ldz_volume.cpp
@@ -64,9 +64,12 @@ void calculateVolumeAveragedFields(
             );
             
             // Calculate volume average of B:
-            volGrid0->at(fsgrids::volfields::PERBXVOL) = perturbedCoefficients[Rec::a_0];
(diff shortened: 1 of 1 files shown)


ykempf:
Note that in https://github.com/fmihpc/vlasiator/blob/master/fieldsolver/ldz_volume.cpp#L63 we call 2nd order as we want 2nd-order accurate fields.

We reconstruct 3rd order terms when we compute the Hall term, because we use 3rd-order B to go into the curl to get a 2nd-order J crossed with the 2nd-order B to get the 2nd-order Hall term.

So the question is then, do we want to upgrade to 3rd order Bvol by adding these terms, even if they are computed from a 2nd order accurate B?

PR #541, opened 2021-03-01
vlsvdiff fix

kstppd:
This fixes a bug that would show irrational stats when --diff was not passed as an arguement and handles fsgrid datatypes correctly.


tools/vlsvdiff.cpp
```diff
@@ -1759,6 +1778,10 @@ int main(int argn,char* args[]) {
 
    // 4th arg is its component, 0 for scalars, 2 for z component etc
    uint compToExtract = atoi(argsVector[4].c_str());
+   if (compToExtract > 2){
```
markusbattarbee, line 1781: Incorrect, we do have some variables which can have more components. Not many yet, but the precipitating flux datareducer is one.

MAKE/Makefile.minihpc
```diff
@@ -0,0 +1,84 @@
...
+
+#======== Allocator =========
+#Use TBB malloc
+
+CMP = mpic++
+LNK = mpic++
+
+PAPI_FLAG =
+
+#======== Vectorization ==========
```
markusbattarbee, line 27: You have vectorization set twice

MAKE/Makefile.minihpc
```diff
@@ -0,0 +1,84 @@
...
+ifeq ($(DISTRIBUTION_FP_PRECISION),SPF)
+#Single-precision
+	VECTORCLASS = VEC8F_AGNER
+else
+#Double-precision
+	VECTORCLASS = VEC4D_AGNER
+endif
+
+FLAGS =
+#CXXFLAGS = -I $(HOME)/include  -L $(HOME)/lib -g  -funroll-loops -std=c++0x -fopenmp -W -Wall -pedantic -Wno-unused -fabi-version=0 -mavx
```
markusbattarbee, line 44: We should probably consolidate towards std=c++17

Are you using both this and the Makefile.kstppd files, or is the other one outdated?

Also, note that you have no makefile flags set so although you can turn on testpackage flags for double-precision, it'll still use full optimization.

MAKE/Makefile.minihpc
```diff
@@ -0,0 +1,84 @@
...
+endif
+
+
+#======== Allocator =========
+#Use TBB malloc
+
+CMP = mpic++
+LNK = mpic++
+
+PAPI_FLAG =
```
markusbattarbee, line 25: PAPI_FLAG does not exist (though this does no harm, it can be confusing if you want to turn it on at some point). And if you set PAPI active with CXX_FLAGS, it needs to be lower down (see the makefile cleanup PR to urs' repo)

markusbattarbee (changes requested):
Correction to vector variables with more than 3 components and also (as I was looking at makefiles) improvement suggestions to that as well.

kstppd:
@markusbattarbee  I guess this in now in better shape, right?

markusbattarbee (approved):
Thanks! Much better. :)

FYI, the Makefilecleanup2 PR https://github.com/fmihpc/vlasiator/pull/534/files notes that turning on papi / jemalloc etc for testpackage requires repeating the receipe, like so:

```
#Add PAPI_MEM define to use papi to report memory consumption?
CXXFLAGS +=  -DPAPI_MEM
testpackage: CXXFLAGS +=  -DPAPI_MEM
```
that is, if you want to include them in testpackage (which I think is a good idea). But that's optional at this point.

PR #542, opened 2021-03-02
Include the string lib

henry2004y:
GCC10 complains about missing the inclusion of string header. This fix only helps compilation: there are still errors in the linking stage, probably due to the change in library linking order.

The change has no effect on GCC7. Haven't test on other versions.

### DIFF
--- a/item_storage.h
+++ b/item_storage.h
@@ -24,6 +24,7 @@
 #define ITEM_STORAGE_H
 
 #include <map>
+#include <string>
 
 #include "definitions.h"
 
--- a/mesh_data.h
+++ b/mesh_data.h
@@ -29,6 +29,7 @@
 #define MESH_DATA_H
 
 #include <cstdlib>
+#include <string>
 
 namespace mesh {
 


markusbattarbee:
I think I've encountered this string issue as well before - have you confirmed it's a GCC10 issue, i.e. replicated it on different computers? The alternative is it can be a broken installation on some cluster.

henry2004y:
I don't get a chance to test it on other platforms other than my laptop with Ubuntu 18.04. But in any case, it does not hurt including the string headers. I can only confirm that with these two changes the code can be compiled with GCC10 using my new Makefile.

ykempf:
I can confirm it's also occurring with GCC 10.2.0 on Hawk for instance.

ykempf:
But wasn't that included in some bumper merge coming in soon? Anyway, doesn't hurt so can be merged in my opinion.

ursg:
We encountered the same with GCC10.2 on puhti, too, when we were doing OpenMP GPU tests there. So I guess it's pretty universal.

Since it doesn't seem to break any backwards compatibility, let's merge it right away!

PR #539, opened 2021-02-25
MPI datatype trans optimization

markusbattarbee:
Currently we use a large neighbourhood when doing AMR to ensure we have all necessary cells for the vlasov solver. This update ensures the VDFs are transferred only for those cells which are actually needed.

Diffs (using a 100-second 843-timestep 1 AMR level IPshock run):
```
markusb@turso01:/wrk/users/markusb/AMRpaper2$ /proj/markusb/vlsvdiff_DP AMR1_tp/bulk.0000100.vlsv AMR1B2/bulk.0000100.vlsv proton/vg_rho 0
INFO Reading in two files.
Statistics on file: size 6.95303e-310 min = 4.94066e-324 max = 6.95303e-310 average = 6.95303e-310 standard deviation 0
Statistics on file: size 9000 min = 1.00011e+06 max = 2.27098e+06 average = 1.7351e+06 standard deviation 3091.99
The absolute 0-distance between both datasets is 209.829
The relative 0-distance between both datasets is 9.23959e-05
The average-shifted absolute 0-distance between both datasets is 211.274
The average-shifted relative 0-distance between both datasets is 9.30323e-05
The absolute 1-distance between both datasets is 345875
The relative 1-distance between both datasets is 2.21489e-05
The average-shifted absolute 1-distance between both datasets is 345580
The average-shifted relative 1-distance between both datasets is 2.213e-05
The absolute 2-distance between both datasets is 4771.79
The relative 2-distance between both datasets is 2.85836e-05
The average-shifted absolute 2-distance between both datasets is 4769.82
The average-shifted relative 2-distance between both datasets is 2.85718e-05
```

speedup:
```
AMR1_tp 
(TIME) total run time 3951.27 s, total simulated time 100.044 s
Propagate 3798s, imbalance 0.1668, 3.142e+06 Cells/s
translate proton 2634s, imbalance 0.007595

AMR1B2
(TIME) total run time 3854.56 s, total simulated time 100.044 s
Propagate 3704s, imbalance 0.1687, 3.222e+06 Cells/s
translate proton 2555s, imbalance 0.006713
```

PR #543, opened 2021-03-04
Optimize seeds and mpi trans

markusbattarbee:
Contains the following improvements:
 * Only communicate VDFs for those spatial cells which are required, when doing AMR translation. (E.g. with a VLASOV_STENCIL_WIDTH of 2, we use a neighborhood of 3, which can include up to 24 cells, when we only need up to 8).
 * Calculate pencils only at load balance. Save each direction of pencils at that time.
 * Optimize selection of seed-cellids for pencils in order to execute the minimum required amount of Lagrangian remapping

Supercedes #539 

Testpackage is clean
[log_OptimizeSeedsAndMPITrans_tp.txt](https://github.com/fmihpc/vlasiator/files/6082460/log_OptimizeSeedsAndMPITrans_tp.txt)

_Improvement using a 3D two-AMR-level IPshock run (done for the AMR 6D method paper):_
**spatial space**
long 9.112e+06 Cells/s
MPI 9.883e+06 Cells/s 
seeds 8.578e+06 Cells/s 
both 1.006e+07 Cells/s

AMR_comparison_extra

Interesting that just seed-ids is actually slower, but combining them is faster. 

TBD: performance analysis vs thread counts at scale


markusbattarbee:
Yann's performance tests on Hawk show it's consistently faster:
image

Also, now made long pencils default and moved the input parameter to the AMR group.

ykempf:
Clean diffs, better performance, what is there not to love?

ykempf:
Do we need the new DRO though?

markusbattarbee:
Could be useful for visualization of MPI load in 3D runs? :) And it doesn't do any harm.

markusbattarbee:
Performance after the skip empty vectors enhancement:

image

And the testpackage log:

[log.txt](https://github.com/fmihpc/vlasiator/files/6122825/log.txt)

markusbattarbee:
So any comments on this, do we want to merge into dev (in preparation for 5.2) or fine-tune something more?

ykempf:
The testpackage log looked photoshopped until I saw the last test. ;) Good to go in my opinion, the clean diffs and performance boost speak for themselves.

ykempf (approved):
And now as a formal review. Go!

markusbattarbee:
Fixing this one seedid selection bug seems to also give a nice increase:
before: 9.594e+06 Cells/s  
after:  9.931e+06 Cells/s
Will confirm once test run has completed.

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -716,37 +830,129 @@ void getSeedIds(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGr
...
+      /* -----------------------------------------
+         | A |   | B |   |_|_|_|_|   |   | C |   |
+         |   |   |   |   | | | | |   |   |   |   |
+         -----------------------------------------
+         For optimal pencil generation, we need seedids at A, B, and C.
+         A Is triggered in the first if-clause. Pencils starting from B
+         will be split (but won't cause A to split), and pencils from
+         C will be able to remain again un-split. These checks need to be done
+         only if we aren't already at the maximum refinement level.
+      */
```
ursg, line 858: This is awesome. Really good optimization, and elegant in its implementation, too!

ursg (approved):
Yep, I like this alot. The code is much more readable (on a conceptual level) than before, too.
And the performance results speak for themselves.

markusbattarbee:
Run completed
after: 1.021e+07 Cells/s
this version is the AMRbothblocks2 bar, so we're close to per-cell performance of the old 2D approach now!
AMR_comparison_extra

ursg:
Alright, into the dev branch you go!

Issue #544, opened 2021-03-09
List of cells not updated during initialisation sequence


ykempf:
https://github.com/fmihpc/vlasiator/blob/dev/grid.cpp#L342

This (and later) uses of `cells` are incorrect as there's been a load balance since it was first declared. I think at least.

markusbattarbee:
So in vlasiator.cpp we have
```
const std::vector<CellID>& getLocalCells() {
   return Parameters::localCells;
}

void recalculateLocalCellsCache() {
     {
        vector<CellID> dummy;
        dummy.swap(Parameters::localCells);
     }
   Parameters::localCells = mpiGrid.get_cells();
}
```
which means that doing a new getLocalCells() will do nothing (it re-points the vector to the same address) but indeed a local cell cache recalculation is required - however, that is included inside the balanceLoad() call. So it does currently work correctly, but there is room for improvement in cleaning and consolidating the code regarding how we access local cells.

PR #545, opened 2021-03-09
Local cells caching consolidation

markusbattarbee:
Clarification for non-issue #544
Consolidates calls to local cells to use cached value as often as possible. Also consolidates commented out lines in some projects.

Zero testpackage diffs. Might give a miniscule speedup. (Ignore the value in magnetosphere_small, the comparison dev run was likely abnormally slow due to concurrent `make -j 8` or similar.)

Also includes fsgrid dimension int-conversion cleanup to get rid of compiler warnings.

### DIFF
--- a/grid.cpp
+++ b/grid.cpp
@@ -318,9 +318,10 @@ void initializeGrids(
       exit(1);
    }
    
-   //Balance load before we transfer all data below
+   // Balance load before we transfer all data below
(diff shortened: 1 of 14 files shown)


markusbattarbee:
and the log as well
[log_localcellscache_tp.log](https://github.com/fmihpc/vlasiator/files/6110440/log_localcellscache_tp.log)

PR #546, opened 2021-03-10
Updated mpt gcc makefile for Hawk.

ykempf:


markusbattarbee:
so out of interest, what happened on Hawk to require now building boost by ourselves?

ykempf:
Some maintenance a while back, after that it didn't work. I think the module doesn't even exist any more.

markusbattarbee:
Guess it's good then!

PR #547, opened 2021-03-10
Directory change

henry2004y:
This is the first attempt in moving source codes under the `src` folder. I tested it locally and on Vorna with a selection of tests. 
There are some small changes in including the header files due to the movement. The majority of the original Makefile has been moved under the src folder also to minimize the change, while a new top level Makefile is created trying to make it consistent with previous compilation process.

Looking forward to your feedbacks!


henry2004y:
One thing to notice, however, is that this still has error in the linking stage with GCC10. This issue does not happen with GCC7/8. I haven't tested on other GCC versions or other compilers.

ykempf (commented):
Well this looks like git(hub) understood what was done. :) One could have git mv'ed the Makefile and created a new top-level one but that's really not worth changing, that's not what we'll be diff'ing in the future.

ykempf:
What linking issue(s) do you get? I managed to get rid of many such issues with various compilers in recent months, maybe I can help...

henry2004y:
The linking error for GCC10 looks like this:
```
mpic++ -fopenmp -lgomp -o vlasiator version.o memoryallocation.o backgroundfield.o quadr.o dipole.o linedipole.o vectordipole.o constantfield.o integratefunction.o datareducer.o datareductionoperator.o dro_populations.o amr_refinement_criteria.o donotcompute.o ionosphere.o outflow.o setbyuser.o setmaxwellian.o sysboundary.o sysboundarycondition.o particle_species.o project.o projectTriAxisSearch.o read_gaussian_population.o Alfven.o Diffusion.o Dispersion.o Distributions.o Firehose.o Flowthrough.o Fluctuations.o Harris.o KHB.o Larmor.o Magnetosphere.o MultiPeak.o VelocityBox.o Riemann1.o Shock.o Template.o test_fp.o testAmr.o testHall.o test_trans.o IPShock.o object_wrapper.o verificationLarmor.o Shocktest.o grid.o ioread.o iowrite.o vlasiator.o logger.o common.o parameters.o readparameters.o spatial_cell.o mesh_data_container.o vlasovmover.o londrillo_delzanna.o fs_common.o fs_limiters.o gridGlue.o cpu_acc_intersections.o cpu_acc_map.o cpu_acc_sort_blocks.o cpu_acc_load_blocks.o cpu_acc_semilag.o cpu_acc_transform.o cpu_moments.o cpu_trans_map.o cpu_trans_map_amr.o -L/usr/lib/x86_64-linux-gnu/ -lboost_program_options -L/home/local/hongyang/Vlasiator/vlasiator/lib/zoltan/lib -lzoltan  -L/home/local/hongyang/Vlasiator/vlasiator/lib/phiprof/lib -lphiprof -L/home/local/hongyang/Vlasiator/vlasiator/lib/vlsv -lvlsv -L/home/local/hongyang/Vlasiator/vlasiator/lib/jemalloc/lib -ljemalloc   ldz_magnetic_field.o ldz_volume.o derivatives.o ldz_electric_field.o ldz_hall.o ldz_gradpe.o
/usr/bin/ld: cpu_trans_map.o: undefined reference to symbol 'GOMP_loop_nonmonotonic_guided_start@@GOMP_4.5'
//usr/lib/x86_64-linux-gnu/libgomp.so.1: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
Makefile:479: recipe for target 'vlasiator' failed
make[1]: *** [vlasiator] Error 1
make[1]: Leaving directory '/home/local/hongyang/Vlasiator/vlasiator/src'
Makefile:11: recipe for target 'vlasiator' failed
make: *** [vlasiator] Error 2
```
I searched online but haven't found a solution yet. My guess is that there is some kind of dependency order that should be followed in GCC 10. Since this is not an issue with the new Makefile I have, I don't feel urgent to fix this.

ykempf:
Hmm no that's not one I've had. @ursg will maybe have a solution for us. :)

markusbattarbee:
This does indeed look much better! 

Albeit, the makefile thing as mentioned by yann would have been good as both eVlasiator and the new ionosphere edit the makefile, so now one needs to manually moved over.

Sadly, however, when I merged this on top of my eVlasiator branch, new sysboundaries were *not* transferred to the correct directory:

```
markusb@dx7-flafo-01:/home/local/markusb/git/vlasiator$ ls sysboundary/
static.cpp  static.h  staticionosphere.cpp  staticionosphere.h
```

henry2004y:
I tried to make git recognize my attempt to move the original Makefile into the `src` directory, but I had no luck. The difficulty of keep the original Makefile is that during compilation, the object files will be generated on the top level. With this workaround, the obj files will be in the `src` folder, and I don't have to manually add `src/` to every file. However, if you compile, say `vlsvdiff`, the obj files will still be on the top level for now.

There are some other ways to do this, but again that requires some changes to the Makefile and I am not very familiar with it.

ursg:
> Sadly, however, when I merged this on top of my eVlasiator branch, new sysboundaries were _not_ transferred to the correct directory

Hum. I suppose there is not much we can do about this, given git's architecture.
This means that any old pull request to be merged after this requires an explicit ```git mv``` of any newly added files then. Not great, not terrible. I would say we can live with that.
The actual number of additionally added files tends to be rather small (if even in the massive change that is eVlasiator it ends up being only 5 files that end up having to be moved by hand).

markusbattarbee:
I wanted to test this branch on top of recent optimizations, and for that I also wanted to add a testpackage target, so I added this to the root-level file:
```
+testpackage:
+       cd src; make testpackage
+       mv src/vlasiator .
```

however, this target does not work:
```
$ make clean
rm -rf src/*.o *.o *~ */*~ */*/*~ vlasiator particle_post_pusher particles/*.o
$ make testpackage -j 8
make: 'testpackage' is up to date.
``` 

What gives?

henry2004y:
The reason is that there is a top level directory named `testpackage`, so `make` thinks that if this directory exists, and has no dependencies that are more recent, then this target does not need to be rebuilt.

The solution is to use PHONY targets:
```
.PHONY: testpackage
```

I have pushed another commit to fix this.

henry2004y:
Hmm maybe the problem was that I did it in one commit. If I made two subsequent commits, it may work. Let me try and see how it goes.

PR #550, opened 2021-04-17
Added vlsvdump to tools

kstppd:
This is a little tool I wrote  last Christmas but never pushed it.  It was "inspired" by ncdump and h5dump.  
So given a vlsv file one can do  :
    `vlsvdump  [options] <vlsv file>` to get a quick summary of what the file contains . Of course this can be done with analysator as well but I keep using it so you might find it helpful.

Example output from EGL (`-h` option. With `-hv` it also prints unit,arraysize datasize etc for each variable):

h

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ default: vlasiator
 
 tools: parallel_tools not_parallel_tools
 
-parallel_tools: vlsvextract vlsvdiff
+parallel_tools: vlsvextract vlsvdiff vlsvdump
 
 testpackage: vlasiator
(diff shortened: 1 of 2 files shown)


kstppd:
And I had some issued where it would not print the first two fg_variables. Fixed

kstppd:
So this can now use zstd (I deleted the lz4 functions because with zstd we get compression level control)  to compress datasets. However it is RAM hungry currently as I do compress full datasets. I will make it compress block by block but later.

PR #552, opened 2021-04-23
At ionosphere, retain only boundary-normal components of copied perB values

markusbattarbee:
Testing is underway but looks promising in reverting near-inner boundary behaviour to Vlasiator 4 -style. Perhaps code approach can still be cleaned up?

### DIFF
--- a/fieldsolver/ldz_magnetic_field.cpp
+++ b/fieldsolver/ldz_magnetic_field.cpp
@@ -192,6 +192,36 @@ void propagateSysBoundaryMagneticField(
    }
 }
 
+/*! \brief Low-level magnetic field projection function.
+ *
+ * Projects the magnetic field according to the system boundary conditions.
+ *
+ * \param perBGrid fsGrid holding the perturbed B quantities at runge-kutta t=0
+ * \param perBDt2Grid fsGrid holding the perturbed B quantities at runge-kutta t=0.5
+ * \param technicalGrid fsGrid holding technical information (such as boundary types)
+ * \param i,j,k fsGrid cell coordinates for the current cell
+ * \param sysBoundaries System boundary conditions existing
+ * \param RKCase Element in the enum defining the Runge-Kutta method steps
+ *
+ * \sa propagateMagneticFieldSimple propagateMagneticField
+ */
+void projectSysBoundaryMagneticField(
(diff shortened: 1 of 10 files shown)


fieldsolver/ldz_magnetic_field.cpp
```diff
@@ -192,6 +192,36 @@ void propagateSysBoundaryMagneticField(
...
+ * \param perBGrid fsGrid holding the perturbed B quantities at runge-kutta t=0
+ * \param perBDt2Grid fsGrid holding the perturbed B quantities at runge-kutta t=0.5
+ * \param technicalGrid fsGrid holding technical information (such as boundary types)
+ * \param i,j,k fsGrid cell coordinates for the current cell
+ * \param sysBoundaries System boundary conditions existing
+ * \param RKCase Element in the enum defining the Runge-Kutta method steps
+ *
+ * \sa propagateMagneticFieldSimple propagateMagneticField
+ */
+void projectSysBoundaryMagneticField(
```
ursg, line 208: One thing that really confused me when looking at this diff for the first time is the two different meanings of the word "project" in the source code.

There is Project::setProjectBField, which deals with the magnetic field set by a project.
And now there is projectSysBoundaryMagneticField, which dieals with the magnetic field *projection*.

Should we, to minimize confusion, maybe try to consistently use the form "projection" here? So this identifier would be sysBoundaryMagneticFieldProjection?

sysboundary/ionosphere.cpp
```diff
@@ -707,6 +705,41 @@ namespace SBC {
       }
    }
 
+   /*! We want here to
+    *
+    * -- Retain only the boundary-normal projection of perturbed face B
+    */
+   void Ionosphere::fieldSolverBoundaryCondMagneticFieldProject(
```
ursg, line 712: Likewise here, I suggest Project → Projection

ursg (approved):
Functionally approved!
(But note my comments about the terminology of "Project" vs "Projection")

markusbattarbee:
Fair enough, changed! I'll still recompile and run a testpackage, though of course now we can expect diffs whenever there's an ionosphere involved :)

markusbattarbee:
Confirmed, it works. And diffs only in magnetospheric runs.

PR #556, opened 2021-06-01
Fluxfunction update

lkotipal:
Add two more paths to calculating fluxfunction, ignore paths that touch the inner boundary.


markusbattarbee:
Copying some sample plots over from the Teams discussion:

ABC
AGB
BGA
BCQ

Results look good!

PR #561, opened 2021-07-08
Handle unrecognized input option; simplified readparamters

henry2004y:
I propose two things in this PR:

1. Abort the code when any unrecognized input option from cfg file is detected, so that we can avoid typos and misplacement of  options (e.g. `nVelocitySamples` under multiple commands). The current dev branch allows reading unrecognized options and the user is not aware whether he or she has made a mistake in the cfg file. For example
```yaml
[gridbuilder]
x_length = 20
# comments are not affected
y_lenght = 20 # typo, ignored in current dev, caught in this PR
z_length = 1 
```
This PR will display error message like
```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::program_options::unknown_option> >'
  what():  unrecognised option 'gridbuilder.y_lenght'
```

2.  A simplified and extended implementation of `Readparameters`. The current dev branch applies function overloading over `add` and `get` methods, which requires the programmer to explicitly list all the possible input types. The new implementation uses templates. Besides, error handling is done within the `add` and `get` methods, so we no longer need to return a `bool` type:
```c++
// current dev pattern
int myRank;
if(!Readparameter::get("Template.param", this->param)) {
if(myRank == MASTER_RANK) cerr << __FILE__ << ":" << __LINE__ << " ERROR: This option has not been added!" << endl;
   exit(1);
}

// proposed new approach, error handling is done within the method
Readparameter::get("Template.param", this->param);
```
Since we can specify the error message specifically for the input string argument, this will not create extra burden for debugging.

I have only run one test, `Flowthrough_x_inflow_y_outflow`.

### DIFF
--- a/parameters.cpp
+++ b/parameters.cpp
@@ -20,28 +20,29 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+#include "parameters.h"
+#include "object_wrapper.h"
+#include "particle_species.h"
+#include "readparameters.h"
 #include <algorithm>
 #include <cstdlib>
 #include <iostream>
-#include "parameters.h"
-#include "readparameters.h"
 #include <limits>
 #include <set>
 #include <unistd.h>
-#include "object_wrapper.h"
-#include "particle_species.h"
 
 #ifndef NAN
-   #define NAN 0
+#define NAN 0
 #endif
 
 using namespace std;
 
 typedef Parameters P;
 
-//Using numeric_limits<Real>::max() leads to FP exceptions inside boost programoptions, use a slightly smaller value to avoid...
+// Using numeric_limits<Real>::max() leads to FP exceptions inside boost programoptions, use a slightly smaller value to
+// avoid...
 
-const Real LARGE_REAL=1e20;
+const Real LARGE_REAL = 1e20;
 // Define static members:
 int P::geometry = geometry::XYZ6D;
 Real P::xmin = NAN;
@@ -90,7 +91,7 @@ vector<int> P::systemWriteDistributionWriteZlineStride;
 vector<Real> P::systemWriteDistributionWriteShellRadius;
 vector<int> P::systemWriteDistributionWriteShellStride;
 vector<int> P::systemWrites;
-std::vector<std::pair<std::string,std::string>> P::systemWriteHints;
+std::vector<std::pair<std::string, std::string>> P::systemWriteHints;
 
 Real P::saveRestartWalltimeInterval = -1.0;
 uint P::exitAfterRestarts = numeric_limits<uint>::max();
@@ -120,7 +121,7 @@ Real P::electronDensity = 0.0;
 Real P::electronPTindex = 1.0;
 
 string P::restartFileName = string("");
-bool P::isRestart=false;
+bool P::isRestart = false;
 int P::writeAsFloat = false;
 int P::writeRestartAsFloat = false;
 string P::loadBalanceAlgorithm = string("");
@@ -132,7 +133,7 @@ vector<string> P::diagnosticVariableList;
 
 string P::projectName = string("");
 
-Real P::maxSlAccelerationRotation=10.0;
+Real P::maxSlAccelerationRotation = 10.0;
 Real P::hallMinimumRhom = physicalconstants::MASS_PROTON;
 Real P::hallMinimumRhoq = physicalconstants::CHARGE;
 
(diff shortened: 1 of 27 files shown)


ykempf:
I think we all agree that this is a very nice contribution. I suggest that you run all the testpackage cases to clean up their cfgs, I'd be surprised if they didn't have stale lines.

What do @ursg and @markusbattarbee say, should we use the occasion to clean up the projects' cfgs, or do we leave that for a later stage and/or delegate the task?

parameters.cpp
```diff
@@ -144,216 +145,278 @@ uint P::amrMaxVelocityRefLevel = 0;
...
 Realf P::amrCoarsenLimit = 0.5;
 string P::amrVelRefCriterion = string("");
-int P::amrMaxSpatialRefLevel = 0;
-int P::amrBoxHalfWidthX = 1;
-int P::amrBoxHalfWidthY = 1;
-int P::amrBoxHalfWidthZ = 1;
+uint P::amrMaxSpatialRefLevel = 0;
+uint P::amrBoxHalfWidthX = 1;
+uint P::amrBoxHalfWidthY = 1;
+uint P::amrBoxHalfWidthZ = 1;
```
ykempf, line 151: Let's hope this doesn't introduce int vs uint warnings.

readparameters.cpp
```diff
@@ -20,786 +20,325 @@
...
+                                                        "read options from the run's configuration file arg "
+                                                        "(relative to the current working directory). Options "
+                                                        "given in this file override options given in the user's "
+                                                        "and global configuration files. Options given in "
+                                                        "this override options given in the user's and global "
+                                                        "configuration files. Options given in this file are "
+                                                        "overridden by options given in environment variables "
+                                                        "(prefixed with MAIN_) and the command line");
+   }
+}
```
ykempf, line 344: As @markusbattarbee noted, it owuld be good to add this new line here.

readparameters.h
```diff
@@ -22,57 +22,169 @@
...
+
+   static std::string global_config_file_name;
+   static std::string user_config_file_name;
+   static std::string run_config_file_name;
+
+   static void addDefaultParameters();
 };
 
-#endif
+#endif
```
ykempf, line 190: Also here, new line.

vlasiator.cpp
```diff
@@ -881,9 +882,9 @@ int main(int argn,char* args[]) {
       
       // Apply boundary conditions
       if (P::propagateVlasovTranslation || P::propagateVlasovAcceleration ) {
-         phiprof::start("Update system boundaries (Vlasov post-translation)");
+         phiprof::start("Update system sysBoundaries (Vlasov post-translation)");
          sysBoundaries.applySysBoundaryVlasovConditions(mpiGrid, P::t+0.5*P::dt, false);
-         phiprof::stop("Update system boundaries (Vlasov post-translation)");
+         phiprof::stop("Update system sysBoundaries (Vlasov post-translation)");
          addTimedBarrier("barrier-boundary-conditions");
```
ykempf, line 888: "system sysBoundaries" is a bit of a repeat now, either "system boundaries" or "sysBoundaries" I'd say.

vlasiator.cpp
```diff
@@ -954,9 +955,9 @@ int main(int argn,char* args[]) {
       addTimedBarrier("barrier-after-acceleration");
       
       if (P::propagateVlasovTranslation || P::propagateVlasovAcceleration ) {
-         phiprof::start("Update system boundaries (Vlasov post-acceleration)");
+         phiprof::start("Update system sysBoundaries (Vlasov post-acceleration)");
          sysBoundaries.applySysBoundaryVlasovConditions(mpiGrid, P::t+0.5*P::dt, true);
-         phiprof::stop("Update system boundaries (Vlasov post-acceleration)");
+         phiprof::stop("Update system sysBoundaries (Vlasov post-acceleration)");
```
ykempf, line 960: Ditto

markusbattarbee:
Yep, agreed, this is very nice. Indeed, running the standard testpackage to check that all inputs are still parsed correctly is what's needed.

I don't think we need to keep this PR waiting for cleaning up of less-used old configs - especially as it isn't clear what needs to be done to them. Output variable names have already been updated, and I guess they should be multipop-compatible, unless in the unsupported subdirectory?

One minor point is that in the already-accepted-but-not-merged PR #543 I moved the shortPencils flag under the AMR group, so whichever PR gets merged last needs to be updated to reflect that change.

ykempf (commented):
You'll end up with more code deletions than additions on the balance. But it'll look much better, and the template approach is so much more elegant. We just didn't have/use templates then.

readparameters.h
```diff
@@ -22,57 +22,169 @@
...
 #include <mpi.h>
 #include <stdint.h>
 #include <string>
+#include <typeinfo>
 #include <vector>
 
 #include "common.h"
+#include "version.h"
+
+namespace PO = boost::program_options;
```
ursg, line 41: Hmm, in principle, it is bad style to declare a namespace shorthand like this in a header file.
After all, now every c++ file that includes it will also have the PO name resolved to boost::program_options.

I see that this is unavoidable for a template implementation (unless we want to spell out boost::program_options in every point it is used, which is not pretty, either), but I'm worried that "PO" might be an identifier somebody accidentally uses at one point, and is then confused by the error message.

readparameters.h
```diff
@@ -22,57 +22,169 @@
...
+      MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+      if (rank == MASTER_RANK) {
+         options[name] = "";
+         isOptionParsed[name] = false;
+         descriptions->add_options()(name.c_str(), PO::value<std::string>(&(options[name]))->default_value(defValue),
+                                     desc.c_str());
+      }
+   }
+
+   template <typename T> static void add(const std::string& name, const std::string& desc, const T& defValue) {
```
ursg, line 66: I'm wondering (without having tested): Does this need a separate specialization for ```const char *```? In the current implementation, there's the issue that ```const char *``` is treated as a numeric type, since it can be typeconverted to one. Hence, their string parameter values are typeconverted to a number, which ends up being 0, so the string becomes a nullpointer at runtime.

The "proper" way to use string parameters is ```RP::get("name", "description", std::string("default"));``` at the moment. I wondr if a simple extra specialization would fix this.

readparameters.h
```diff
@@ -22,57 +22,169 @@
...
+         int rank;
+         MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+         if (rank == MASTER_RANK) {
+            std::cerr << __FILE__ << ":" << __LINE__ << name + " not found in the configuration file!" << std::endl;
+            MPI_Abort(MPI_COMM_WORLD, 1);
+         }
+      }
+   }
+
+   template <typename T> static void get(const std::string& name, T& value) {
```
ursg, line 116: Hmm, how does the return type (void) play with our existing use pattern here? In all old code, ```RP::get()``` is called like
```
if(!RP::get("name",target)) {
   std::cerr << "Some error message" << std::endl;
}
```
So if this becomes void, that line will throw an error.

Since you changed all calls to this in all projects anyway (awesome work! I think it makes the code a lot better), but I'm wondering how this'll affect any as-yet-unmerged pull requests. Three options:
 - If we leave this to be a bool, they will probably simply work. But their code will probably never be cleaned up, because why would it?
 - If we leave it as void, they will throw errors, and whoever works on them gets to fix their code once they merge.
 - The ideal case would be to set it to bool and make the compiler emit a warning if it is actually used. Something like the opposite of [the nodiscard attribute](https://en.cppreference.com/w/cpp/language/attributes/nodiscard) or the opposite of [GCC's warn_unused_result](https://gcc.gnu.org/onlinedocs/gcc-4.1.1/gcc/Function-Attributes.html) attribute. Except these don't seem to exist. Ho-hum.

vlasiator.cpp
```diff
@@ -478,7 +479,7 @@ int main(int argn,char* args[]) {
    phiprof::stop("Init DROs");  
    
    // Free up memory:
-   readparameters.finalize();
+   readparameters.~Readparameters();
```
ursg, line 482: Uh-oh, since readparameters is an auto variable, this explicit call of th destructor is a bit dangerous.
Especially since it calls unprotected ```delete``` on some pointers in there. Won't this cause a double free, when th readparameters object is eventually deconstructed at the end of main?
Not that it would really matter if it crashes at the very end, but I would suggest putting a nullpointer check around the deletes in the destructor.

ursg (approved):
Awesome! I really like how this makes the project and boundary code more readable, since it gets rid of lots of boilerplate.
Some commnts below, but they are not showstoppers. We could just merge this as-is, and fix it in dev if any issues with other branches still arise.

parameters.cpp
ursg: I actually believe it fixes them, instead of introducing them.

henry2004y: There are several places where we have loops over AMR widths and levels with local variable of type int. I thought `auto` would do the work, but in fact I had to declare them as `uint` to avoid the warnings. Some of them have already been modified in this commit. I will check again.

readparameters.h
henry2004y: Will do

parameters.cpp
ykempf: Perfect then. Anyway this is used rarely and I don't think it will cause issues as this won't overflow, nor will it be set to negative values.

vlasiator.cpp
henry2004y: I think I messed up with my other branch where `sysboundary` are renamed to `boundary`. I will revert the changes here.

parameters.cpp
```diff
@@ -365,273 +428,276 @@ bool Parameters::getParameters(){
...
+   if (geometryString == "XY4D")
+      P::geometry = geometry::XY4D;
+   else if (geometryString == "XZ4D")
+      P::geometry = geometry::XZ4D;
+   else if (geometryString == "XY5D")
+      P::geometry = geometry::XY5D;
+   else if (geometryString == "XZ5D")
+      P::geometry = geometry::XZ5D;
+   else if (geometryString == "XYZ6D")
+      P::geometry = geometry::XYZ6D;
```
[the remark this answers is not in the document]
henry2004y: Will do

vlasiator.cpp
ykempf: ok

readparameters.h
henry2004y: Indeed. I will remove using namespaces in headers.

readparameters.h
henry2004y: I notice that in some of the projects, the `get` function in current dev is called as
```c++
RP::get("name", target)
```
which does not follow the design pattern but it works as long as no error in reading occurs.

I don't know which is a proper way to handle this backward compatibility like problem. Leave for you to decide.

vlasiator.cpp
henry2004y: This is inherited from the original struct of `Readparameters`, but I agree it is dangerous. I kept it here as it was because maybe freeing the memory manually here would save some memory in the actual time marching loops. 

One thing I noticed in this destructor
```c++
Readparameters::~Readparameters() {
   delete descriptions;
   delete variables;
   descriptions = NULL;
   variables = NULL;
}
```
was that without setting these two pointers to `NULL`, the program will crash in the end. I am not sure why. Will there be an enforced auto free at the end of `main()`? What should I do to make it better?

ursg: The readparameters object itself will fall out of scope at the end of main, so the destructor will run again in that case. If you don't set them to NULL, that will be a certain double free. This probably causes the crash.

We could either guard the deletes:
```c++
if(descriptions != nullptr) {
   delete descriptions;
}
```
or we decide that we don't care about this little bit of wasted memory in a simulation that's multiple gigabytes heavy anyway.

readparameters.h
henry2004y: I haven't seen an example for parsing character type. Perhaps it would be better to treat everything as string first and then do the type conversion. If we want to test this then perhaps we should add a test for this edge case. For now I will ignore this potential issue.

henry2004y:
I think it's ready. Tested on my local ubuntu laptop with testpackage tests. No abort found in existing tests' cfg files.

markusbattarbee:
> I think it's ready. Tested on my local ubuntu laptop with testpackage tests. No abort found in existing tests' cfg files.

Did the test outputs also match the previous version? (i.e. did all the accepted parameters get parsed correctly?)

henry2004y:
> > I think it's ready. Tested on my local ubuntu laptop with testpackage tests. No abort found in existing tests' cfg files.
> 
> Did the test outputs also match the previous version? (i.e. did all the accepted parameters get parsed correctly?)

Not all of them. For example, `transtest_amr` has differences within `1e-8` with single precision VDFs. I am not 100% sure if it works for all after all the changes with the current testpackage settings, because some of them take too long to run on a laptop. So I just decrease the timesteps and compare with that.

ykempf:
Well single precision doesn't help... It should be run with testpackage flags. But reducing the timesteps is OK I think.

On 15 July 2021 08:53:02 EEST, Hongyang Zhou ***@***.***> wrote:
>> > I think it's ready. Tested on my local ubuntu laptop with
>testpackage tests. No abort found in existing tests' cfg files.
>> 
>> Did the test outputs also match the previous version? (i.e. did all
>the accepted parameters get parsed correctly?)
>
>Not all of them. For example, `transtest_amr` has differences within
>`1e-8` with single precision VDFs. I am not 100% sure if it works for
>all after all the changes with the current testpackage settings,
>because some of them take too long to run on a laptop. So I just
>decrease the timesteps and compare with that.
>
>-- 
>https://github.com/fmihpc/vlasiator/pull/561#issuecomment-880416152

-- 

henry2004y:
Tested with testpackage settings, double precision VDFs. Exactly identical output files, except `Magnetosphere_3D_small`, where the saved variables are exactly the same but the system diff showed that the two files are different.

ykempf:
Great then! :)

sysboundary/sysboundary.cpp
```diff
@@ -76,10 +76,13 @@ SysBoundary::~SysBoundary() {
...
-   Readparameters::add("boundaries.periodic_x","If 'yes' the grid is periodic in x-direction. Defaults to 'no'.","no");
-   Readparameters::add("boundaries.periodic_y","If 'yes' the grid is periodic in y-direction. Defaults to 'no'.","no");
-   Readparameters::add("boundaries.periodic_z","If 'yes' the grid is periodic in z-direction. Defaults to 'no'.","no");
+   Readparameters::addComposing(
+       "boundaries.boundary",
+       "List of boundary condition (BC) types to be used. Each boundary condition to be used has to be on a new line "
+       "boundary = YYY. Available options are: Outflow, Ionosphere, Maxwellian, User.");
+   Readparameters::add("boundaries.periodic_x", "Set the grid periodicity in x-direction. true(default)/false.", true);
+   Readparameters::add("boundaries.periodic_y", "Set the grid periodicity in y-direction. true(default)/false.", true);
+   Readparameters::add("boundaries.periodic_z", "Set the grid periodicity in z-direction. true(default)/false.", true);
```
markusbattarbee, line 85: I only now (when trying to bring another branch up-to-date and finding conflicts) see that here the default value of periodicity has changed. 

Although it can make sense (e.g. if no other boundary has been set, then assume periodicity) it should have been documented in the PR. If some other boundary type is set, does that now automatically unset periodicity, or can there be a clash?

PR #563, opened 2021-07-16
Recover BC periodicity inputs as strings

henry2004y:
During the last commit, I happened to use boolean type in `add` and string type in `get` for `boundaries.periodic_x`... This won't crash the code, because if the user inputs boolean types, they are still recognized as strings. However, this causes inconsistencies so we'd better fix it.


ursg:
Ah, good catch. Indeed, no point to change the config file syntax here.

PR #567, opened 2021-08-16
Add boundary exception handling

henry2004y:
Trying to handle #564, with some other minor additional changes:
1. Add exception handling for unknown boundary types during initialization;
2. Remove Boolean type comparison with equal sign;
3. Formatting adjustment 

With the new PR, unrecognized boundary type names listed under command `[boundaries]` in the configuration file will trigger `MPI_abort` during initialization:
```shell
Filtering is off and max number of Passes is = 	 0
Parameter data file (sw1.dat) has 1 values
Parameter data file (sw1.dat) has 1 values
Unknown type of boundary read: haha
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 1 in communicator MPI_COMM_WORLD
with errorcode 1.
```

PR #570, opened 2021-09-05
Time-varying BC

henry2004y:
Working time-varying boundary. Tested on Vorna, Puhti, and local Ubuntu. 0 test diffs, same performance without the new time-varying option.

Major changes:
1. The source files under `sysboundary` is fully reorganized. `setmaxwellian.cpp` --> `maxwellian.cpp`, `setbyuser.cpp` --> `user.cpp`, and a new file `inflow.cpp` is introduced.
2. The logic behind inflow boundary condition is reimplemented. `Maxwellian` is now a subclass of `Inflow`, as compared to the previous logic where `SetMaxwellian` is a subclass of `SetByUser`. This leaves space for potentially other kind of inflow BCs that are not necessarily Maxwellian. The new User class is an empty class for customized boundary condition for future users.
3. New implementation of boundary file reader in `inflow.cpp`, following C++ standard whereas the previous reader follows C. Additional functionalities include
  * Comments are allowed in lines started with `#` (anywhere in the file):
```
# time density temperature Vx Vy Vz Bx By Bz
0.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0	0.0
```
  * Automatic error checking for missing inputs or extra inputs
```
# time density temperature Vx Vy Vz Bx By Bz
0.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0	0.0
2.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0
```
`Error: Missing input values at line 2`
```
# time density temperature Vx Vy Vz Bx By Bz
0.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0	0.0
2.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0   0.0 1.0
```
`Error: Extra input values at line  2`
```
2.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0	0.0
1.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0   0.0
```
`Parameter data must be in ascending order!`
  * Arbitrary number of lines of data
4. Most SBC class methods have now return type `void` instead of `bool`, as the execution will stop as long as any error is encountered.
5. Add a new 1D flowthrough test with varying density at the left boundary.
6. Coding style follows the new Vlasiator format guidelines.

---

One thing to note is that currently if `dynamic==false` and there are multiple time stamp inputs, it seems that only the **first line** (rather than the last line) will be read. I don't know if we want to change this behaviour, or leave as it is to remind the user that this is probably not you want to have multiple inputs in a constant inflow run.


ursg:
Awesome! I like how you have made the semantics of "sysBoundary" vs. "Boundary" a lot more consistent there.

I'm a bit scared that this might make merging the ionosphere implementation a pain in the neck though (as you have naturally also reformatted and updated some things in ionosphere.cpp). Let me first try to see how much pain it is to rebase my ionosphere branch on top of this.

henry2004y:
If it turns out to be too hard to resolve the conflicts by hand, here are some major differences before and after the PR:

1. The method `SBC::getSysBoundary` is now renamed to `SBC::getBoundary`.
2. Bool variable `isPeriodic[3] is renamed to `periodic[3]`, since there's also a helper function `isPeriodic()`.
3. New method `updateState()` is added for time-varying BC. For ionosphere it would be empty.
4. Bool variable `isThisDynamic` is renamed to `anyDynamic`.

These may not be complete, but mostly likely are the main changes for `ionosphere.cpp`.

henry2004y:
I realize that the input file reading function now always searches from the first given timestamp each time it applies. This is not an issue if the number of lines in the input file is small, but certainly can be improved in the next version.

ursg:
After merging #567, this actually does not cleanly apply to master anymore.

I merged the formatting conflicts and sent you a pull request about it, [here](https://github.com/henry2004y/vlasiator/pull/2).

(Equally, I guess this pull request will require lots of fixing up by others, so I'm a little bit apprehnsive about it).

henry2004y:
There is potentially another issue related to electron pressure gradient term. Currently the upstream electron pressure term (which I assume refers to the inflow boundary cells) are still fixed from the electron temperature and density given in the configuration file. Maybe this needs to be adjusted as well if we assume charge quasi-neutrality.

markusbattarbee:
Needs verifying, but I doubt this will in fact be an issue. The upstream value is just a normalization point, and we don't want to change the normalization point throughout the run. So if the actual inflow values change, the local derivatives should still be calculated using the parameter-declared normalization:
https://github.com/fmihpc/vlasiator/blob/4550ac8a0064f88363a8ef0a9d08508a82ebda39/fieldsolver/derivatives.cpp#L64

sysboundary/sysboundary.cpp
```diff
@@ -431,46 +283,43 @@ bool belongsToLayer(const int layer, const int x, const int y, const int z,
...
                                 FsGrid<fsgrids::technical, FS_STENCIL_WIDTH>& technicalGrid) {
-   bool success = true;
-   const vector<CellID>& cells = getLocalCells();
+   vector<CellID> cells = mpiGrid.get_cells();
    auto localSize = technicalGrid.getLocalSize().data();
 
-   /*set all cells to default value, not_sysboundary*/
-   for (uint i = 0; i < cells.size(); i++) {
+   // Set all cells to default value
+   for (uint i = 0; i < cells.size(); i++)
```
markusbattarbee, line 298: Again, I find that for-loops would be better with the curly braces, but it's a matter of preference.
And in matters of preference, I recommend to err on the side of minimazing changes.

sysboundary/sysboundary.cpp
```diff
@@ -531,9 +380,9 @@ bool SysBoundary::classifyCells(dccrg::Dccrg<spatial_cell::SpatialCell, dccrg::C
       }
    }
 
-   // The following is done so that everyone knows their neighbour's
-   // layer flags.  This is needed for the correct use of the system
-   // boundary local communication patterns.
+   // The following is done so that everyone knows their neighbour's layer
+   // flags.  This is needed for the correct use of the boundary local
```
markusbattarbee, line 384: Oops! Here it actually does mean system boundary local communication patterns. Exactly the difference between system boundaries and process boundaries!

sysboundary/sysboundary.cpp
```diff
@@ -646,62 +492,66 @@ bool SysBoundary::classifyCells(dccrg::Dccrg<spatial_cell::SpatialCell, dccrg::C
...
 
-   return success;
+void SysBoundary::updateState(const dccrg::Dccrg<SpatialCell, dccrg::Cartesian_Geometry>& mpiGrid,
+                              FsGrid<std::array<Real, fsgrids::bfield::N_BFIELD>, FS_STENCIL_WIDTH>& perBGrid,
+                              creal t) {
+   list<SBC::SysBoundaryCondition*>::iterator it;
+   if (anyDynamic) {
+      for (it = sysBoundaries.begin(); it != sysBoundaries.end(); it++) {
+         // Skip when not restarting or not requested.
+         if (Parameters::isRestart && !(*it)->doApplyUponRestart())
```
markusbattarbee, line 522: IMPORTANT! 
Are you sure about this?!?
If called within a restarted run but not requesting reapplyuponrestart, then it is in effect turning dynamic boundary conditions off. 

In my opinion, reapplyuponrestart is only for non-dynamic boundary conditions, as... well, dynamic conditions should be updated regardless of it being a restart or not, so such a flag doesn't make sense for those!

Parameters::isRestart is not turned off after initial handling.

Also, at least in the past, reapplyuponrestart affected the call of applyInitialState().

sysboundary/sysboundary.cpp
```diff
@@ -817,22 +660,17 @@ bool getBoundaryCellList(const dccrg::Dccrg<SpatialCell, dccrg::Cartesian_Geomet
...
-   phiprof::start("updateSysBoundariesAfterLoadBalance");
+void SysBoundary::updateBoundariesAfterLoadBalance(dccrg::Dccrg<SpatialCell, dccrg::Cartesian_Geometry>& mpiGrid) {
    vector<uint64_t> local_cells_on_boundary;
    getBoundaryCellList(mpiGrid, mpiGrid.get_cells(), local_cells_on_boundary);
-   // Loop over sysboundaries:
-   for (list<SBC::SysBoundaryCondition*>::iterator it = sysBoundaries.begin(); it != sysBoundaries.end(); ++it) {
-      (*it)->updateSysBoundaryConditionsAfterLoadBalance(mpiGrid, local_cells_on_boundary);
+   // Loop over boundaries:
+   for (std::list<SBC::SysBoundaryCondition*>::iterator it = sysBoundaries.begin(); it != sysBoundaries.end(); ++it) {
+      (*it)->updateBoundaryConditionsAfterLoadBalance(mpiGrid, local_cells_on_boundary);
```
markusbattarbee, line 674: zing

markusbattarbee (changes requested):
Hi, I now went through most of the code. It's a bit challenging to see the actual changes because setmaxwellian.cpp was deleted and instead there's the new maxwellian.cpp so I can't actually see the diffs. I'll look into a possibility of re-implementing the file history. 

There were other issues as well with this PR (both technical like already-merged code showing up as new changes) and philosophical.

Although Urs commented positively about renaming some methods (like coding practive 15,  15. The name of the object is implicit, and should be avoided in a method name.) I strongly belive this does NOT take precedence over not disrupting efforts of other people. This is not the only PR dealing with SBCs, we have several! Although they all probably break each other to some degree, each contributor should strive to minimize disruption.

Also, in general, if you want to do whitespace / line return placing / cosmetic PRs, do them separately. Even with turning whitespace diffs off in the github interface, this was a chore to read because of all the changed line breaks - and that does not affect code behaviour.

Also, I see that like you stated, you've changed many functions to return voids. Did you verify carefully that we don't lose any error-catching capability here?

markusbattarbee:
I had posted this on Flowdock but storing it here so it won't get lost:

> I was able to use these instructions [https://stackoverflow.com/questions/42287533/restore-deleted-file-with-history-in-git](https://stackoverflow.com/questions/42287533/restore-deleted-file-with-history-in-git) to restore history to a deleted / renamed file. Just follow the instructions (including the odd double-copying of files) and it should work! Just be careful about renaming afterwards as that may again do strange things to the history.

henry2004y:
And a note for myself: `git log --follow` can display the whole history of renamed/moved files.

PR #572, opened 2021-09-13
Turso03 makefile updates

ykempf:
Please someone report successful building and running to check that permissions are OK for other users than me.


MAKE/Makefile.carrington_gcc_openmpi
```diff
@@ -11,20 +11,21 @@
 # module load gnu9/9.3.0
 # module load openmpi4/4.0.5
 # module load pmix/3.1.4
-
+# Oneliner for command line:
+# module purge ; module load gnu9/9.3.0 ; module load openmpi4/4.0.5 ; module load pmix/3.1.4
```
markusbattarbee, line 15: Add: 
remember to compile inside interactive session:
/usr/bin/srun -M carrington --interactive -n 16 --mem=4G --time=00:20:00 -p short --pty bash

MAKE/Makefile.vorna_gcc
```diff
@@ -1,9 +1,11 @@
...
 #
 # Reccomended modules in .bashrc:
-# module load GCCcore/8.3.0
-# module load OpenMPI/3.1.4-GCC-8.3.0
-# module load Eigen
+# module purge
+# module load GCC/10.2.0
+# module load OpenMPI/4.0.5-GCC-10.2.0
+# Same in one line to past on the command line:
+# module purge ; module load GCC/10.2.0 ; module load OpenMPI/4.0.5-GCC-10.2.0
```
markusbattarbee, line 9: Add:
remember to compile inside interactive session:
/usr/bin/srun -M vorna --interactive -n 16 --mem=4G --time=00:20:00 -p test --pty bash

MAKE/Makefile.vorna_gcc
```diff
@@ -57,38 +59,36 @@ testpackage: CXXFLAGS += -DUSE_JEMALLOC -DJEMALLOC_NO_DEMANGLE
...
 
 #header libraries
-INC_FSGRID = -I$(LIBRARY_PREFIX)/fsgrid/
-INC_DCCRG = -I$(LIBRARY_PREFIX)/dccrg/
-INC_VECTORCLASS = -I$(LIBRARY_PREFIX)/vectorclass/
-
+INC_FSGRID = -I/proj/group/spacephysics/libraries/fsgrid/
+INC_DCCRG = -I/proj/group/spacephysics/libraries/dccrg/
+INC_VECTORCLASS = -I/proj/group/spacephysics/libraries/vectorclass/
+INC_EIGEN = -I/proj/group/spacephysics/libraries/
```
markusbattarbee, line 92: I think these should use the LIBRARY_PREFIX variable

markusbattarbee (changes requested):
Was able to compile and run with both of these. I'd also recommend purging all old vorna and carrington makefiles in this PR.

MAKE/Makefile.carrington_gcc_openmpi
ykempf: Actually there the "old" interactive call still works. But good reminder. Shall I instead add "remember to compile in an interactive session (see wiki for current best practices)"?

ykempf:
So shall I nuke MAKE/Makefile.vorna_intel too or do you/someone want to test an intel version too?

MAKE/Makefile.carrington_gcc_openmpi
markusbattarbee: sounds good (and make sure the wiki is up to date :) )

markusbattarbee:
> So shall I nuke MAKE/Makefile.vorna_intel too or do you/someone want to test an intel version too?

probably not worth the hassle. If we need to, it should be a clean start anyway as it's a completely new OS release.

markusbattarbee:
right, so I see this has the 2D shells VDF output in it - noteworthy that it is not the same behaviour as in earlier 3D runs, as instead of having the kinda-arbitrary-but-simple cellID stride, it also for 3D runs calculates the dominant coordinate and uses it as the stride.

ykempf:
Oh no. I didn't notice the cherry-pick made it in.

MAKE/Makefile.vorna_gcc
ykempf: Done

MAKE/Makefile.carrington_gcc_openmpi
ykempf: Juha confirmed that interactive is deprecated also on Carrington.

PR #573, opened 2021-09-20
Set the tapering with an outer and inner radius and allow temperature to be tapered too

ykempf:
- Should I move them all to Magnetosphere instead of having the tapering under the ionosphere?
- Remind me that I should update the projects and testpackage cfgs before merging if I forget.
- I don't intend to test the linear tapering, and we haven't used it since the year of the sword & helmet. Shall I remove it altogether?

Not to be merged yet, for discussion so far, will be test-run in the very near future.

### DIFF
--- a/projects/Magnetosphere/Magnetosphere.cpp
+++ b/projects/Magnetosphere/Magnetosphere.cpp
@@ -183,6 +183,12 @@ namespace projects {
             }
             abort();
          }
+         if(sP.taperOuterRadius > 0 && sP.taperOuterRadius <= this->ionosphereRadius) {
+            if(myRank == MASTER_RANK) {
+               cerr << "Error: " << pop << "_Magnetosphere.taperOuterRadius is non-zero yet smaller than ionosphere.radius! Aborting." << endl;
+            }
+            abort();
+         }
          if(sP.taperInnerRadius == 0 && sP.taperOuterRadius > 0) {
             if(myRank == MASTER_RANK) {
                cerr << "Warning: " << pop << "_Magnetosphere.taperInnerRadius is zero (default), now setting this to the same value as ionosphere.radius, that is " << this->ionosphereRadius << ". Set/change " << pop << "_Magnetosphere.taperInnerRadius if this is not the expected behavior." << endl;
--- a/projects/Magnetosphere/Magnetosphere.cpp
+++ b/projects/Magnetosphere/Magnetosphere.cpp
@@ -96,6 +96,9 @@ namespace projects {
    }
    
    void Magnetosphere::getParameters(){
+      int myRank;
+      MPI_Comm_rank(MPI_COMM_WORLD,&myRank);
+
       Project::getParameters();
 
       Real dummy;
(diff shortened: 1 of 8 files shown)


ykempf:
This ran on Mahti now, maybe I should generate one plot? :)

projects/Magnetosphere/Magnetosphere.cpp
```diff
@@ -158,12 +158,18 @@ namespace projects {
...
          RP::get(pop + "_ionosphere.VX0", sP.ionosphereV0[0]);
          RP::get(pop + "_ionosphere.VY0", sP.ionosphereV0[1]);
          RP::get(pop + "_ionosphere.VZ0", sP.ionosphereV0[2]);
-         RP::get(pop + "_ionosphere.taperRadius", sP.ionosphereTaperRadius);
+         RP::get(pop + "_ionosphere.taperInnerRadius", sP.ionosphereTaperInnerRadius);
+         RP::get(pop + "_ionosphere.taperOuterRadius", sP.ionosphereTaperOuterRadius);
+         if(sP.ionosphereTaperInnerRadius > sP.ionosphereTaperOuterRadius) {
+            cerr << "Error: " << pop << "_ionosphere.taperInnerRadius should be <= taperOuterRadius" << endl;
+            abort();
+         }
```
markusbattarbee, line 171: Great to finally have ionosphere.T!
Perhaps we should add failsafes to both this and the inner taper radius: if they aren't set in the .cfg (i.e. are still zero), then set T to mangetosphere.T and innertaperradius to the ionosphere radius. That way when people re-use older cfg:s things won't break unexpectedly.

ykempf: Like so?

ykempf:
I'll update the cfgs as per bullet 2 up there, what about the other two bullets?

projects/Magnetosphere/Magnetosphere.cpp
markusbattarbee: still needed: if innertaperradius==0, set it to ionosphereradius

ykempf: I was thinking that we might want to have a partial profile and set the inner radius on purpose < ionosphere.radius. If that's not the case I can force it, sure.

markusbattarbee:
Agreed that linear tapering can be purged.

Moving to magnetosphere would make sense as it is in fact not inner boundary stuff that we're dealing with, but of course needs careful checking.

projects/Magnetosphere/Magnetosphere.cpp
markusbattarbee: well, we can still allow it, as long as the innertaperradius isn't zero. But by setting the default to zero and in that case changing it to our best guess (ionosphere radius), I think we're maintaining compatibility with the previous approach in a sane way.

ykempf: Not quite because of boost aborting upon unknown options (the old <pop>_ionosphere.taperRadius). But I agree with your suggestion, I'll add that.

ykempf:
Like so?

projects/Magnetosphere/Magnetosphere.cpp
```diff
@@ -169,12 +171,24 @@ namespace projects {
...
          if(sP.taperInnerRadius > sP.taperOuterRadius) {
             if(myRank == MASTER_RANK) {
-               cerr << "Error: " << pop << "_Magnetosphere.taperInnerRadius should be <= taperOuterRadius" << endl;
+               cerr << "Error: " << pop << "_Magnetosphere.taperInnerRadius should be <= taperOuterRadius! Aborting." << endl;
             }
             abort();
          }
+         if(sP.taperInnerRadius == 0 && sP.taperOuterRadius > 0) {
```
markusbattarbee, line 186: If taperInnerRadius==0 and taperOuterRadius < ionosphereRadius, this might end up pear-shaped?

ykempf: Hence the test at 174 for you. :)

markusbattarbee: but that's earlier - this check might put us in that region after that check has occurred

ykempf: But if the 174 triggers we abort. I don't follow.

ykempf: Aaah, now. So one more check to ensure taperOuterRadius is !=0 and > ionosphereRadius?

sysboundary/ionosphere.cpp
```diff
@@ -95,7 +95,7 @@ namespace SBC {
         Readparameters::get(pop + "_ionosphere.VY0", sP.V0[1]);
         Readparameters::get(pop + "_ionosphere.VZ0", sP.V0[2]);
         Readparameters::get(pop + "_ionosphere.fluffiness", sP.fluffiness);
-        Readparameters::get(pop + "_Magnetosphere.T", sP.T);
+        Readparameters::get(pop + "_ionosphere.T", sP.T);
```
ursg, line 98: It is a bit confusing that is gets set to sP.T here but sP.ionosphere_T in Magnetosphere.cpp. Is this parameter fetch even still required, if the whole actual handling happens in the magnetosphere project?

markusbattarbee: well, ionosphere.T did not exist here, but it's the only temperature that ionosphere needs, so sP.T makes sense. Whereas in Magnetosphere.cpp it's the secondary temperature that we taper towards, different from the actual sP.T which is magnetosphere.T... so makes sense to me.

ykempf: Also, I think you are confusing the internal naming with the cfg naming?

ykempf: Yes, we need the ionosphere's T for the vdf generation.

ursg:
Turns out, this blows up my Ionosphere test runs almost instantly and I don't quite understand why yet:
E_crash

I had replaced ionosphere.taperRadius by proton_magnetosphere.taperOuterRadius and set proton_magnetosphere.taperInnerRadius to ionosphere.radius. But that appears not to work out-of-the-box.

ursg:
Even the initial grid already has rho going NaN for me:
rho_crash

ursg:
Ah, and now I found the culprit: while the code of this pull requests fixes ionosphere.T to be identical to *_magnetosphere.T *for the magnetosphric temperature*, a separate piece of code in sysboundary/ionosphere.cpp doesn't. So the ionosphere boundary cells get filled with zero temperature, which leads to division by zero.

ursg:
Ok, I fixed this with a [corresponding check](https://github.com/ursg/vlasiator/commit/ee8ddd5e3de0f486e30e6864d81be8c8c018fbba) in the ionosphere implementation and can confirm that it doesn't crash anymore.

PR #574, opened 2021-09-28
Update of Dispersion project to modern standards.

ykempf:
Now passing perBGrid to the hook() function to allow direct output of perB components.

I wrote this update as I ran some dispersion tests lately (which ended up to be the chase for the lasers in parallel dispersion). So it runs OK.


ursg:
I believe nobody else except you has ever evn tried to run that project, so into the dev branch we go!

PR #575, opened 2021-09-28
Whistler faster dispersion

ykempf:
So this seems to have helped a lot, even if some issues seem to have stemmed from the divergent dipole slicing.

In 1D parallel dispersion there is no difference, as seen in my recent tests.

But this is completely untested in 3D so far. What/how do we want to test for that?

### DIFF
--- a/fieldsolver/ldz_electric_field.cpp
+++ b/fieldsolver/ldz_electric_field.cpp
@@ -169,7 +169,11 @@ void calculateWaveSpeedYZ(
    // for details.
    const Real vA2 = divideIfNonZero(Bmag2, pc::MU_0*rhom); // Alfven speed
    const Real vS2 = divideIfNonZero(p11+p22+p33, 2.0*rhom); // sound speed, adiabatic coefficient 3/2, P=1/3*trace in sound speed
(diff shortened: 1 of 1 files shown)


ursg:
Hmm, tricky. The differences it causes are mostly really subtle *except* for the cases where it then does not blow up.

Since we don't have a 3D test case that would have blown up so far, the only thing we can heuristically test for is whether it otherwise disrupts our 3d sim geometry. Since I'm anyway running medium-sized 3D runs for ionosphere testing, how about I merge it there and see if everything looks ok?

ykempf:
Sure, go for it.

ursg:
My preliminary results from testing with the ionosphere runs is that they alle crashed instantly (with fieldsolver dt going to zero) when restarted with this pull request. This does not look too promising for 3D runs.
Will test in more detail tomorrow.

ursg:
The crashes turned out to actually be caused by an [unrelated bug](https://github.com/fmihpc/vlasiator/pull/573#issuecomment-930380716), so after that was fixed, I continued testing. Now it sems the runs are going fine and stable. While that still does not constitute an indicator thar the whistler bug would be fixed completely, at least it seems that this pull doesn't break anything.

ykempf:
So apparently Jonas ran a test that had more blips in 2D... So I'm not sure now. The fix worked well, but in the as-we-now-know pathological dipole slice setup.

ykempf:
Hmm, either we take this in, or not, let's not make it too complicated.

PR #576, opened 2021-09-28
2dsphericalshell

markusbattarbee:
Implemented 2D spherical shells and changed the stride in 3D to be evaluated based on dominant coordinate direction instead of a simple cellid-stride


ursg:
Interesting idea to use the dominant coordinate. Its not the prettiest piece of code (couldn't the same have been implemnted by just sorting the coordinates in a temporary array and then using the [0] member of that), but just fine if it works.

What do the resulting patterns look like? Do you have a plot of that already?

ursg:
(This code also reminds me a lot of [The midpoint circle algorithm](https://en.wikipedia.org/wiki/Midpoint_circle_algorithm), but that one is not, actually, directly applicable to our needs here)

markusbattarbee:
yeah, I do have test runs form Vorna. Here are 2D polar and equatorial plots. Will post the 3D plots when I next visit the office. As can be seen, there are some areas where two neighboring cells are both saved, but there's no obvious way of getting past that. I guess one option would be to step through the midpoint circle method and save every STRIDEth cellid into a masking array, and I'm not quite sure it's worth it...

equa3_map_fg_b_0000000
polar3_map_fg_b_0000000

ursg:
Yeah, not certain if changing the implementation to a "maybe slightly better" version is worth it, either. This one appears to produce reasonably-spaced sample points, and that is, in the end, what we aim for.
Nice looking patterns, indeed!

markusbattarbee:
Here from a 3D simulation: one shell
Screenshot from 2021-09-30 09-03-57


plus just one octant visible
Screenshot from 2021-09-30 09-05-39

ursg:
Ah well, yep, that shows some neighbouring cells still being fSaved, but it still makes for a useful-looking pattern.

So, since this apparently works, should we just save it right away? Or does the changing placement of fSaved sites somehow affect anybody else's work? (In particular, I'm wondering about the carrington runs -- would it be bad if they now changed inbetween)?

markusbattarbee:
The reason I made this PR is that the Carrington runs have already been using this. 
But, in fact, yesterday evening I had an idea how to implement this more elegantly - stay tuned...

markusbattarbee:
First 2D plots:
Now there's two options, one is the straightforward:
polar2
equa2

 the second forces shells on at the 45 degree angle where it's at the boundary between treatment regions and can otherwise have a longer gap:
polar3
equa3

markusbattarbee:
Improved selection logic at 45 degree lines in 2D runs:

polar2
equa2

Much reworked 3D: Had to resort to some special case checks but I think it's pretty decent?

Screenshot 2021-10-06 at 12 31 23
Screenshot 2021-10-06 at 12 31 47

markusbattarbee:
ping @ursg

ursg:
This looks really nice! And the code is pretty elegant, too.
I'm cool with it, if you are. :)

PR #577, opened 2021-09-28
Testpackage and tools

markusbattarbee:
New testpackage running scripts for carrington and vorna (now set the binaries in the job script)
added an INC_BOOST to allow making tools


ursg:
But why does this pull request delete the job scripts for puhti (which is still very much alive and kicking) and docker (which, even though we do not regularly use, is still a good idea in principle)?

markusbattarbee:
The puhti script removal was a whoopsie-daisy, I reinstated it. Never heard of anyone using docker, but back it goes as well.

PR #581, opened 2021-10-06
Support for proper multi-unit read of fsgrids.

ykempf:
Seems to work, cleanup TBD as well as testing of when the units get chunked by VLSV,
and the float conversion case. Optimisation of the vector of vectors could be done too.
Works with my branch multiread_with_offsets of VLSV at the moment.

WIP, up for discussions. Do not merge!


ykempf:
Four-way testing of the normal and small-chunk VLSV variants (cf. https://github.com/fmihpc/vlsv/pull/44) for a small 3D Magnetosphere (the testpackage one but with 30³ base grid instead of 22³), run for 3 s, restarted either on the same or a 4x larger task count, run 2 s further, with `testpackage` flags, yield relative diffs of ~5e-11 at the end in proton/vg_vz, a few 1e-13 in proton/vg_rho and 1e-15 in fg_b.

ykempf:
So this latest one compiles but I don't remember what it takes to activate that conversion and test it. Do you have a test rig for that @ursg?

ykempf:
Further question: should we recompute the decomposition shebang or shall I cache it? It might start to take a few milliseconds more when there's 100k tasks?

markusbattarbee (changes requested):
Logging it here as well.
https://github.com/ykempf/vlasiator/blob/415a3a8a876b7fa6fc39e84cfbc3da06a00b5b86/ioread.cpp#L966
needs to be adapted if convertFloatType is set.

Also, though we don't really do it, the code will fail if we try to read from double into float.

I guess the overlap calculations could be done just once and stored for the second round, to clean up the code, but works like this as well.

ioread.cpp
```diff
@@ -790,14 +793,17 @@ template<unsigned long int N> bool readFsGridVariable(
    attribs.push_back(make_pair("name",variableName));
    attribs.push_back(make_pair("mesh","fsgrid"));
 
+   phiprof::start("getArrayInfo");
    if (file.getArrayInfo("VARIABLE",attribs,arraySize,vectorSize,dataType,byteSize) == false) {
       logFile << "(RESTART)  ERROR: Failed to read " << endl << write;
       return false;
```
ursg, line 799: This return fails to close the phiprof group just opened above.
(I suppose we are anyway going to fail in this case, but it would be good style to always have closed groups, also in crashing cases)

ursg (changes requested):
So far, so good. I suppose this has been semi-production-tested well enough to be merged soon?

markusbattarbee:
Note: In testing this on Lumi we actually didn't achieve the speedups we were hoping for. 
Also, IIRC this is dependent on 
https://github.com/fmihpc/vlsv/pull/44

ykempf:
So it looks like this might help us not breaking Mahti's lustre. Can we merge it?

ykempf:
up

markusbattarbee:
Leo's timer object fix is about to go in - how about this wait until that, and along with having conflicts fixed, move to the timer object formalism plus better group definitions (looking at names like "task shebang 2")

lkotipal:
Timers were merged into phiprof/master so updating this shouldn't depend on #721. The Carrington makefile has the correct version.

ioread.cpp
```diff
@@ -1221,9 +1298,11 @@ bool exec_readGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
    phiprof::start("readFsGrid");
    // Read fsgrid data back in
    int fsgridInputRanks=0;
+   phiprof::start("readScalarParameter");
    if(readScalarParameter(file,"numWritingRanks",fsgridInputRanks, MASTER_RANK, MPI_COMM_WORLD) == false) {
       exitOnError(false, "(RESTART) FSGrid writing rank number not found in restart file", MPI_COMM_WORLD);
```
lkotipal, line 1303: Early return doesn't stop timers. Recommended to change all timers to objects i.e. `phiprof::start(label)` -> `phiprof::Timer foo {label}` and phiprof::stop(label, args)` -> `foo.stop(args)`.

ykempf:
So I suppose the strategy is "change phiprof timers to objects in the code regions you touch"?

lkotipal:
> So I suppose the strategy is "change phiprof timers to objects in the code regions you touch"?

I think that's the way to go, yea. Old timers have already been purged by #721 so this basically just means updating current and future PRs to the same standard.

ioread.cpp
ykempf: But wait this guy I inserted and I can make it an object, but the other guys before and after this are also old-style. Why aren't they changed?

ykempf: And there's early exit ones there too.

lkotipal: Did you merge dev first? Ah, I suppose some might've slipped in between my updates to timers and merging that to dev...

ykempf:
Love this, it's too old to have CI actions. :)

ioread.cpp
```diff
@@ -901,30 +923,91 @@ template<unsigned long int N> bool readFsGridVariable(
...
+         // Read into buffer
+         phiprof::Timer tMemcpy {"memcpy"};
+
+         // Read every source rank that we have an overlap with.
+         if(overlapSize[0]*overlapSize[1]*overlapSize[2] > 0) {
+            if(convertFloatType) {
                for(uint64_t i=0; i< thatTasksSize[0]*thatTasksSize[1]*thatTasksSize[2]*N; i++) {
-                  buffer[i]=readBuffer[i];
+                  vectorOfBuffers.at(task).at(i)=vectorOfFloatBuffers.at(task).at(i);
                }
```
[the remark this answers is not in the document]
ykempf: This is converting the float buffer to double/Real again. But hey we're `write_as_float = 1` indeed.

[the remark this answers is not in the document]
kstppd: For some reason my diff on web VS code on github is completely broken.

ioread.cpp
```diff
@@ -936,15 +1019,12 @@ template<unsigned long int N> bool readFsGridVariable(
...
-                     memcpy(targetGrid.get(x - localStart[0], y - localStart[1], z - localStart[2]), &buffer[index*N], N*sizeof(Real));
+                     memcpy(targetGrid.get(x - localStart[0], y - localStart[1], z - localStart[2]), &(vectorOfBuffers.at(task).at(index*N)), N*sizeof(Real));
                   }
                }
             }
-         } else {
-            // If we don't overlap, just perform a dummy read.
-            file.endMultiread(fileOffset);
          }
-         fileOffset += thatTasksSize[0] * thatTasksSize[1] * thatTasksSize[2];
```
ykempf, line 947: See? We *were* doing something to that fileOffset, we don't any more. But hmm we should start at fileOffset 0 and use the manual offsets I suppose. If I hadn't botched this, that is...

PR #582, opened 2021-10-07
Maxwellian Inflow  Fix

kstppd:


kstppd:
I do not know exactly what happened :D , but I think this is only what we want

markusbattarbee:
Yep, looks good now, I say we merge so this bug is cleared. 

As a side-note,  if we wanted to inflow a bi- or tri-maxwellian, this methodology wouldn't really work, and we should use something akin to https://github.com/fmihpc/vlasiator/blob/82a5bdf17eb3fcebdd397d4fa834099b0371de04/projects/projectTriAxisSearch.cpp#L59

ykempf:
Yes, but that would also be lacking parameters at the moment. This is equivalent to the base project class and indeed triAxisSearch would be the next step. Hongyang might have done something to that in the time-varying conditions.

PR #584, opened 2021-10-11
Fix array indices in Firehose project. This is not Fortran.

ursg:
Also make the profile function *actually* do its job there.

Untested. (But seriously...)


ursg:
Roger that. Here's to you, future PhD student who reads this pull request in 5 years and smirks. :)

Issue #586, opened 2021-10-18
Spectacular physical failure when the distribution reaches the v-space limits


ykempf:
Since the use of the semi-Lagrangian solver the vdf cannot exit the v-space limits gracefully. When the vdf reaches the wall there seems to be some folding back into the box which results in spectacular plots but useless physics.
Ideas (suggestions welcome):
- Warn the user when the limits are reached.
- Bail out (optionally) when the limits are reached.
- Make it a user-defined parameter, how many blocks away from the limit to warn/bailout.

PR #587, opened 2021-10-18
Vspace wall bailout

ykempf:
Addresses issue #586.


henry2004y:
It looks good to me.

ykempf:
Here's the new version. Compiles on Carrington.

markusbattarbee:
Ok, looks good now! Though should we make the default value 1, so that it actually gets used as well? :)

ursg:
Yep, fine by me!

PR #588, opened 2021-10-25
Replace std::unordered_map with a custom open bucket fibonacci-hashed hashtable in vmesh.

ursg:
This is the result of me scratching my head about our profiling results for quite a while: Why do we spend so much time in our hash function? Because we need to resolve sparse velocity space accesses quite a lot. (The actual solution was triggered in my head by reading [these tweets](https://twitter.com/rygorous/status/1446980556323909633) and [this article](https://www.andreinc.net/2021/10/02/implementing-hash-tables-in-c-part-1).)

The std::unordered_map that we normally use in that spot uses a trivial "divisional" hash function, in which the global ID is simply reduced as ```arrayIndex = globalID % arraySize;```. The result is directly used for a memory acces, so the modulo division is not properly pipelined and thus slow.
Furthermore, std::unordered_map implements "closed buckets", in which every hash bucket holds a linked list of members. In hash collisions, this list grows. Since we have typically continuous ranges of velocity space GIDs in our box, and the divisional hash function maps them to continuous ranges of array indices, we are quite heavily hash-colliding. Hence we jump along linked lists a lot, which is very cache inefficient, and we end up memory latency bound.

This pull request fixes both of these issues:
 - The hash function is replaced by a fibonacci hash, which only requires a single multiplication and some bit shifts. This is much faster.
 - The properties of the fibonnaci hash also spread continuous GID ranges more evenly in the hash buckes, leading to much fewer collisions.
 - Most importantly however, the new hash table uses open buckets: Every bucket only contains a single value, and in a hash collision, the second value leaks over into the next bucket. Because they are lying in a compact array in memory and no longer require pointer chasing, even a multiple-way hash collision remains efficient, as the following buckets are still resident in L1 cache. How leaky a bucket can be is a template argument to the hash table (default value of 8. I haven't even benchmarked these yet.)

As an added bonus, this data structure should also nicely lend itself to GPU implementation, since no pointers need to be translated.

Testpackage shows between 20% and 80% performance gain on my work machine.

Please test thoroughly.


markusbattarbee (approved):
My own tests showed zero diffs as well, so looks good. Only thing I wonder is if we would want to make the hashtable iterator calls explicit instead of using auto (so that porting to GPUs will be smoother) but that can be tackled when it's relevant.

ursg:
This is now good to go. I'm testing whether any additional performance benefits can be gained by replacing some of DCCRG's internal hashtables by the same, but that will be a separate pull request.

PR #590, opened 2021-11-22
LUMI-r and LUMI-ds instead of old Sisu in transferPraceData.sh

ykempf:
I left Sisu-g in so we have a trace of how it was done, in case we need it in future. To be tested shortly with @JonasSuni.


ykempf:
Works.

ursg:
Great, let's have it then!

PR #592, opened 2021-11-25
Fix bug in memory bailout determination.

ykempf:


markusbattarbee (approved):
good!

ykempf:
Ah I forgot, should I add a comment about the potential pitfall of MASTER_RANK and the zero ranks only matching by "coincidence"?

ursg:
Heh. Weird that this ever caused problems, as the double variables *should* always have been zero-initialized. But maybe a overzealous compiler optimization decided to axe that?
Anyway, I approve!

And I kinda don't believe that we will ever have a different MASTER_RANK than zero, given that the mpi standard prescribes that.

ykempf:
FWIW we manually declare MASTER_RANK to be 0, but I guarantee many places will break where we assume it to be zero, so no use to ever try to change that.

PR #593, opened 2021-11-29
Mpiio restart hints

ykempf:
Compiles... To be tested on LUMI at least.


markusbattarbee:
Looks sane, let's test

ykempf:
Remind me to make consistent what we call `*_bulk_*` and what we call `*_system_* ` options, if I forget.

ykempf:
There...

ykempf:
Now we have the cfg options called `write_*_stripe_factor` but it sets the hint `striping_factor` or we can set `striping_unit` with the manual MPI hint cfg flags. Shall I make it consistent using `striping`, even though that breaks compatibility of old cfgs with the `write_restart_stripe_factor` option set explicitly?

iowrite.cpp
```diff
@@ -1373,6 +1373,18 @@ bool writeRestart(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
...
+      MPI_Info_create(&MPIinfo);
+
+      for (std::vector<std::pair<std::string,std::string>>::const_iterator it = P::restartWriteHints.begin();
+           it != P::restartWriteHints.end();
+           it++)
+      {
+         MPI_Info_set(MPIinfo, it->first.c_str(), it->second.c_str());
+      }
+   }
    if (stripe == 0 || stripe < -1){
```
markusbattarbee, line 1388: now if stripe is zero or smaller than -1, the manually provided hints will be removed.

markusbattarbee: also on line 1192 for bulk files.

markusbattarbee (commented):
I don't think it's worth changing the old config values to say "striping" instead of "stripe", as there's no fixed rule between the two in external documentation either.

iowrite.cpp
ykempf: So the subsequent `MPI_Info_set` call would override the previous `MPI_Info_set(MPIinfo, it->first.c_str(), it->second.c_str());` calls?

ykempf: No OK sorry I read too quickly.

ykempf:
Aha, we can inherit the stripe count if we set it manually to 0 in the cfg instead of the default that goes to -1. Maybe worth changing the default then.

markusbattarbee:
So wasn't it that we need to re-allow passing stripe=0, and make that the default as well?

parameters.cpp
```diff
@@ -92,12 +92,13 @@ vector<Real> P::systemWriteDistributionWriteShellRadius;
 vector<int> P::systemWriteDistributionWriteShellStride;
 vector<int> P::systemWrites;
 vector<pair<string, string>> P::systemWriteHints;
+vector<pair<string, string>> P::restartWriteHints;
 
 Real P::saveRestartWalltimeInterval = -1.0;
 uint P::exitAfterRestarts = numeric_limits<uint>::max();
 uint64_t P::vlsvBufferSize = 0;
 int P::restartStripeFactor = -1;
```
markusbattarbee, line 100: Perhaps still also init these as zero instead of -1?

markusbattarbee:
I approve.

PR #595, opened 2021-12-15
Cudasiator mutator, split plm/ppm/pqm etc back into .cuh files

markusbattarbee:


markusbattarbee:
Also added Realf functions, operation not tested yet, and cleaned up the kernel code by removing a lot of temporary debug comments.

ursg:
Looks good in principle. The cuda versions of most things here are now *really* identical with the cpu-side versions, basically differing only by their ```__device__``` specification, right? So for those alone, I don't actually see why we need a separate cuda implementation at all, and can't just HOSTDEV them.

But things are different for their scalar implementations, so it anyway makes sense to keep those files around.

markusbattarbee:
The cons Vec half(0.5) etc had been moved to inside each function themselves, there was some issue with including them from a header, so that's another change. But I'd hope that the compiler optimizes them away the same way regardless, so no loss there.

One option also would be to template these calls, and to add a Realf version of select(bool check,T l, T r) - would be neater, but left as an exercise to the lead developer if/when this line of development proves fruitful.

ursg:
Right. Anyway, no reason not to merge this.

PR #600, opened 2022-01-07
Ionosphere temperature second fix

markusbattarbee:
Now it actually initializes the ionosphere with the correct temperature (see Urs' comment on the original PR #573)


ursg (approved):
Yup, all good to me.

PR #607, opened 2022-02-10
Vlsv version info

kstppd:
With this PR, version info and the configuration file is now included in the bulk and restart files so we can then recover them later on. The PR goes hand in hand with the analysator PR I also opened that can print the info with `f.print_version().`


markusbattarbee:
Cool beans!
One firm request is to reinstate the compile date into --version output :)
I would also consider perhaps making it clear that the word "version" as well as .vlsv xml output tags with "version" would be what we've considered the version before, and there should be a completely separate string and XML field for "config". Do you think that'd be doable? Cheers!

kstppd:
@markusbattarbee  I think I now get it. So do you mean having essentially one field for version (as it was) and a separate one for the config file. So someone could print version and config separately?

markusbattarbee:
I think the `generate_version.sh` script should place into `version.cpp` the following functions:
`std::string getVersion()`
`std::string getConfig(const char* filename)`

then, in `iowrite.cpp` there should be
`bool writeVersionInfo(vlsv::Writer& vlsvWriter,MPI_Comm comm)` and
`bool writeConfigInfo(std::string version,vlsv::Writer& vlsvWriter,MPI_Comm comm)`

where `writeVersionInfo` writes the *version* information with the XML tag "VERSION" and
`writeConfigInfo` writes the *config file* with the XML tag "CONFIG" and

markusbattarbee:
Plus in the files changed you can see your returning of the date included whitespace changes which will break alignment. It'll work but it won't be as pretty as before :)

kstppd:
Yep this is indeed a nice way of structuring it. Ok I will implement accordingly and also fix the white spaces :D

kstppd:
This should now be inline with Markus' suggestions

markusbattarbee:
Great!
I guess the writer functions could have queried the version.cpp functions directly without passing them along in vlasiator.cpp but either way is good :) 

Assuming you've tested this inside out along with the analysator printouts (please re-verify with the latest changs), I approve!

iowrite.cpp
```diff
@@ -827,6 +827,54 @@ bool writeMeshBoundingBox( Writer & vlsvWriter,
...
+ \param comm MPI comm
+ \return Returns true if operation was successful
+ */
+bool writeConfigInfo(std::string config,vlsv::Writer& vlsvWriter,MPI_Comm comm){
+  
+   int myRank;
+   MPI_Comm_rank(comm, &myRank);
+
+   std::map<std::string, std::string> xmlAttributes;
+   xmlAttributes["name"] ="version" ;
```
markusbattarbee, line 865: This should perhaps be "config_file"?

kstppd: True changed that too

ursg (approved):
Nice. Thumbs up from me.
My only suggestion would have been: Since the config is basically a static string that's known at compile time, it could also be stored as such and all the config handling functions could then simply return references to it.
But realistically: We are dealing with terabytes and terabytes of data in this code. A temporary kilobyte more of config data somewhere isn't even worth the effort.

So let's have it!

ykempf:
Watcha talking about? cfg is not known at compile time??

ursg:
Ah, nonsense indeed. I meant the version string. But that even more underlines the point: let's not bother optimizing more then neccessary and just take this as it is.

generate_version.sh
```diff
@@ -27,6 +27,9 @@ cd $curdir
 cat > version.cpp <<EOF
 #include <iostream>
 #include "mpi.h"
+#include "vector"
```
markusbattarbee, line 30: is vector needed?

markusbattarbee (approved):

ursg:
Looking good. Mergy time!

PR #608, opened 2022-02-11
Fix heap buffer overflow

kstppd:


markusbattarbee:
Won't this order mean it'll always immediately exit with the current (valid) bucket, instead of incrementing?

kstppd:
We increment when we get EMPTYBUCKETS so now I break when we get a non EMPTYBACKET. It is a `!=` now. Pfff my brain stopped working now. Am I wrong?

markusbattarbee:
The purpose of this operator is to move to the next valid bucket.

If we assume that we are already at a valid bucket, it won't increment, and will return the same bucket we are already at. There needs to be at least one forced increment at the start (assuming we aren't already at the end of the bucketlist when it's called)

kstppd:
@markusbattarbee  Indeed you are right. Should we also here guard against being already at the end of the bucketlist?

markusbattarbee:
Apparently it should increment past-the-end, e.g. found this example from `cpu_trans_map_amr.cpp`:
`for (auto it = distances.begin(); it != distances.end(); ++it) {`
and [https://www.cplusplus.com/reference/vector/vector/end/](https://www.cplusplus.com/reference/vector/vector/end/)
tells us: "Returns an iterator referring to the past-the-end element in the [vector](https://www.cplusplus.com/vector) container."

@ursg can confirm?

kstppd:
Yep I was also looking at different stl iterators. Ok I will wait for Urs to answer as well.

kstppd:
Now that I think about again I think that since Urs has defined the iterator's `end` and `begin` we are all good. It can be range accessed fine and the iterations will stop when `it != end` as that should be never dereferenced anyway. @ursg  do you agree?

kstppd:
I also applied the same fix to the const iterator

markusbattarbee:
Ok now at least looks good to me.

open_bucket_hashtable.h
```diff
@@ -186,11 +186,16 @@ template <typename GID, typename LID, int maxBucketOverflow = 4, GID EMPTYBUCKET
...
+         while(index < hashtable->buckets.size()){
+            if (hashtable->buckets[index].first != EMPTYBUCKET){
+               break;
+            }
             index++;
-         } while (hashtable->buckets[index].first == EMPTYBUCKET && index < hashtable->buckets.size());
+         }
          return *this;
       }
+
```
ursg, line 198: Isn't the new code here doing exactly the same as the old one?
Index gets incremented at least once, and then exactly the same conditions break out of the loop.

markusbattarbee: The check against buckets.size() is now separate and before the check against EMPTYBUCKET, preventing out-of-bounds.

ursg:
Ah, of course. Roger that!
Operator precedence and evaluation order in c++ are sometimes a dark art...

open_bucket_hashtable.h
kstppd: Yep as Markus said

PR #611, opened 2022-03-16
fix logical comparisons for iterators of obh

kstppd:
So these would actually crash. Please have a look and let me know what you think.


kstppd:
I actually disagree with myself here. It is legal what’s here and should not crash. I am closing it next week. The only reason I am not closing it now is maybe people prefer this syntax. But no issue after all

PR #612, opened 2022-03-23
Cleaning up obsolete readparameters comment

alhom:


ykempf:
While you're at it, you could change the wording of the abort message as valid default parameters will be OK, it will "only" abort if a parameter passed to `get()` wasn't `add()`ed.

ykempf:
Fine by me! :)

PR #613, opened 2022-04-08
Ionosphere outflow VDFs

ursg:
This code builds the boundary VDFs from three parts:

- The downward travelling half of the VDF is simply whatever happens to flow in from the magnetosphere. For that, we can just run the normal transport algorithm or even just interpolate the moments from the nearest cells (we don't actually so much need to take care that we get this right, since it anyway just gets eaten away by the boundary)
- Part of that downward travelling VDF is outside the loss cone and gets reflected back up. We could just do it exactly like that: Reflect the incoming VDF in the radial direction, for vcells outside the loss cone.
- On top of that there is the ionospheric outflow. This is the part that should get its velocity as ionospheric E×B, and we  model that simply as a maxwellian added on top of the others.

Discussion (and testing) please!


ursg:
Ah dangit. Created this pull request against the wrong branch in the wrong repository. Ignore please!

PR #589, opened 2021-11-08
Fieldsolver dt2, boundary, and moment fixes

markusbattarbee:
Several fixes:

1) accelerates inflowing Maxwellian cells so that (in case of low v-space resolution) they are in sync with accelerated inner cells, removing the inflow artefact which was being built at the boundary.

2) Applies boundary conditions and re-calculates moments at end of initialization. This fixes some corner blips in fields.

3) Corrects the ordering of use of moments and moments_dt2 in RK2 fieldsolver. I think this approach is correct and better matches classical 2nd-order-accurate schemes. This includes changes to ghost updates etc to match the requirements.

4) Perhaps controversial: perB_dt2 (at t=t'+dt/2) is now updated as a dt-length step from the previous round perB_dt2 (at t=t'-dt/2), using E at t=t'. Previously, it was built on top of perB at t=t'. This method is more of a proper 2nd-order-accurate method, but I am not yet completely convinced it doesn't contain the risk of diverging B-grids at the full and intermediate steps.

5) re-activate Hall term derivatives communication for phase RK_ORDER2_STEP2 when EgradPe-term is active. This was overlooked in previous approach.

PR #615, opened 2022-04-14
Fieldsolver boundary, and moment fixes (cherry-pick of 589)

markusbattarbee:
This PR includes the non-controversial parts of #589:

1. accelerates inflowing Maxwellian cells so that (in case of low v-space resolution) they are in sync with accelerated inner cells, removing the inflow artefact which was being built at the boundary.

2. Applies boundary conditions and re-calculates moments at end of initialization. This fixes some corner blips in fields.

3. re-activate Hall term derivatives communication for phase RK_ORDER2_STEP2 when EgradPe-term is active. This was overlooked in previous approach.

The dt2 changes are then pending more testing with interpolated fields, but these should go in soon.

### DIFF
--- a/fieldsolver/ldz_electric_field.cpp
+++ b/fieldsolver/ldz_electric_field.cpp
@@ -1666,16 +1666,20 @@ void calculateUpwindedElectricFieldSimple(
    
    timer=phiprof::initializeTimer("MPI","MPI");
    phiprof::start(timer);
+   // Update ghosts if necessary, unless previous terms have already updated them
    if(P::ohmHallTerm > 0) {
       EHallGrid.updateGhostCells();
    }
    if(P::ohmGradPeTerm > 0) {
       EGradPeGrid.updateGhostCells();
    }
-   if(P::ohmHallTerm == 0 && P::ohmGradPeTerm == 0) {
+   if(P::ohmHallTerm == 0) {
       dPerBGrid.updateGhostCells();
+   }
+   if(P::ohmHallTerm == 0 && P::ohmGradPeTerm == 0) {
       dMomentsGrid.updateGhostCells();
    }
+   
    phiprof::stop(timer);
    
    // Calculate upwinded electric field on inner cells
--- a/fieldsolver/ldz_hall.cpp
+++ b/fieldsolver/ldz_hall.cpp
@@ -804,23 +804,22 @@ void calculateHallTermSimple(
    FsGrid< std::array<Real, fsgrids::bgbfield::N_BGB>, FS_STENCIL_WIDTH> & BgBGrid,
    FsGrid< fsgrids::technical, FS_STENCIL_WIDTH> & technicalGrid,
    SysBoundary& sysBoundaries,
-   cint& RKCase,
-   const bool communicateMomentsDerivatives
+   cint& RKCase
 ) {
    int timer;
    //const std::array<int, 3> gridDims = technicalGrid.getLocalSize();
    const int* gridDims = &technicalGrid.getLocalSize()[0];
    const size_t N_cells = gridDims[0]*gridDims[1]*gridDims[2];
-   
+
    phiprof::start("Calculate Hall term");
    timer=phiprof::initializeTimer("MPI","MPI");
    phiprof::start(timer);
    dPerBGrid.updateGhostCells();
-   if(communicateMomentsDerivatives) {
+   if(P::ohmGradPeTerm == 0) {
       dMomentsGrid.updateGhostCells();
    }
    phiprof::stop(timer);
-   
+
    phiprof::start("Compute cells");
    #pragma omp parallel for collapse(3)
    for (int k=0; k<gridDims[2]; k++) {
@@ -835,6 +834,6 @@ void calculateHallTermSimple(
       }
    }
    phiprof::stop("Compute cells");
-   
+
    phiprof::stop("Calculate Hall term",N_cells,"Spatial Cells");
 }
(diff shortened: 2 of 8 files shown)


markusbattarbee:
Hmm, actually still need to check the logic in case we should set `hallTermCommunicateDerivatives = true;` in other places as well. Do not merge quite yet.

ursg:
Great. Should we still sanity-check these with the testpackage once, or did you do that already?

markusbattarbee:
Yeah I'll be running some testpackage stuff soon.
What I'm scratching my head about now is this comment:
```
         // We need to calculate derivatives of the moments at every substep, but they only
         // need to be communicated in the first one.
         ```
Why? There is only one dMoments grid so if we re-calculate moments but don't ghost communicate, we have incorrect derivatives in the ghost domains. Electric fields do need ghost domain dMoments!

Now when there's lots of FS substepping it might be worth making a dMomentsDt2 grid, to cut down on MPI. And when we move to interpolating moments per-substep, I guess we could decide if we want to interpolate derivatives or differentiate interpolated values. :)

**Comment by markusbattarbee** [comment]
Ok so that was a misformed comment, fixed it now. Running testpackage...

**Comment by markusbattarbee** [comment]
Here is a verification of required ghost updates. They do not follow the rule of "update what you personally need" but that's at least in some parts for the best in order to avoid duplicate ghost updates.

```

propagateMagneticFieldSimple (
	Needs perB, does *not* ghostupdate it. (this has been done since the last perB update by derivatives calculation)
	RK2_STEP2 also needs PerBDt2, does *not* ghostupdate it. (this has been done since the last perB update by derivatives calculation)
	Needs E/Dt2, does *not* ghostupdate it. (this has been done since the last E/Dt2 update by the E-field calculation itself)
	Updates perB/Dt2.
		RK_ORDER1 uses E, updates perB
		RK_ORDER2_STEP1 uses perB, E, updates perBDt2
		RK_ORDER2_STEP2 uses perB, EDt2, updates perB
	After solving simulation cells, ghostupdates the updated version of perB/Dt2, then handles boundaries L1/L2. L2 perB/Dt2 ghostupdate remains undone.)
calculateDerivativesSimple (
	Needs perB/Dt2, ghostupdates it itself. 
	Needs moments/Dt2, ghostupdates it itself.
	Updates dMoments.
	Updates dPerB.)
calculateGradPeTermSimple (
	Needs moments/Dt2, ghostupdate done at derivative calculation. 
	Needs dMoments, ghostupdates it itself.
	Updates EGradPe.)
calculateHallTermSimple (
	Needs perB/Dt2, ghostupdate done at derivative calculation.
	Needs moments/Dt2, ghostupdate done at derivative calculation. 
	Needs dPerBgrid, ghostupdates it itself. 
	Needs dMoments, ghostupdates if EGradPe inactive.. 
	Needs BGB, ghostupdate done at init.
	Updates EHall.) 
calculateUpwindedElectricFieldSimple(
	Needs PerB/Dt2, ghostupdate done at derivative calculation.
	Needs EHall, ghostupdates if active.
	Needs EGradPe, ghostupdates if active.
	Needs moments/Dt2, ghostupadte done at derivative calculation.
	Needs dPerB, ghostupdates if Hall term inactive.
	Needs dMoments, ghostupdates if both Hall term and EgradPe inactive.
	Needs BGB, ghostupdate done at init.
	Updates E/Dt2. (STEP1-> Dt2, ORDER1 / ORDER2_STEP2 -> E)
    --> Performs ghostupdate for the now updated version of E/Dt2

markusbattarbee:
Here is the testpackage log against dev:
[fieldsolver_tplog.txt](https://github.com/fmihpc/vlasiator/files/8518902/fieldsolver_tplog.txt)
So quite large changes in any simulations with a Maxwellian inflow boundary. I guess I should do some plots as well to verify.

markusbattarbee:
So for Magnetosphere_small diff:
`pt.plot.plot_colormap(filename='/wrk-vakka/users/markusb/testpackage/run_2022.04.19_15.21.30/Magnetosphere_small/bulk.0000001.vlsv',draw=1, var='proton/vg_rho')`
image

Where a regular plot is:
image

The inflow acceleration indeed causes diffs.

For Magnetosphere_polar_small:
image
image

For Flowthrough_x_inflow_y_outflow_acc:
image
image

Flowthrough_amr shows large diffs partially because the timestepping has changed very slightly so we're looking at different times.:
---------- tstep = 101 t = 183.132 dt = 1.39163 FS cycles = 1 ----------
---------- tstep = 101 t = 183.182 dt = 1.38898 FS cycles = 1 ----------
image
image

markusbattarbee:
Hmm, perhaps worth looking at For Flowthrough_x_inflow_y_outflow_acc a bit more still:
so diff:
image
dev:
image
PR version:
image

So the y+ boundary (None) remains the same as it should, but everywhere else there's a surprisingly large diff. Hmm. Something due to the moment fixes with a changing diagonal inflow?

markusbattarbee:
Ah, I think it's because both the y- and x- boundaries with inflow are now accelerated, so the value evolves over time. It is 1595 timesteps, so plenty of acceleration rounds, with only 15 blocks per dimension (though it's a resolution of 20 km/s and a temperature of 1e5 K so not very well resolved)

This makes me think, should we make accelerated inflow maxwellians re-scale phase-space to conserve mass? (Activate the flag there, it already exists)

markusbattarbee:
Let's keep it simple - this lack of acceleration at boundaries was a minor issue at best, so I've put it behind a parameter flag.

ursg:
Especially those miniscule fp test diffs and the flowthrough diff really looks negligible now.  Those Magnetosphere diffs still are somewhat macroscopic, but I suppose that's now to be expected.

markusbattarbee:
So any issues with this @ursg @ykempf or can we merge?

PR #616, opened 2022-04-29
Implementation of CUDA streams (one per thread)

markusbattarbee:


markusbattarbee:
Also cleaned away the old kernel, implemented host memory pinning and CUDA memory allocation / de-allocation at load balance with some room to spare.

PR #579, opened 2021-10-04
Make backgroundfield mechanism threadsafe.

ursg:
This is done by replacing the homegrown T3Dfunc - mechanism by modern
C++ std::function and std::bind invocations.

This is extensively tested in the ionosphere implementation, where fieldline tracing inside openMP regions required a threadsafe implementation of the dipole.

PR #617, opened 2022-05-12
Fix vector dipole limit distance checks

ursg:
This was discovered as part of #579

The vector dipole has some wrong (and anyway probably useless) tests for the range limit in there. This commit corrects them, but should be tested whether it actually is still identical.

### DIFF
--- a/backgroundfield/vectordipole.cpp
+++ b/backgroundfield/vectordipole.cpp
@@ -153,7 +153,7 @@ double VectorDipole::operator()( double x, double y, double z, coordinate compon
    IMFdS2cart[1] = 0;     //(r[1]/r1)*dS2dr;
    IMFdS2cart[2] = 0;     //(r[2]/r1)*dS2dr;      
 
-   if((derivative == 0) && (r1 > xlimit[0])) {
+   if((derivative == 0) && (r[0] > xlimit[0])) {
      /* Within transition range (between xlimit[0] and xlimit[1]) we
         multiply the magnetic field with the S2 smootherstep function
         and add an additional corrective term to remove divergence. This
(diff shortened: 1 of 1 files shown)


markusbattarbee:
so actually those checks do nothing. We are already within the region `xlimit[0] < r[0] < xlimit[1]` because regions outside that are dealt with earlier, and `r1 >= r[0]` in all situations.

So we could remove them and just leave the check for whether it should calculate a derivative or not.

ursg:
Automatic testpackage output for pull request #617.

Based on commit ee669249d8ce31fbecb827aa8c494567d41fb462.

Process started by user ```uganse``` on host ```turso02.cs.helsinki.fi``` at Mon, 06 Jun 2022 13:41:03 +0300.

:heavy_check_mark: Compilation succeeded with 202 warnings.

:heavy_check_mark: restart_read: Zero diffs.
:heavy_check_mark: Magnetosphere_small: Zero diffs.
:large_orange_diamond: Magnetosphere_polar_small: Nonzero diffs: `proton/vg_v_nonthermal_0` has absolute error 5.82077e-10, relative error 7.27931e-16.
:large_orange_diamond: Flowthrough_amr: Nonzero diffs: `proton/vg_rho_0` has absolute error 4.94849e-05, relative error 2.29534e-11.

Full testpackage run output is available in `/home/uganse/work/testpackage/testpackage_run_ee669249d8ce31fbecb827aa8c494567d41fb462.log` on host `turso02.cs.helsinki.fi`

markusbattarbee:
strange nonzero there in mp_polar_small but so small that I guess we can just shrug at it.

Since you already merged quite quickly the auto-pr thingy, I'll put the request here: The output should also list the hash/commit we are comparing against. A link would also be cool :)

ursg:
Yup, I agree, I have been seeing random diffs of this magnitude in both Magnetosphere_polar_small and Flowthrough_AMR even in changes that had absolutely *nothing* to do with these two tests, so I suspect they are just particularly sensitive to numerical fluctuations.

That change in the automatic testpackage script is coming up separately then (and I suspect it would also make sense to define an acceptance range for each test).

PR #618, opened 2022-05-25
Amr filtering update

kstppd:
So this is the filtering update, now using the triangle 3D kernel. The default behavior for the filtering configuration is now changed .

- If the user does not select the number of passes then the filtering will be applied automatically (if 3D-AMR).  
**Example**:
```
 max_spatial_level = 2

Fitlering:
        Refinement Level 0==> Passes 2
        Refinement Level 1==> Passes 1
        Refinement Level 2==> Passes 0
```

- Otherwise the user can select the number of passes per refinement level, same as before.
- Also the filtering is reported in the ```logfile``` now instead of  ```stdout```

### DIFF
--- a/fieldsolver/gridGlue.cpp
+++ b/fieldsolver/gridGlue.cpp
@@ -104,9 +104,9 @@ void filterMoments(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
 
 
    // Kernel Characteristics
-   const int stencilWidth = 5;   // dimension size of a 3D kernel
+   const int stencilWidth = 5;   // Stencil width of a 3D 5-tap triangle kernel
    const int kernelOffset = 2;   // offset of 5 pointstencil 3D kernel => (floor(stencilWidth/2);)
-   const Real kernelSum=729.0;
+   const Real kernelSum=729.0;   // the total kernel's sum 
    const static Real kernel[5][5][5] ={
                                  {{ 1,  2,  3,  2,  1},
                                  { 2,  4,  6,  4,  2},
--- a/fieldsolver/gridGlue.cpp
+++ b/fieldsolver/gridGlue.cpp
@@ -186,11 +186,15 @@ void filterMoments(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
                      for (int a=-kernelOffset; a<=kernelOffset; a++){
                         cell = momentsGrid.get(i+a,j+b,k+c);
                         for (int e = 0; e < fsgrids::moments::N_MOMENTS; ++e) {
-                           swap->at(e)+=cell->at(e) *kernel[kernelOffset+a][kernelOffset+b][kernelOffset+c]/kernelSum;
+                           swap->at(e)+=cell->at(e) *kernel[kernelOffset+a][kernelOffset+b][kernelOffset+c];
                         } 
                      }
                   }
                }//inner filtering loop
+               //divide by the total kernel sum
+               for (int e = 0; e < fsgrids::moments::N_MOMENTS; ++e) {
+                  swap->at(e)/=kernelSum;
+               }
             }
          }
       } //spatial loops
--- a/fieldsolver/gridGlue.cpp
+++ b/fieldsolver/gridGlue.cpp
@@ -92,6 +92,121 @@ template <typename T, int stencil> void computeCoupling(dccrg::Dccrg<SpatialCell
   }
 }
 
+/*
+Filter moments after feeding them to FsGrid to alleviate the staircase effect caused in AMR runs.
+This is using a 3D, 5-point stencil triangle kernel.
+*/
+void filterMoments(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
+                           FsGrid< std::array<Real, fsgrids::moments::N_MOMENTS>, FS_STENCIL_WIDTH> & momentsGrid,
+                           FsGrid< fsgrids::technical, FS_STENCIL_WIDTH> & technicalGrid) 
+{
+
+
+
+   // Kernel Characteristics
+   const int stencilWidth = 5;   // dimension size of a 3D kernel
+   const int kernelOffset = 2;   // offset of 5 pointstencil 3D kernel => (floor(stencilWidth/2);)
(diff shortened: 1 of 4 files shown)


fieldsolver/gridGlue.cpp
```diff
@@ -92,6 +92,118 @@ template <typename T, int stencil> void computeCoupling(dccrg::Dccrg<SpatialCell
...
+      // Blurring Pass
+      #pragma omp parallel for collapse(2)
+      for (int k = 0; k < mntDims[2]; k++){
+         for (int j = 0; j < mntDims[1]; j++){
+            for (int i = 0; i < mntDims[0]; i++){
+
+               //  Get refLevel level
+               int refLevel = technicalGrid.get(i, j, k)->refLevel;
+
+               //  Get refLevel level
```
markusbattarbee, line 163: Clean up comments please

fieldsolver/gridGlue.cpp
```diff
@@ -92,6 +92,118 @@ template <typename T, int stencil> void computeCoupling(dccrg::Dccrg<SpatialCell
...
+                  swap->at(e)=0.0;
+               }
+
+               // Perform the blur
+               for (int c=-kernelOffset; c<=kernelOffset; c++){
+                  for (int b=-kernelOffset; b<=kernelOffset; b++){
+                     for (int a=-kernelOffset; a<=kernelOffset; a++){
+                        cell = momentsGrid.get(i+a,j+b,k+c);
+                        for (int e = 0; e < fsgrids::moments::N_MOMENTS; ++e) {
+                           swap->at(e)+=cell->at(e) *kernel[kernelOffset+a][kernelOffset+b][kernelOffset+c]/kernelSum;
```
markusbattarbee, line 190: might be faster to do a division by kernelsum at the end (one division) instead of 27 divisions.

parameters.cpp
```diff
@@ -553,67 +554,49 @@ void Parameters::getParameters() {
...
-         printf("Filtering is on with max number of Passes= \t%d\n", maxNumPassesInt);
-         int lev = 0;
-         for (auto& iter : P::numPasses) {
-            printf("Refinement Level %d-->%d Passes\n", lev, iter);
-            lev++;
+         //sort the filtering passes per refLevel
+         numPasses.clear();
+         std::sort(blurPassString.begin(), blurPassString.end(),
+                  [](const std::string& lhs, const std::string& rhs)->bool 
+                  { return (lhs.size() < rhs.size()) || (lhs.size() == rhs.size() && lhs < rhs);});
```
markusbattarbee, line 573: why are you sorting based on the size of a string? Shouldn't you just convert the string contents to a vector of ints and sort that?

parameters.cpp
```diff
@@ -553,67 +554,49 @@ void Parameters::getParameters() {
...
+         int maxPasses=g_sequence(P::amrMaxSpatialRefLevel-1);
+         for (uint refLevel=0; refLevel<=P::amrMaxSpatialRefLevel; refLevel++){
+            numPasses[refLevel] =maxPasses;
+            maxPasses/=2; 
          }
-
-         printf("Filtering is off and max number of Passes is = \t %d\n",
-                *max_element(P::numPasses.begin(), P::numPasses.end()));
       }
+      P::maxFilteringPasses = numPasses[0];
```
markusbattarbee, line 597: I think it would be better to construct these looping from uint refLevel=P::amrMaxSpatialRefLevel with refLevel-- and doing maxPasses*=2 each loop.

fieldsolver/gridGlue.cpp
```diff
@@ -92,6 +92,118 @@ template <typename T, int stencil> void computeCoupling(dccrg::Dccrg<SpatialCell
...
+      for (int k = 0; k < mntDims[2]; k++){
+         for (int j = 0; j < mntDims[1]; j++){
+            for (int i = 0; i < mntDims[0]; i++){
+
+               //  Get refLevel level
+               int refLevel = technicalGrid.get(i, j, k)->refLevel;
+
+               //  Get refLevel level
+               // Skip pass
+               if (blurPass >= P::numPasses[refLevel] ||
```
markusbattarbee, line 165: You changed P::numPasses into a map instead of a vector - now we're doing a map search for every cell for every refinement pass. I think this is probably not a fast approach and would suggest returning to a vector, unless there's a good reason?

fieldsolver/gridGlue.cpp
kstppd: Indeed. Ok will handle that.

fieldsolver/gridGlue.cpp
kstppd: Roger!

fieldsolver/gridGlue.cpp
```diff
@@ -92,6 +92,118 @@ template <typename T, int stencil> void computeCoupling(dccrg::Dccrg<SpatialCell
...
+                                 { 6, 12, 18, 12,  6},
+                                 { 4,  8, 12,  8,  4},
+                                 { 2,  4,  6,  4,  2}},
+
+                                 {{ 1,  2,  3,  2,  1},
+                                 { 2,  4,  6,  4,  2},
+                                 { 3,  6,  9,  6,  3},
+                                 { 2,  4,  6,  4,  2},
+                                 { 1,  2,  3,  2,  1}}
+                                 };
```
ursg, line 140: I would say this is a textbook example for consteval, if I have ever seen one. :)

However, I appreciate that this code is readable, whereas a consteval function probably wouldn't.

Anyway, my suggestion: make this ```const static```. Not sure if it makes a big difference performance-wise though.

ursg (changes requested):
I mostly concur with Markus'es comments here. No big issues there, just make it a little cleaner and it's good to go in.

fieldsolver/gridGlue.cpp
kstppd: Consteval is a c++20 feauture I think. Will make it a const static then, thanks!

markusbattarbee:
Have you run updated benchmarks with the change to use a vector instead of a map, and the division outside the loop?

kstppd:
@markusbattarbee I have yes and this was a tad faster than the map version. But the tests I run were small. You think I should do a comparison similar to the one in the manuscript ?

ursg (approved):
Yup, this has my thumbs up now, too.

PR #624, opened 2022-07-01
Ionosphere dormand prince

kstppd:


sysboundary/ionosphere.cpp
```diff
@@ -1102,6 +1102,112 @@ namespace SBC {
...
+      Real maxStepsize,
+      TracingFieldFunction& BFieldFunction,
+      bool outwards
+    ){
+
+      Real k1x,k2x,k3x,k4x,k5x,k6x,k7x;
+      Real k1y,k2y,k3y,k4y,k5y,k6y,k7y;
+      Real k1z,k2z,k3z,k4z,k5z,k6z,k7z;
+      
+      assert(r.size()==3 && "Erroneus size for R");
```
ursg, line 1118: Well, r is an ```array<Real, 3>```, so this assert is *always true*

sysboundary/ionosphere.cpp
```diff
@@ -1102,6 +1102,112 @@ namespace SBC {
...
+      std::array<Real, 3>& b,
+      Real& step,
+      Real maxStepsize,
+      TracingFieldFunction& BFieldFunction,
+      bool outwards
+    ){
+
+      Real k1x,k2x,k3x,k4x,k5x,k6x,k7x;
+      Real k1y,k2y,k3y,k4y,k5y,k6y,k7y;
+      Real k1z,k2z,k3z,k4z,k5z,k6z,k7z;
```
ursg, line 1116: These could really be vectors, too!

sysboundary/ionosphere.cpp
```diff
@@ -1102,6 +1102,112 @@ namespace SBC {
...
+      Real zf=r[2] + (35./384.)*k1z + (500./1113.)*k3z + (125./192.)*k4z - (2187./6784.)*k5z +(11./84.)*k6z;
+      
+      Real errx=abs((71./57600.)*k1x -(71./16695.)*k3x + (71./1920.)*k4x -(17253./339200.)*k5x+(22./525.)*k6x -(1./40.)*k7x );
+      Real erry=abs((71./57600.)*k1y -(71./16695.)*k3y + (71./1920.)*k4y -(17253./339200.)*k5y+(22./525.)*k6y -(1./40.)*k7y );
+      Real errz=abs((71./57600.)*k1z -(71./16695.)*k3z + (71./1920.)*k4z -(17253./339200.)*k5z+(22./525.)*k6z -(1./40.)*k7z );
+         
+      Real err=std::max( std::max(errx, erry), errz);
+
+      //Estimate proper stepsize
+      Real s=pow((Ionosphere::eps/(2*err)),1./5.);
```
ursg, line 1195: This pull request would also be an excellent opportunity to rename eps to something more... memorable.

sysboundary/ionosphere.cpp
```diff
@@ -1102,6 +1102,112 @@ namespace SBC {
...
+      step=step*s;
+      if (step>maxStepsize){
+         std::cerr<<"Stepsize in DM method exceeded the max Stepsize. Defaulting to maxStepsize"<<std::endl;
+         step=maxStepsize;
+      }
+      if (err>Ionosphere::eps){
+         dormandPrinceStep(r, b, step, maxStepsize, BFieldFunction, outwards);
```
ursg, line 1202: There seems to be nothing here that prevents infinite recursion of the method. How does this behave in really pathological situations (like the centre of the dipole)? Should there be some check against that?

sysboundary/ionosphere.h
```diff
@@ -145,7 +145,8 @@ namespace SBC {
 
       enum IonosphereCouplingMethod { // Field line integrator for Magnetosphere<->Ionosphere coupling
          Euler, // Euler stepping
-         BS     // Bulirsch-Stoer Stepping
+         BS,     // Bulirsch-Stoer Stepping
+         DM      // Dormand-Prince Stepping
```
ursg, line 149: Shouldn't this be "DP"?

sysboundary/ionosphere.cpp
```diff
@@ -1102,6 +1102,112 @@ namespace SBC {
...
+      }else{
+         //Need to also evaluate b here. But b_unit right?
```
ursg, line 1204: This comment should probably say "Evaluate B at the final stepping point to get the v vector of the fieldline".

ursg (changes requested):
This whole looks unnecessarily scalar to me. Wouldn't this a very nice place to use vectors basically everywhere?

sysboundary/ionosphere.cpp
kstppd: Indeed. I actually lifted this off of my mini app which uses vectors so these stayed there. Doing all those changes in a minute

sysboundary/ionosphere.cpp
kstppd: max_allowed_error ?

sysboundary/ionosphere.cpp
kstppd: I will have too look into StackOverflow of how to avoid a stack overflow :D

sysboundary/ionosphere.cpp
```diff
@@ -1103,7 +1105,159 @@ namespace SBC {
...
+      //Error calculation
+      std::array<Real,3>rf;
+      std::array<Real,3>error_xyz;
+      rf[0]=r[0] +(35./384.)*kx[0] + (500./1113.)*kx[2] + (125./192.)*kx[3] - (2187./6784.)*kx[4] +(11./84.)*kx[5];
+      rf[1]=r[1] +(35./384.)*ky[0] + (500./1113.)*ky[2] + (125./192.)*ky[3] - (2187./6784.)*ky[4] +(11./84.)*ky[5];
+      rf[2]=r[2] +(35./384.)*kz[0] + (500./1113.)*kz[2] + (125./192.)*kz[3] - (2187./6784.)*kz[4] +(11./84.)*kz[5];
+      
+      error_xyz[0]=abs((71./57600.)*kx[0] -(71./16695.)*kx[2] + (71./1920.)*kx[3] -(17253./339200.)*kx[4]+(22./525.)*kx[5] -(1./40.)*kx[6] );
+      error_xyz[1]=abs((71./57600.)*ky[0] -(71./16695.)*ky[2] + (71./1920.)*ky[3] -(17253./339200.)*ky[4]+(22./525.)*ky[5] -(1./40.)*ky[6] );
+      error_xyz[2]=abs((71./57600.)*kz[0] -(71./16695.)*kz[2] + (71./1920.)*kz[3] -(17253./339200.)*kz[4]+(22./525.)*kz[5] -(1./40.)*kz[6] );
```
ursg, line 1191: These *still* look like proper vectorization (using agner's vectorclass) would benefit them. Or are you sure that the compiler actually manags to autovectorize those?

sysboundary/ionosphere.cpp
```diff
@@ -1207,13 +1361,33 @@ namespace SBC {
...
+      uint32_t attempts=0;
       switch(method) {
          case Euler:
             eulerStep(x, v,stepsize, BFieldFunction, outwards);
             break;
+         case ADPT_Euler:
+            do{
+               reTrace=!adaptiveEulerStep(x, v, stepsize, maxStepsize, BFieldFunction, outwards);
+               attempts+=1;
+            }while(reTrace && attempts<= Ionosphere::max_dormand_prince_attempts);
```
ursg, line 1374: This is a bit dirty: The adaptive euler now uses max_dormand_prince_attempts. Should that variable rather be "max_tracer_attempts"?

sysboundary/ionosphere.cpp
```diff
@@ -1207,13 +1361,33 @@ namespace SBC {
...
+            if (reTrace){
+               std::cerr<<"Adaptive Euler field line tracer exhausted all available attempts and still did not converge..."<<std::endl;
```
ursg, line 1376: What would we want the code to do in this situation?
Abort? (Because otherwise we have a physically nonsensical mapping)
Just continue?
Keep comlaining every timestep? (I guess that's what it would do right now)

ursg (commented):
As of now, #625 and this one are so closely entangled that I'm wondering if it wouldn't be easier to turn them into a single, combined PR.

ykempf:
I merged into my branch so indeed we can continue on #625...

sysboundary/ionosphere.cpp
kstppd: Have not checked but I can use vectorclass here indeed.

sysboundary/ionosphere.cpp
kstppd: Ah right. I wanted to change that to tracer_max_attempts but forgot to. Will do

sysboundary/ionosphere.cpp
kstppd: Yeah I am not sure and maybe we need to think about it. I would lean towards aborting because we would not want to proceed in the simulation with bad mapping. What do you say?

PR #627, opened 2022-07-11
Particle pusher update

lkotipal:
Add fsgrid support to the particle pusher, along with feature updates by @markusbattarbee.

Reading fields now uses field names specified in `particles.cfg`. Due to this change, fluxfunction must also be run with the command line parameter `--run_config=particles.cfg`.


ursg:
I just wanted to check if this can be merged to dev (since there's no point just leaving it dangling here). After fixing the straightforward conflict, I still get a compile error:
```
In file included from particles/readfields.cpp:27:
particles/readfields.h: In function ‘void readfields(const char*, Field&, Field&, Field&, Field&, bool, bool)’:
particles/readfields.h:357:25: error: could not convert ‘onFsGrid’ from ‘std::vector<bool>’ to ‘bool’
  357 |          double shift = onFsGrid ? 0 : fields[idx]->dx[i] / 2;
      |                         ^~~~~~~~
      |                         |
      |                         std::vector<bool>
particles/readfields.h: In instantiation of ‘void readfields(const char*, Field&, Field&, Field&, Field&, bool, bool) [with Reader = vlsvinterface::Reader]’:
particles/readfields.h:440:63:   required from here
```

PR #631, opened 2022-07-31
CUDA hashmap

kstppd:
This is a first attempt to port over the cuda hashmap and the cuda vector in a vlasiator branch.


markusbattarbee:
Let's keep the hashmap and splitvector in their separate repository for now

PR #633, opened 2022-08-08
Cast CELLID to int before printing

henry2004y:
Internally vspace CELLID is stored as float32. We can improve the error message by casting it to int32 before printing.


markusbattarbee (approved):

ursg:
Yup, makes sense. Let's merge it right away.

PR #634, opened 2022-08-09
Add file triggers to run stuff on compute nodes on demand.

ursg:
On platforms where there's no direct ssh access to the compute nodes, this can sometimes be helpful to run some auxillary commands during runtime.

CMD0, to run a command on rank 0 only, is already here. CMDALL coming up next.

### DIFF
--- a/ioread.cpp
+++ b/ioread.cpp
@@ -52,6 +52,7 @@ typedef Parameters P;
  * If a file KILL was written and is readable, then a bailout without a restart is initiated.
  * If a file SAVE was written and is readable, then restart writing without a bailout is initiated.
  * If a file DOLB was written and is readable, then a new load balancing is initiated.
+ * If a file CMD0 was written and is readable, run its contents with system() on rank 0
  * To avoid bailing out upfront on a new run the files are renamed with the date to keep a trace.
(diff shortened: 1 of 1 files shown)


ursg:
And now there's CMDALL as well.
I am not 100% certain this is free of race conditions though:
What if the CMDALL file gets created in *the very moment* where the task checks for it, and some ranks already have it, whereas others don't? The MPI_Barrier() would then block forever, I suppose.

markusbattarbee:
Indeed it would. How about adding a flag which is lifted by the master rank (a bit like for the DOLB, SAVE, KILL, STOP) and then all processes acting on that flag? That should be safer.

ursg:
Yep, makes sense afterall. So no shortcutting that. Ok, coming up next.

ykempf:
So what kinds of auxiliary commands to you expect to be able to run with this? The couple of ideas I got wouldn't work or we could get out with existing mechanisms, but I guess you had some use case there. And are we infringing usage conditions/introducing features that admins of multi-user-per-node systems won't like?

ursg:
Since we recently discovered how our favourite supercomputers' queueing systems allow us to start up additional commands on the compute nodes anyway, I suppose this pull request is no longer really needed.
Closing it. If we ever feel the need for it again, we might reopen it.

PR #635, opened 2022-08-10
Ignore txt files

henry2004y:
All text files (e.g. `log.txt`, `diagnostic.txt`, `phiprof*.txt`) shall be safely ignored by git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 *.dat
-*.txt
+logfile.txt
+phiprof*.txt
+diagnostic.txt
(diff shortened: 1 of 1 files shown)


ursg:
Hmm, there are a number of legitimate .txt files in the source tree as well, in particular under doc/
Of course gitignore doesn't affect already-checked-in files, but I'd advocate making this more specific, and blocking logfile.txt, phiprof*.txt and diagnostic.txt.

ursg:
Yeps, great!

PR #638, opened 2022-08-10
warning on master only

henry2004y:
Show the noncubic spatial cell warning only once.

### DIFF
--- a/vlasiator.cpp
+++ b/vlasiator.cpp
@@ -450,10 +450,14 @@ int main(int argn,char* args[]) {
       = {P::xmin, P::ymin, P::zmin};
 
    // Checking that spatial cells are cubic, otherwise field solver is incorrect (cf. derivatives in E, Hall term)
-   if((abs((technicalGrid.DX-technicalGrid.DY)/technicalGrid.DX) > 0.001) ||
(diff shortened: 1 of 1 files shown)


markusbattarbee:
I dunno, this is such an important warning that in my opinion it's fine for it to be repeated over and over! :)

henry2004y:
If we know that for 3D noncubic cell computations are definitely wrong, then we should probably throw an error instead of a warning here.

ursg:
I don't know if we have tested that at all in the last couple of years.
In 2.5D, they are sorta-kinda-correct-ish, in that they still mostly behave as if they were cubic (some strategic places in the code simply assume that DX=DY=DZ and only use DX in their calculations). In 3D this might lead to nonconservation of divB or something like that.
I would still keep this as a (non-crashing) warning though. Because who knows, maybe somebody does something smart that only looks stupid with it?

Despite that: thumb up for the actual PR content here.

ykempf:
Just for people's information: I *think* that the field solver is OK as we have pulled through DX/DY/DZ and not assumptions of them being equal in handling derivatives, now is the upwinding sane with a non-cubic grid? I'm not sure. And for the semi-Lagrangian Vlasov solver I suppose it shouldn't have issues either? Checking that everything is fine, comprehensively, sounds like a daunting task to me, and I am not so sure we'd gain a lot from allowing non-cubic cells anyway.

PR #640, opened 2022-08-22
Project init

henry2004y:
This PR aims at clarifying the initialization step for projects. `initialize()` is required for projects because extra parameters may be read for a specific project which are then used to set the internal parameters; otherwise all the initialization should be done in the class constructors. In order to reflect the sequences, `initialize()` is moved after `addParameters()` and `setParameters()`. `initialize()` is enforced for the inherited classes by setting `virtual void initialize() = 0` in the base class. Similarly, `setProjectBField()` is enforced for the inherited classes and renamed to `setInitialBField` to reflect its purpose.

### DIFF
--- a/projects/project.cpp
+++ b/projects/project.cpp
@@ -545,85 +545,61 @@ namespace projects {
    
 Project* createProject() {
    Project* rvalue = NULL;
-   if(Parameters::projectName == "") {
-      cerr << "No project specified! Please set 'project' parameter!" << endl;
-      abort();
-   }
-   if(Parameters::projectName == "Alfven") {
+
+   if (Parameters::projectName == "Alfven") {
       rvalue = new projects::Alfven;
-   }
-   if(Parameters::projectName == "Diffusion") {
+   } else if (Parameters::projectName == "Diffusion") {
       rvalue = new projects::Diffusion;
-   }
-   if(Parameters::projectName == "Dispersion") {
+   } else if (Parameters::projectName == "Dispersion") {
       rvalue = new projects::Dispersion;
-   }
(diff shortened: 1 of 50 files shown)


projects/project.cpp
```diff
@@ -600,85 +545,61 @@ namespace projects {
...
+   } else if (Parameters::projectName == "") {
+      cerr << "No project specified! Please set 'project' parameter!" << endl;
+      MPI_Abort(MPI_COMM_WORLD, 1);
+   } else {
       cerr << "Unknown project name!" << endl;
-      abort();
+      MPI_Abort(MPI_COMM_WORLD, 1);
    }
 
-   getObjectWrapper().project = rvalue;
```
ykempf, line 681: Why did you remove this? Is it unused for sure?

ykempf:
>  In order to reflect the sequences, `initialize()` is moved after `addParameters()` and `setParameters()`.

I don't see where the order changed, can you point that out to me? Maybe I'm just not fit for this today... ;)

projects/project.cpp
henry2004y: It is set twice, once here and once in Vlasiator.cpp

henry2004y:
> > In order to reflect the sequences, `initialize()` is moved after `addParameters()` and `setParameters()`.
> 
> I don't see where the order changed, can you point that out to me? Maybe I'm just not fit for this today... ;)

Ah, I mean the order of definition, not the order of execution. Sorry about the confusion.

projects/project.cpp
ykempf: All right.

Issue #642, opened 2022-08-22
`Project::refineSpatialCells` does refinement?


henry2004y:
In the base class of Project, there is an implementation of `refineSpatialCells()` up to 2 levels. In the method comment, https://github.com/fmihpc/vlasiator/blob/b26df8f3e75f6d81c23e531849b4d16938794869/projects/project.cpp#L516

it says that this function does nothing and most likely should not be called. Do we need to keep the base method? Based on my quick scan the testAmr project https://github.com/fmihpc/vlasiator/blob/b26df8f3e75f6d81c23e531849b4d16938794869/projects/testAmr/testAmr.cpp#L277

basically has the same implementation.

ykempf:
Yes we do as we use it e.g. in Flowthrough_amr for the testpackage.

ykempf:
Well actually the comment is wrong, it doesn't do nothing, it refines for a certain width around a certain center: https://github.com/fmihpc/vlasiator/blob/b26df8f3e75f6d81c23e531849b4d16938794869/projects/project.cpp#L536.

PR #644, opened 2022-08-23
Fix comments in project base method refineSpatialCells

henry2004y:
Correct comments and format the function. See #642.

### DIFF
--- a/projects/project.cpp
+++ b/projects/project.cpp
@@ -512,92 +512,92 @@ namespace projects {
    }
 
    /*
-     Refine cells of mpiGrid. Each project that wants refinement shoudl implement this function. 
-     Base class function prints a warning and does nothing.
+     Refine cells of mpiGrid up to 2 levels in a box region. This can be overrided in each child project.
     */
-   bool Project::refineSpatialCells( dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid ) const {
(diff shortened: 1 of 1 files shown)


projects/project.cpp
```diff
@@ -512,92 +512,92 @@ namespace projects {
...
-               
-               std::array<double,3> xyz;
+               std::array<double, 3> xyz;
                xyz[0] = P::amrBoxCenterX + (0.5 + i - P::amrBoxHalfWidthX) * P::dx_ini;
                xyz[1] = P::amrBoxCenterY + (0.5 + j - P::amrBoxHalfWidthY) * P::dy_ini;
                xyz[2] = P::amrBoxCenterZ + (0.5 + k - P::amrBoxHalfWidthZ) * P::dz_ini;
-               
+
                CellID myCell = mpiGrid.get_existing_cell(xyz);
+#ifndef NDEBUG
```
markusbattarbee, line 541: Umm, this is now incorrect - we do want it to refine even if debug mode isn't on! Same for L575.

henry2004y: Yes right. I misunderstood the method `refine_completely_at()` in the if statement.

ursg:
This is basically only cosmetic changes to the code now, right? The actual functionality hasn't been touched.
So, sure, let's merge it!
(Except github tells me it has a merge conflict in project.cpp)

Issue #645, opened 2022-08-23
`ObjectWrapper` calling method


henry2004y:
The struct `ObjectWrapper` is used for passing global variables, and a method `getObjectWrapper()` is defined to pass the reference of the object `objectWrapper` declared in Vlasiator.cpp to get/set the global variables. An alternative way, which is maybe more straightforward, is to make the object `extern` in the header file and call `objectWrapper` directly.

A side note is that the struct name is maybe a bit too abstract for a global container.

PR #646, opened 2022-08-23
Call objectWrapper directly

henry2004y:
See #645

### DIFF
--- a/amr_refinement_criteria.cpp
+++ b/amr_refinement_criteria.cpp
@@ -55,7 +55,7 @@ namespace amr_ref_criteria {
          Realf f_cen = array[vblock::padIndex<PAD>(ic+1,jc+1,kc+1)];
          
          #warning In here should we use SpatialCell::getVeloctyBlockMinValue()?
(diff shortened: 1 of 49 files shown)


markusbattarbee:
I'm afraid I don't really see any practical benefit in these changes. And what it will do is lead to lots of merge conflicts in other development branches, which would be forced to implement this same change.

The only way to prevent merge conflicts would be to still also keep the getObjectWrapper() method, but that would then lead to there being two methods for achieving the same thing instead of one, and that's not something we want to encourage either.

ursg:
It does make the code a tiny little bit prettier, but it is quite a hefty changeset for that, as Markus said.

Also, I believe there is a reason why the objectWrapper access was hidden behind this getter function. At the moment (and for the forseeable future), this is a pure singleton object. So the function call really does nothing else than the code you have now replaced it by.

However: There have been considerations in the past that the Vlasiator C++ codebase could be further modularized and potentially get a python interface to actually build and run it. In that case, there could concievably be multiple simulation instances existing in parallel, and each would carry its own object wrapper. Granted, this is not something that's realistically happening any time soon, but I think it's quite ok to just leave the global function call as it is for now.

PR #648, opened 2022-08-29
Add time-varying BC

henry2004y:
This is an updated PR for adding the time-varying BC to the dev branch on 29/08/2022. I simply copy & paste the summary from the previous PR #570. This has not been tested with the test package suite, and maybe @ursg can trigger an automated test on Vorna? The new test for the time-varying BC can also be added to the list. With respect to the changes of variable and function names, feel free to modify them if they are not appropriate, as all maintainers are allowed for editing.

Major changes:
1. The source files under `sysboundary` is fully reorganized. `setmaxwellian.cpp` --> `maxwellian.cpp`, `setbyuser.cpp` --> `user.cpp`, and a new file `inflow.cpp` is introduced.
2. The logic behind inflow boundary condition is reimplemented. `Maxwellian` is now a subclass of `Inflow`, as compared to the previous logic where `SetMaxwellian` is a subclass of `SetByUser`. This leaves space for potentially other kind of inflow BCs that are not necessarily Maxwellian. The new User class is an empty class for customized boundary condition for future users.
3. New implementation of boundary file reader in `inflow.cpp`, following C++ standard whereas the previous reader follows C. Additional functionalities include
  * Comments are allowed in lines started with `#` (anywhere in the file):
```
# time density temperature Vx Vy Vz Bx By Bz
0.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0	0.0
```
  * Automatic error checking for missing inputs or extra inputs
```
# time density temperature Vx Vy Vz Bx By Bz
0.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0	0.0
2.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0
```
`Error: Missing input values at line 2`
```
# time density temperature Vx Vy Vz Bx By Bz
0.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0	0.0
2.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0   0.0 1.0
```
`Error: Extra input values at line  2`
```
2.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0	0.0
1.0	1.0e6	500000.0	-750000.0	0.0	0.0	0.0	0.0   0.0
```
`Parameter data must be in ascending order!`
  * Arbitrary number of lines of data
4. Most SBC class methods have now return type `void` instead of `bool`, as the execution will stop as long as any error is encountered.
5. Add a new 1D flowthrough test with varying density at the left boundary.
6. Coding style follows the new Vlasiator format guidelines.

---

One thing to note is that currently if `dynamic==false` and there are multiple time stamp inputs, it seems that only the **first line** (rather than the last line) will be read. I don't know if we want to change this behaviour, or leave as it is to remind the user that this is probably not you want to have multiple inputs in a constant inflow run.


henry2004y:
The B field values in the ghost cells shown in the output VLSV files are 0s, while internally they seem to have the correct template cell values. I need to figure out how to set the field values in the ghost cell outputs right.

sysboundary/inflow.cpp
```diff
@@ -0,0 +1,497 @@
...
+      creal y = cell->parameters[CellParams::YCRD] + 0.5 * dy;
+      creal z = cell->parameters[CellParams::ZCRD] + 0.5 * dz;
+
+      bool isThisCellOnAFace[6];
+      determineFace(&isThisCellOnAFace[0], x, y, z, dx, dy, dz, true);
+
+      for (uint i = 0; i < 6; i++) {
+         if (facesToProcess[i] && isThisCellOnAFace[i]) {
+            copyCellData(&templateCells[i], cell, false, popID, true); // copy also vdf, _V
+            copyCellData(&templateCells[i], cell, true, popID, false); // don't copy vdf again but copy _R now
```
henry2004y, line 275: Maybe I made unnecessary copies here. `copyCellData(&templateCells[i], cell, false, popID, doCalcMomentsV)` shall be enough. But even then, I copy VDFs twice, once post translation and once post acceleration. Should it be done only once, since the Vlasov BC is always applied at `P::t+0.5*P::dt`? @ykempf

sysboundary/maxwellian.cpp
```diff
@@ -0,0 +1,303 @@
...
+                                                  vyCell + 0.5 * dvyCell - Vy, vzCell + 0.5 * dvzCell - Vz);
+                  }
+
+                  if (avr != 0.0) {
+                     data[blockLID * WID3 + cellIndex(ic, jc, kc)] = avr;
+                  }
+               } // for-loop over cells in velocity block
+      }          // for-loop over velocity blocks
+
+      templateCell.get_population(popID).RHO_R = templateCell.get_population(popID).RHO;
```
henry2004y, line 263: Maybe this should be done after `calculateCellMoments`? Otherwise it would get the RHO value from last templateCell?

PR #649, opened 2022-09-05
Datareducer for calculating the heat flux vector

vetarvus:
Implemented the datareducer based on the existing PTensorDiagonal code.


ursg:
I'm all good with merging this. Seems to work, looks ok, doesn't break anything, what more is there to want!

PR #656, opened 2022-10-18
Initialise variable used to check whether load balances or restarts a…

ykempf:
…re done.


ykempf:
This points to master now, can be changed to dev if you prefer.

ursg:
I would suggest pointing this to dev. Putting it into master won't really make a difference, because we're anyway going to merge dev before the next release.

PR #657, opened 2022-10-25
Fix global buffer overflow when restarting from with MPI tasks

kstppd:
I got this when playing around with sanitizers. It would produce a global buffer overflow.


ursg:
Wow, ok. This is clearly overwriting memory when Real is double. As far as I can tell, the next thing in memory after that is the transShortPencils value, so did that usually get garbled?

Also confusing that the same parameter gets assigned a floating point value (1.0) just two lines above the offending one.

Anyway, good catch. Let's merge it!

PR #658, opened 2022-10-28
Fix non existent spatial cells in shrink to fit

ursg:
This is the same as https://github.com/ursg/vlasiator/pull/58/files fixed and pointed onto dev.


markusbattarbee:
I would recommend adding the cellparams match check to this as well, to safeguard against surprising DCCRG neighbourhood behaviour.

ykempf:
Oh boy. Really? :facepalm:

ursg:
Obviated by #781

PR #625, opened 2022-07-01
Field tracing

ykempf:
For discussion and cherry-picking for others at the moment.

One tangent question is why my attempt at making an int DRO for ionosphere failed miserably? Did I oversee something @ursg?

PR #659, opened 2022-11-01
Simplified scaling of FACs from the coupling radius down to the ionosphere using the ratio of magnetic field strengths

ykempf:
This is the commit I've been talking about.

### DIFF
--- a/MAKE/Makefile.mahti_gcc
+++ b/MAKE/Makefile.mahti_gcc
@@ -57,7 +57,8 @@ testpackage: CXXFLAGS += -DUSE_JEMALLOC -DJEMALLOC_NO_DEMANGLE
 
 MPT_VERSION = 4.0.3
 JEMALLOC_VERSION = 5.2.1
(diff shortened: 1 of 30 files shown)


ykempf:
Question @ursg: at https://github.com/ykempf/vlasiator/blob/ionosphere_FAC_scaling/sysboundary/ionosphere.cpp#L1786-L1790 `upmappedArea` is computed as normal to the flux tube. This variable is dropped from the FAC calculation, only some Poynting calculation needs it at https://github.com/ykempf/vlasiator/blob/ionosphere_FAC_scaling/sysboundary/ionosphere.cpp#L1874. I had modified the upmappedArea to be normal to the sphere, but that got dropped altogether with the new scaling. Let me know which is the correct one to have for the Poynting calculation, I can commit it if there it should be normal to the sphere or rather just the area of the triangle and not projected on the flux tube.

ykempf:
Right this needs features out of #625 and doesn't compile as such, fixes TBD ASAP (I can port some of the fixes for this PR but need to scoot now).

PR #660, opened 2022-11-01
Ionosphere fac scaling

ykempf:
So fixing the git wreck of the previous PR...

I hand-picked what was needed from #625.

However note that I am not at all convinced by the Poynting flux calculation and I "broke" the current version by modifying upmappedArea to be the triangle area and not that dotted with the local B at upmapped coordinates. If we want that back, I need to change back some of the reordering of the loops and cell validity checks.


ykempf:
Nooo I should have stopped, another wreck. This I just pushed to the wrong branch...

ykempf:
So if we agree on this one I suppose it could be merged next.

ykempf:
Right, now at least I agree. Anyone else?

PR #661, opened 2022-11-02
Define -DIONOSPHERE_SORTED_SUMS flag to solve the ionosphere with sorted threaded sums.

ykempf:
This is added to all testpackage flag sets in the Makefiles.

I was able to compile both without and with the flag but it is not run-tested.

@alhom it would be great as part of the setup performance testing if you could run a pair without and with this flag (maybe both testpackage actually?) and some ionosphere solving. The goal is to get a massively better performance of the ionosphere solver with "palatable" diffs (e.g. potential values reported in the logfile will differ at the 1e-5 or 1e-6 level unless I botched something).


ykempf:
Oh yes I introduced race conditions on the shared sums, e.g. bknum gets +='ed by all threads, I need to rewrite differently and use `#pragma omp for reduction(+:bknum)` for example. Bl**dy threading.

ykempf:
@alhom can you give this version a spin?

ykempf:
Gosh this stack of pragmas is ugly, but it's easier this way in terms of running e.g. testpackage than having that as a cfg option. In the latter case we'd always compile both versions but we'd need to duplicate the solver code I think or it'd get even less readable.

If you want me to document this stuff, tell me what you'd want mentioned and where.

ursg (approved):

PR #662, opened 2022-11-02
Fix conductingsphere temperature value in Magnetosphere_polar_small test

ursg:
Also, make the warning for a zero-valued ionosphere temperature a little more up-to-date w.r.t. parameter names.


testpackage/tests/Magnetosphere_polar_small/Magnetosphere_polar_small.cfg
```diff
@@ -99,6 +99,7 @@ precedence = 2
 
 [proton_conductingsphere]
 rho = 1.0e6
+T=100000.0
```
markusbattarbee, line 102: this should be 5e5, according to line 137-138

ursg: Oh, true. My bad.

PR #663, opened 2022-11-03
Warnings reduction after dAMR merge.

ursg:
The dynamic AMR merge (PR #497) introduced a couple of new compiler warnings, mostly signed/unsigned comparisons.

This PR cleans them up again.

### DIFF
--- a/common.h
+++ b/common.h
@@ -378,7 +378,7 @@ namespace fsgrids {
    };
    
    struct technical {
-      int sysBoundaryFlag;  /*!< System boundary flags. */
+      uint sysBoundaryFlag;  /*!< System boundary flags. */
       int sysBoundaryLayer; /*!< System boundary layer index. */
       Real maxFsDt;         /*!< maximum timestep allowed in ordinary space by fieldsolver for this cell**/
       int fsGridRank;       /*!< Rank in the fsGrids cartesian coordinator */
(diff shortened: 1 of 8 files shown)


common.h
```diff
@@ -378,7 +378,7 @@ namespace fsgrids {
    };
    
    struct technical {
-      int sysBoundaryFlag;  /*!< System boundary flags. */
+      uint sysBoundaryFlag;  /*!< System boundary flags. */
```
ursg, line 381: This is the only part of this PR that I don't consider completely trivial. But since sysBoundaryFlag anyway can only take enum values, there is no reason to assume it would ever be negative.

sysboundary/setbyuser.cpp
```diff
@@ -288,7 +288,7 @@ namespace SBC {
          std::array<bool, 6> isThisCellOnAFace;
          determineFace(isThisCellOnAFace, mpiGrid, id);
 
-         int max = 6;
+         uint max = 6;
```
markusbattarbee, line 291: wouldn't this be better done with just a loop from 0 to <6, and a break?

ursg: True. Good sugestion. Done.

sysboundary/setbyuser.cpp
```diff
@@ -288,12 +288,11 @@ namespace SBC {
          std::array<bool, 6> isThisCellOnAFace;
          determineFace(isThisCellOnAFace, mpiGrid, id);
 
-         int max = 6;
-         for (uint i=0; i < max; i++) {
+         for (uint i=6; i < max; i++) {
```
markusbattarbee, line 291: `         for (uint i=0; i < 6; i++) {`

ursg: Goddamnit, what was I thinking there? Or why was I not?
Anyway, fixed.

PR #664, opened 2022-11-04
Forcing bulk velocity in cells next to the ionosphere boundary.

ursg:
The boundary cells set a new cell parameter, which is read in the acceleration update to shift bulkv with an additional transformation step.

This is highly preliminary and untested. It compiles though!


vlasovsolver/cpu_acc_transform.cpp
```diff
@@ -123,6 +123,26 @@ Eigen::Transform<Real,3,Eigen::Affine> compute_acceleration_transformation(
...
 
+   // If a bulk velocity is being forced here, perform that first, and let everything below continue in the new frame.
+   if(spatial_cell->parameters[CellParams::FORCING_CELL_NUM] > 0) {
+      Eigen::Matrix<Real,3,1> forced_bulkv(spatial_cell->parameters[CellParams::BULKV_FORCING_X],
+                                           spatial_cell->parameters[CellParams::BULKV_FORCING_Y],
+                                           spatial_cell->parameters[CellParams::BULKV_FORCING_Z]);
+      forced_bulkv /= spatial_cell->parameters[CellParams::FORCING_CELL_NUM];
+
+      Eigen::Matrix<Real,3,1> bulkDeltaV = forced_bulkv - bulk_velocity;
+      total_transform=Translation<Real,3>(bulkDeltaV) * total_transform;
```
markusbattarbee, line 134: I think the actual rotation has to take place in the Hall frame still, so the forcing should be added after everything else to just nudge the VDF into the correct bulk velocity. Otherwise looks preliminarily sane, as sane as this forcing can be!

markusbattarbee:
One question still is:
 The vDrift from the ionospheric solver is calculated purely from the ExB drift. Well, we don't have reliable B derivatives to be able calculate the Hall term there.

However, in L2 non-sysB cells, we can. And we also have EgradPe. Should those come into play?

vlasovsolver/cpu_acc_transform.cpp
ursg: Ah, you mean that otherwise we'll get the whole distribution agyrotropic and rotate it by a bit of a gyration fraction, until we drag it back to the forced bulkv frame in the next timestep? Hmm, I guess I imagined that upside down. In that case: yep, let's do the forcing after rotation.

sysboundary/ionosphere.cpp
```diff
@@ -3768,6 +3770,18 @@ namespace SBC {
...
          Real density = 0;
          switch(boundaryVDFmode) {
+            case ForceL2EXB:
+               {
+               // EXB forcing is assigned to the L2 Neighbour cells here, so they can update their VDFs in acceleration
+               const vector<CellID>& closestCells = getAllClosestNonsysboundaryCells(cellID);
+               for (CellID celli : closestCells) {
+                  mpiGrid[celli]->parameters[CellParams::BULKV_FORCING_X] += vDrift[0];
+                  mpiGrid[celli]->parameters[CellParams::BULKV_FORCING_Y] += vDrift[1];
+                  mpiGrid[celli]->parameters[CellParams::BULKV_FORCING_Z] += vDrift[2];
```
ykempf, line 3780: I disagree with the +=, it should be = or we'll just keep incrementing.

markusbattarbee: it gets reset during acc

sysboundary/ionosphere.cpp
```diff
@@ -3768,6 +3770,18 @@ namespace SBC {
...
+                  mpiGrid[celli]->parameters[CellParams::FORCING_CELL_NUM]+=1;
+               }
+               // Fall through, to handle L1 in the same way as fixed moments
```
ykempf, line 3783: This is never called for NOT_SYSBOUNDARY L2 cells, we need to add that call somewhere.

sysboundary/ionosphere.cpp
ykempf: duh thanks :eyeglasses:

ykempf:
Damn, did I now wreck things by also pushing some of the other branches I megred in? Sorry!

sysboundary/sysboundarycondition.h
```diff
@@ -143,6 +143,10 @@ namespace SBC {
             cuint& component
          );
         
+         virtual void mapCellPotentialAndGetEXBDrift(
+            std::array<Real, CellParams::N_SPATIAL_CELL_PARAMS>& cellParams
+         )=0;
+
```
ursg, line 149: Hmm, I'm not particularly fond of this change, since it now requires every sysboundarycondition to implement a dummy version of this same function.

And indeed, you implemented it as an error message in all of them except ionosphere, which in turn warns if they have even been called. At the same time, as part of the dAMR merge, the sysboundary class hierarchy has separated out OuterBoundaryCondition, so why not the following:
- Make all the outer boundary conditions inherit from OuterBoundaryCondition, if that isn't already the case
- Implement this "warning, I have been called!" implementation as part of OuterBoundaryCondition
- Let only ionosphere and conductingsphere even hassle with it (the latter returning a simple zero everywhere)

ursg (commented):
Not fully agreeing to the changes you did to the boundaryconditions there. Functionally ok otherwise.

sysboundary/sysboundarycondition.h
ykempf: Except I need one in the parent class as I am calling it with
```
getObjectWrapper().sysBoundaryContainer.getSysBoundary(sysboundarytype::IONOSPHERE)->mapCellPotentialAndGetEXBDrift(spatial_cell->parameters);
```
in `vlasovsolver/cpu_acc_transform.cpp` so I don't think your suggestion makes much of a difference?

ursg: > Formal nitpicking: when I was thrown into this object-oriented system boundary business I was told to have one class per file or pair of source+header file. But thanks for the heads-up, this indeed would have saved me a bit of pasting to some of the files. :)

Indeed, you are right. I think it would be good style to move the outerBoundaryCondition into a file of its own. But that shouldn't happen in this pull request. :)

ykempf: OK I think I got it compiling. BRB

ykempf:
I... tried not to push the dev merge, with great success as you can see.

ykempf:
But this [c90c637](https://github.com/fmihpc/vlasiator/pull/664/commits/c90c6372f91a7731fc089faf8252496b5657011b) implements the fix you required @ursg.

ykempf:
How's this @ursg? (Test pending/queueing but methinks this is better now.)

sysboundary/ionosphere.cpp
```diff
@@ -3720,6 +3722,64 @@ namespace SBC {
...
+      }
+
+      const Real Bsqr = B[0]*B[0] + B[1]*B[1] + B[2]*B[2];
+
+      // Calculate cell bulk velocity as E x B / B^2
+      std::array<Real, 3> vDrift;
+      cellParams[CellParams::BULKV_FORCING_X] = (E[1] * B[2] - E[2] * B[1])/Bsqr;
+      cellParams[CellParams::BULKV_FORCING_Y] = (E[2] * B[0] - E[0] * B[2])/Bsqr;
+      cellParams[CellParams::BULKV_FORCING_Z] = (E[0] * B[1] - E[1] * B[0])/Bsqr;
+      cellParams[CellParams::FORCING_CELL_NUM]=1;
```
ykempf, line 3780: So for cells in the NOT_SYSBOUNDARY L2 halo we reset this to 0 in the acceleration code. But for the boundary cells this remains 1. However we anyway call this function here for boundary cells at 3802 so are we fine with this? It causes a slight befuddlement but resetting to zero is not necessary for the boundary cells. If you'd prefer a *pro forma* reset to be added, let me know.

ykempf:
Oh and my test now finally worked, after also some file system confusion... So I think this is good to go now.

vlasovsolver/cpu_acc_transform.cpp
```diff
@@ -146,5 +148,34 @@ Eigen::Transform<Real,3,Eigen::Affine> compute_acceleration_transformation(
...
 
+   // If a bulk velocity is being forced here, perform that last, after things were gyrated in the Hall frame
+   // If a cell is a remote L2 and was not caught in the loop over neighbours of L1 cells, compute its forcing here
+   // (same in sysboundary/ionosphere.cpp, regarding the time check:)
+   // TODO Make this a more elegant solution
+   // Now it's hacky as the counter is incremented in vlasiator.cpp
+   if( (((Parameters::t                > (SBC::Ionosphere::solveCount-1) * SBC::Ionosphere::couplingInterval)
+      && (Parameters::t-Parameters::dt < (SBC::Ionosphere::solveCount-1) * SBC::Ionosphere::couplingInterval)
+      && SBC::Ionosphere::couplingInterval > 0)
+      || SBC::Ionosphere::couplingInterval == 0)
```
ykempf, line 159: @markusbattarbee you noted this, it is a copy from the same mechanism in ionosphere.cpp, do we want to make this more elegant now? I think we have other cats to flog right now maybe?

sysboundary/ionosphere.cpp
markusbattarbee: I suppose a reset to zero would allow for debug DRO of this value so that we could verify that all necessary cells are being nudged?

ykempf: I have that DRO stashed, do you want it pushed? Or do you want to see some screenshots?

vlasovsolver/cpu_acc_transform.cpp
markusbattarbee: We do have many cats to flog, but if we don't flog this now, I fear it will remain unflogged for a long time, if not forever. We could do "bool globalflags::solveIonosphere = 0" in vlasiator.cpp and flip that with this test, no?

ykempf:
Before the cat flogging @ursg @markusbattarbee I pushed the fix to the smoothing, do you agree with the latest commit?

vlasovsolver/cpu_acc_transform.cpp
ykempf: [ec2f002](https://github.com/fmihpc/vlasiator/pull/664/commits/ec2f002601fbe5da0425e087713040cf6a71759d) is trying to fix this. Test in queue.

ursg:
Yep, these last two commits both look useful and sane to me.

ykempf:
OK it restarted and ran, nothing suspicious in the logfile so I believe this ran fine.

ykempf:
So this makes the "solve every interval" scheme work well (see Zulip) except at restarts, need to fix that. Tomorrow.
Also: this is not at all multipopy. But... how to make it multipop?

ykempf:
Yep so running the tests for a full day confirmed good results so the last fix is good. Now on to restarting.

ykempf:
Here are the latest two we discussed today.

markusbattarbee:
Automatic testpackage output for pull request #664.

Based on commit c46ab864c419257d628e1b0021ea28c00006d092.

Process started by user ```marbat``` on host ```mahti-login11.mahti.csc.fi``` at Wed, 21 Dec 2022 13:28:36 +0200.

:heavy_check_mark: Compilation succeeded with 12 warnings.

:large_orange_diamond: restart_read: Nonzero diffs: `fg_e_1` has absolute error 0.000504255, relative error 1.
:heavy_check_mark: Magnetosphere_small: Zero diffs.
:large_orange_diamond: Magnetosphere_polar_small: Nonzero diffs: `proton/vg_v_nonthermal_0` has absolute error 6.98492e-10, relative error 8.73517e-16.
:large_orange_diamond: Flowthrough_amr: Nonzero diffs: `proton/vg_rho_0` has absolute error 0.618127, relative error 2.86553e-07.
:large_orange_diamond: Magnetosphere_3D_small: Nonzero diffs: `proton/vg_v_0` has absolute error 3.29304e-06, relative error 4.3893e-12.
:red_circle: Ionosphere_small: Failed to run or died with an error.

Full testpackage run output is available in `/users/marbat/git/vlasiator/testpackage/testpackage_run_c46ab864c419257d628e1b0021ea28c00006d092.log` on host `mahti-login11.mahti.csc.fi`

markusbattarbee:
Automatic testpackage output for pull request #664.

Based on commit c46ab864c419257d628e1b0021ea28c00006d092.

Process started by user ```marbat``` on host ```mahti-login11.mahti.csc.fi``` at Wed, 21 Dec 2022 13:44:07 +0200.

:heavy_check_mark: Compilation succeeded with 12 warnings.

:large_orange_diamond: restart_read: Nonzero diffs: `fg_e_1` has absolute error 0.000504255, relative error 1.
:heavy_check_mark: Magnetosphere_small: Zero diffs.
:large_orange_diamond: Magnetosphere_polar_small: Nonzero diffs: `proton/vg_v_nonthermal_0` has absolute error 5.82077e-10, relative error 7.27931e-16.
:large_orange_diamond: Flowthrough_amr: Nonzero diffs: `proton/vg_rho_0` has absolute error 0.618126, relative error 2.86552e-07.
:large_orange_diamond: Magnetosphere_3D_small: Nonzero diffs: `proton/vg_v_0` has absolute error 3.87701e-06, relative error 5.16768e-12.
:large_orange_diamond: Ionosphere_small: Nonzero diffs: `proton/vg_v_0` has absolute error 0.0135841, relative error 1.81063e-08.

Full testpackage run output is available in `/users/marbat/git/vlasiator/testpackage/testpackage_run_c46ab864c419257d628e1b0021ea28c00006d092.log` on host `mahti-login11.mahti.csc.fi`

markusbattarbee:
nonzero diffs in AMR runs as expected, a bit bigger in ionosphere also as expected. And the diff in restart_read is, I presume, from the change of not running a zero-step fieldsolver at restart. So looks good.

ursg:
Yup, just scrolled through the diff again, that looks all reasonable.
Also thumbs up for the testpackage diffs.

PR #666, opened 2022-11-07
LUMI Cray compilers complain 

alhom:
Cray compilers on LUMI complain about ionosphere in datareduction: grid.productionNumParticleEnergies in a template argument (see below). Works on gnu toolchain, but not with cray, probably since grid is not a constexpr (even though the contained `int` is). Suggested solution: moving the constexpr to SBC namespace (or maybe elsewhere?). This anyway allows compilation on LUMI Cray as it is, now.

```
datareduction/datareducer.cpp:2861:40: error: non-type template argument is not a constant expression
                     std::array< Real, grid.productionNumParticleEnergies+1 > particle_energy;
                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
datareduction/datareducer.cpp:2861:40: note: function parameter 'grid' with unknown value cannot be used in a constant expression
datareduction/datareducer.cpp:2859:128: note: declared here
         outputReducer->addOperator(new DRO::DataReductionOperatorIonosphereNode("ig_precipnumflux", [](SBC::SphericalTriGrid& grid)->std::vector<Real> {
                                                                                                                               ^
                                                                                                                               ```

### DIFF
--- a/datareduction/datareducer.cpp
+++ b/datareduction/datareducer.cpp
@@ -2858,13 +2858,13 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
       if(lowercase == "ig_precipnumflux") {
          outputReducer->addOperator(new DRO::DataReductionOperatorIonosphereNode("ig_precipnumflux", [](SBC::SphericalTriGrid& grid)->std::vector<Real> {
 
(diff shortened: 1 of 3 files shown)


ursg:
Bizzarre compilers need bizarre solutions, I suppose. Whatever works.

PR #667, opened 2022-11-07
The Devil hides in details. Cleaning up #666.

ykempf:
@alhom do you also agree?


ursg:
Seems good. Does it compile? :)

ykempf:
It does! On my laptop, with updated DCCRG and VLSV!

ursg (approved):

alhom:
Yup; I did though use SBC:: explicitly in ionosphere.cpp as well. _Shouldn't_ matter, though.

PR #670, opened 2022-11-09
Lineprec dro

maximegrandin:
Created a new DRO for precipitating particle differential number flux calculation, designed to better capture low energies.
The DRO (hereinafter referred to as 'line DRO') is extremely similar to the original one ('loss-cone DRO') in its implementation, the main difference being the choice of the gate function to select relevant v-space cells for the flux calculation.

In the line DRO approach, we consider all cells which are traversed by the magnetic field direction (or opposite in the southern hemisphere), i.e., all cells which are intersected by the line along which precipitation takes place. This means that we overcome the issue found with the loss-cone DRO, in which at low energies there were no cell centres inside the loss cone due to the v-space resolution being too coarse for such low energies/velocities.

The only caveat to keep in mind is that, if a given energy bin is such that its width corresponds to a delta V lower than the diagonal size of velocity cells, this can still lead to zero flux locally in this energy bin. The way to overcome this is to carefully set the DRO parameters (emin, emax, nChannels) based on the v-space resolution.

The code has been tested using a restart from EGI, and precipitating fluxes from the line DRO are in excellent agreement with those from the loss-cone DRO (see figure). The test package has not been run on the new code at this stage.

spectrum_comparison

### DIFF
--- a/datareduction/datareducer.cpp
+++ b/datareduction/datareducer.cpp
@@ -410,7 +410,7 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
          continue;
       }
       if(lowercase == "populations_precipitationflux" || lowercase == "populations_vg_precipitationdifferentialflux" || lowercase == "populations_precipitationdifferentialflux") {
-         // Per-population precipitation differential flux
+         // Per-population precipitation differential flux (within loss cone)
          for(unsigned int i =0; i < getObjectWrapper().particleSpecies.size(); i++) {
             species::Species& species=getObjectWrapper().particleSpecies[i];
             const std::string& pop = species.name;
@@ -418,6 +418,18 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
 	    std::stringstream conversion;
 	    conversion << (1.0e-4)*physicalconstants::CHARGE;
 	    outputReducer->addMetadata(outputReducer->size()-1,"1/(cm^2 sr s eV)","$\\mathrm{cm}^{-2}\\,\\mathrm{sr}^{-1}\\,\\mathrm{s}^{-1}\\,\\mathrm{eV}^{-1}$","$\\mathcal{F}_\\mathrm{"+pop+"}$",conversion.str());
+         }
+         continue;
+      }
+      if(lowercase == "populations_precipitationlineflux" || lowercase == "populations_vg_precipitationlinedifferentialflux" || lowercase == "populations_precipitationlinedifferentialflux") {
+         // Per-population precipitation differential flux (along line)
+         for(unsigned int i =0; i < getObjectWrapper().particleSpecies.size(); i++) {
+            species::Species& species=getObjectWrapper().particleSpecies[i];
+            const std::string& pop = species.name;
+            outputReducer->addOperator(new DRO::VariablePrecipitationLineDiffFlux(i));
+            std::stringstream conversion;
+            conversion << (1.0e-4)*physicalconstants::CHARGE;
+            outputReducer->addMetadata(outputReducer->size()-1,"1/(cm^2 sr s eV)","$\\mathrm{cm}^{-2}\\,\\mathrm{sr}^{-1}\\,\\mathrm{s}^{-1}\\,\\mathrm{eV}^{-1}$","$\\mathcal{F}_\\mathrm{"+pop+"}$",conversion.str());
          }
          continue;
       }
--- a/datareduction/datareductionoperator.cpp
+++ b/datareduction/datareductionoperator.cpp
@@ -1717,7 +1717,7 @@ namespace DRO {
                yGateCrit = (BnormV[1] - (VY - 0.5*BlockParams::DVY)) * (BnormV[1] - (VY + 0.5*BlockParams::DVY)) <= 0;
                zGateCrit = (BnormV[2] - (VZ - 0.5*BlockParams::DVZ)) * (BnormV[2] - (VZ + 0.5*BlockParams::DVZ)) <= 0;
                bool xyzGateCrit = xGateCrit && yGateCrit && zGateCrit;  // gate function: 1 if the line goes through the v-cell, else 0.
-               countAndGate = (Real) xyzGateCrit;
+               Real countAndGate = (Real) xyzGateCrit;
                const Real energy = 0.5 * getObjectWrapper().particleSpecies[popID].mass * normV*normV; // in SI
 
                // Find the correct energy bin number to update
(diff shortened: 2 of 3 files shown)


datareduction/datareductionoperator.cpp
```diff
@@ -1627,6 +1626,149 @@ namespace DRO {
...
+    * Parameters that can be set in cfg file under [{species}_precipitation]: nChannels, emin [eV], emax [eV]
+    * The energy channels are saved in bulk files as PrecipitationCentreEnergyLine{channel_number}.
+    */
+   VariablePrecipitationLineDiffFlux::VariablePrecipitationLineDiffFlux(cuint _popID): DataReductionOperatorHasParameters(),popID(_popID) {
+      popName = getObjectWrapper().particleSpecies[popID].name;
+      emin = getObjectWrapper().particleSpecies[popID].precipitationEmin;    // already converted to SI
+      emax = getObjectWrapper().particleSpecies[popID].precipitationEmax;    // already converted to SI
+      nChannels = getObjectWrapper().particleSpecies[popID].precipitationNChannels; // number of energy channels, logarithmically spaced between emin and emax
+      for (int i=0; i<nChannels; i++){
+         channels.push_back(emin * pow(emax/emin,(Real)i/(nChannels-1)));
```
ykempf, line 1644: I know it's copied from the previous DRO but why is this with a `nChannels-1` denominator?

datareduction/datareducer.cpp
```diff
@@ -421,6 +421,18 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
          continue;
       }
+      if(lowercase == "populations_precipitationlineflux" || lowercase == "populations_vg_precipitationlinedifferentialflux" || lowercase == "populations_precipitationlinedifferentialflux") {
+         // Per-population precipitation differential flux (along line)
+         for(unsigned int i =0; i < getObjectWrapper().particleSpecies.size(); i++) {
+            species::Species& species=getObjectWrapper().particleSpecies[i];
+            const std::string& pop = species.name;
+            outputReducer->addOperator(new DRO::VariablePrecipitationLineDiffFlux(i));
+            std::stringstream conversion;
+            conversion << (1.0e-4)*physicalconstants::CHARGE;
```
ykempf, line 431: For my own edification: what is this for?

datareduction/datareductionoperator.cpp
```diff
@@ -1627,6 +1626,149 @@ namespace DRO {
...
+
+               // We will use a gate function based on criteria that Vi-0.5*DVi <= BnormV[i] <= Vi+0.5*DVi (for i=x,y,z or 0,1,2)
+               bool xGateCrit, yGateCrit, zGateCrit;
+	       const Real _DVX= parameters[n * BlockParams::N_VELOCITY_BLOCK_PARAMS + BlockParams::DVX];
+               const Real _DVY= parameters[n * BlockParams::N_VELOCITY_BLOCK_PARAMS + BlockParams::DVY];
+               const Real _DVZ= parameters[n * BlockParams::N_VELOCITY_BLOCK_PARAMS + BlockParams::DVZ];
+               xGateCrit = (BnormV[0] - (VX - 0.5*_DVX)) * (BnormV[0] - (VX + 0.5*_DVX)) <= 0;
+               yGateCrit = (BnormV[1] - (VY - 0.5*_DVY)) * (BnormV[1] - (VY + 0.5*_DVY)) <= 0;
+               zGateCrit = (BnormV[2] - (VZ - 0.5*_DVZ)) * (BnormV[2] - (VZ + 0.5*_DVZ)) <= 0;
+               bool xyzGateCrit = xGateCrit && yGateCrit && zGateCrit;  // gate function: 1 if the line goes through the v-cell, else 0.
```
ykempf, line 1722: The `bool` could be `const` but that's cosmetics, don't modify now unless @ursg has a reason to require those to be made const.

datareduction/datareductionoperator.cpp
```diff
@@ -1627,6 +1626,149 @@ namespace DRO {
...
+               Real countAndGate = (Real) xyzGateCrit;
+               const Real energy = 0.5 * getObjectWrapper().particleSpecies[popID].mass * normV*normV; // in SI
+
+               // Find the correct energy bin number to update
+               int binNumber = round((log(energy) - log(emin)) / log(emax/emin) * (nChannels-1));
```
ykempf, line 1727: Same here, why the nChannels-1?

ykempf:
So it's a go from me up to @ursg's decision on the consts.

datareduction/datareductionoperator.cpp
ursg: Nah, I'd wager that would just be premature optimization.

datareduction/datareductionoperator.cpp
```diff
@@ -1627,6 +1626,149 @@ namespace DRO {
...
+               // We will use a gate function based on criteria that Vi-0.5*DVi <= BnormV[i] <= Vi+0.5*DVi (for i=x,y,z or 0,1,2)
+               bool xGateCrit, yGateCrit, zGateCrit;
+	       const Real _DVX= parameters[n * BlockParams::N_VELOCITY_BLOCK_PARAMS + BlockParams::DVX];
+               const Real _DVY= parameters[n * BlockParams::N_VELOCITY_BLOCK_PARAMS + BlockParams::DVY];
+               const Real _DVZ= parameters[n * BlockParams::N_VELOCITY_BLOCK_PARAMS + BlockParams::DVZ];
+               xGateCrit = (BnormV[0] - (VX - 0.5*_DVX)) * (BnormV[0] - (VX + 0.5*_DVX)) <= 0;
+               yGateCrit = (BnormV[1] - (VY - 0.5*_DVY)) * (BnormV[1] - (VY + 0.5*_DVY)) <= 0;
+               zGateCrit = (BnormV[2] - (VZ - 0.5*_DVZ)) * (BnormV[2] - (VZ + 0.5*_DVZ)) <= 0;
+               bool xyzGateCrit = xGateCrit && yGateCrit && zGateCrit;  // gate function: 1 if the line goes through the v-cell, else 0.
+               Real countAndGate = (Real) xyzGateCrit;
```
ursg, line 1723: This is a bit unpretty, and it is by pure numerical coincidence that a true bool value happens to be bit-identical to a floating point number of value 1.
This would not be so on all hardware! Early cray machines used a different floating point memory layout, as did the VAX machines from digital research in the 70ies. But realistically: any computer vlasiator will ever run on has no problem with that.

So I'm fine with it. :)

ursg (approved):
Yep, looks great.

Most of the code is anyway pretty much identical with the old precipitation reducer, only differing in the implementation of the gate test.
Also great that the old reducer remains there, so we can run both side-by-side.

Let's have it (and I'll actually ship the partial-cell-slicing precipitation DRO for yet another comparison as a separate PR).

PR #671, opened 2022-11-10
Ionosphere finite difference derivatives at L1 and L2

ykempf:
Discussion point: the volume derivatives still skip, I guess that should be tested and added too?

### DIFF
--- a/fieldsolver/derivatives.cpp
+++ b/fieldsolver/derivatives.cpp
@@ -484,7 +484,7 @@ void calculateBVOLDerivatives(
    }
    
    // Calculate z-derivatives (is not TVD for AMR mesh):
-   if (technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY || technicalGrid.get(i,j,k)->sysBoundaryLayer == 1) {
+   if (sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY || sysBoundaryLayer == 1) {
       left = volGrid.get(i,j,k-1);
       rght = volGrid.get(i,j,k+1);
       
--- a/fieldsolver/derivatives.cpp
+++ b/fieldsolver/derivatives.cpp
@@ -430,9 +430,13 @@ void calculateDerivativesSimple(
 }
 
 /*! \brief Low-level spatial derivatives calculation.
- * 
- * For the cell with ID cellID calculate the spatial derivatives of BVOL or apply the derivative boundary conditions defined in project.h.
- * 
+ *
+ * Calculate the spatial derivatives of BVOL or set them to zero.
+ *
+ * For sysBoundaryLayer 1 or 2, we are near a boundary, and we wish to use regular centered differences instead of slope limiter-adjusted values.
+ * This is to minimize oscillations as a smooth behaviour is required near artificial boundaries,
+ * unlike at boundaries and shocks inside the simulation domain.
+ *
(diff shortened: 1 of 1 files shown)


ykempf:
Committed, compiles and a test is in queue.

markusbattarbee:
The questions really are:
1) Is this differencing scheme the one we want to use?
 because
 a) this does gives better FACs
 b) we still don't know how well this behaves in a longer run

We know that zeroing derivatives gives worse FACs but at least it doesn't bork the runs. Once we verify that either a) this branch runs well for a long time or b) the spring runs which caused this zeroing of derivatives actually had incorrect tapering, I think we can merge.

ykempf:
Pieni whoopsie fixed here, but no difference in the end as I had just replaced repeated calls to the grids with a stored flag as in other parts of the code. Possibly the compilers would optimise it to the same anyway.

ykempf:
So this behaves well (in conjunction with other branches like #664) in runs up to a good 600 s, are we good with this going in?

markusbattarbee:
Perhaps add a comment at the first instance, along the lines of:
"For sysBoundaryLayer 1 or 2, we are near a boundary, and we wish to use regular centered differences instead of slope limiter -adjusted values. This is to minimize oscillations as a smooth behaviour is required near artificial boundaries, unlike at boundaries and shocks inside the simulation domain."

ykempf:
Like so? Now in the doxygen comment at the top of the function.

fieldsolver/derivatives.cpp
```diff
@@ -149,17 +148,11 @@ void calculateDerivatives(
         dPerB->at(fsgrids::dperb::dPERBzdxx) = leftPerB->at(fsgrids::bfield::PERBZ) + rghtPerB->at(fsgrids::bfield::PERBZ) - 2.0*centPerB->at(fsgrids::bfield::PERBZ);
       }
    } else {
-      // Boundary conditions handle derivatives.
-//      if (sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY) {
-         SBC::SysBoundaryCondition::setCellDerivativesToZero(dPerBGrid, dMomentsGrid, i, j, k, 0);
-//      } else {
-//         sysBoundaries.getSysBoundary(sysBoundaryFlag)->fieldSolverBoundaryCondDerivatives(dPerBGrid, dMomentsGrid, i, j, k, RKCase, 0);
-//      }
```
ykempf, line 136: Three like this are just dropping the commented sections.

markusbattarbee:
It's an approve from me now.

fieldsolver/derivatives.cpp
```diff
@@ -270,60 +252,36 @@ void calculateDerivatives(
       dPerB->at(fsgrids::dperb::dPERBzdxy) = 0.0;
    } else {
       // Calculate xy mixed derivatives:
-      if ((sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY) || (sysBoundaryLayer == 1)) {
-         
+      if (sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY) {
```
ykempf, line 255: Three like this as the if of which we are in the else checks for sysBoundaryLayer == 1.

ursg (approved):
It's not overly pretty (a quick readthrough of the changed code gave me the "It looks reasonable, but I wouldn't have been able to catch if there was a typo in there"-feeling), but it appears to work, and that's what counts.

PR #669, opened 2022-11-07
Fix filtering

lkotipal:
Fixes filtering for outer boundaries with non-zero refinement level. Specifically the issue was that the filtering stencil filtered cells with boundarylayer 1, which segfaults on non-existing cells if the outer boundary cells are on the highest refinement level. The fix is to skip all sysboundary cells when filtering.

PR #673, opened 2022-11-14
Fix magnetosphere project tapering when running with an ionosphere

markusbattarbee:
Fixes the magnetosphere project tapering when running with an ionosphere. (Old method always went for conductingsphere parameters which were set to default values)

Implements a new method for checking existence of boundary conditions, and also adds an ionospheric testpackage test.
In Vlasiator.cpp there's a renaming of the wrapper object for sysBoundaries as I don't like having a class and an instance of that class with basically the same name.


projects/Magnetosphere/Magnetosphere.h
```diff
@@ -117,5 +118,7 @@ namespace projects {
    }; // class Magnetosphere
 } // namespace projects
 
+extern SysBoundary sysBoundaryWrapper;
```
ursg, line 121: I don't quite see yet why this separate object is necessary. Is it just so that the existence of boundaries can be tested through a global object somewhere?

Renaming all the sysboundary accesses further down in vlasiator.cpp will probably again lead to conflicts with the time varying inflow branch (although they won't be hard to correct)

sysboundary/sysboundary.cpp
```diff
@@ -343,6 +343,25 @@ bool SysBoundary::initSysBoundaries(Project& project, creal& t) {
    return success;
 }
 
+/*!\brief Boolean check if queried sysboundarycondition exists
+ * Note: this queries against the parsed list of names, not against the actual existing boundaries. 
+ * This is so that the project configuration (which is handled before grid and boundary initialization)
+ * can use this call.
```
ursg, line 349: Big thumbs up for an actually helpful and constructive comment here!

vlasiator.cpp
```diff
@@ -71,6 +71,7 @@ void fpehandler(int sig_num)
 
 Logger logFile, diagnostic;
 static dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry> mpiGrid;
+SysBoundary sysBoundaryWrapper;
```
ursg, line 74: Instead of placing this here as a global variable, another option would be to move it into the objectWrapper. That way, we are not further polluting the global variable namespace and have a little bit of a better encapsulation of global simulation state.

fieldsolver/gridGlue.cpp
```diff
@@ -161,12 +161,7 @@ void filterMoments(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
                int refLevel = technicalGrid.get(i, j, k)->refLevel;
 
                // Skip pass
-               if (blurPass >= P::numPasses.at(refLevel) ||
-                  technicalGrid.get(i, j, k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE ||
-                  (technicalGrid.get(i, j, k)->sysBoundaryFlag != sysboundarytype::NOT_SYSBOUNDARY &&
-                  technicalGrid.get(i, j, k)->sysBoundaryLayer >= 2)
-               )
-               {
+               if (blurPass >= P::numPasses.at(refLevel) || technicalGrid.get(i, j, k)->sysBoundaryFlag != sysboundarytype::NOT_SYSBOUNDARY) {
```
ursg, line 164: This is basically PR #669, isn't it?

ursg (approved):
Good fixes! My critical comments here in the review probably shouldn't block this from being merged. So let's go for it!

fieldsolver/gridGlue.cpp
markusbattarbee: Yes, I pulled that commit directly into this branch/PR thinking it'd neatly ignore it here, but that isn't the way of github it seems.

projects/Magnetosphere/Magnetosphere.h
markusbattarbee: Yep, I think passing sysboundaries down to all projects doesn't make much sense for this special case, so a global object was my preferred choice.
The naming may indeed cause some clash, but because it's been just a local object and not part of an interface, it should be minor (and I'd say helpful).

vlasiator.cpp
markusbattarbee: That would indeed be an alternative option. I'l look into it.

markusbattarbee:
So now it's better. Should work as intended everywhere. 

Had a look at plots of the various magnetospheric tests and they seem to be somewhat sane. 

I changed the parametrization of the 3D tests (level 0 cell size is now 30e6 m, instead of 40e6 for ionosphere or 27,2727272...e6 for Magnetosphere_3D_small, and extended the tails a bit.

Also now the refinement level check for boundaries is less strict (nearest neighbours should be enough) so we don't have tons of error messages in testpackage runs.

Diffs against dev are zero / testpackage numeric level everywhere except for Ionosphere where it's a tapering fix, of course. How about now @ursg ?

ykempf (commented):
Also, do we want to make use of non-uniform temperatures with tapering to make the test case less homogeneous and more sensitive to stuff?

sysboundary/ionosphere.cpp
```diff
@@ -2978,11 +2978,14 @@ namespace SBC {
         Readparameters::get(pop + "_ionosphere.VZ0", sP.V0[2]);
         Readparameters::get(pop + "_ionosphere.T", sP.T);
 
+        // Failsafe, if density or temperature is zero, read from Magnetosphere
+        // (compare the corresponding verbose handling in projects/Magnetosphere/Magnetosphere.cpp)
```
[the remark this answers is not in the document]
markusbattarbee: The warning is printed in Magnetosphere.cpp, that's why it now says there's *verbose* handling there :)

sysboundary/conductingsphere.cpp
```diff
@@ -100,7 +100,16 @@ namespace SBC {
         Readparameters::get(pop + "_Magnetosphere.nSpaceSamples", sP.nSpaceSamples);
         Readparameters::get(pop + "_Magnetosphere.nVelocitySamples", sP.nVelocitySamples);
 
-        speciesParams.push_back(sP);
+        // Failsafe, if density or temperature is zero, read from Magnetosphere
+        // (compare the corresponding verbose handling in projects/Magnetosphere/Magnetosphere.cpp)
```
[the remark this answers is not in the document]
markusbattarbee: and same, done in Magnetosphere.cpp

testpackage/run_tests.sh
```diff
@@ -136,31 +136,35 @@ do
...
                 absoluteValue=$($run_command_tools $diffbin --meshname=fsgrid  ${result_dir}/${comparison_vlsv[$run]} ${vlsv_dir}/${comparison_vlsv[$run]} ${variables[$i]} ${indices[$i]} |grep "The absolute 0-distance between both datasets" |gawk '{print $8}'  )
-#print the results      
+#print the results
+                echo "${variables[$i]}_${indices[$i]}                $absoluteValue                 $relativeValue    "
+
+            elif [[ "${variables[$i]}" == "ig_"* ]]
+            then
+                relativeValue=$($run_command_tools $diffbin --meshname=ionosphere  ${result_dir}/${comparison_vlsv[$run]} ${vlsv_dir}/${comparison_vlsv[$run]} ${variables[$i]} ${indices[$i]} |grep "The relative 0-distance between both datasets" |gawk '{print $8}'  )
+                absoluteValue=$($run_command_tools $diffbin --meshname=ionosphere  ${result_dir}/${comparison_vlsv[$run]} ${vlsv_dir}/${comparison_vlsv[$run]} ${variables[$i]} ${indices[$i]} |grep "The absolute 0-distance between both datasets" |gawk '{print $8}'  )
```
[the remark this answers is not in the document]
markusbattarbee: Indeed no, awaiting the vlsvdiff update.

projects/Magnetosphere/Magnetosphere.cpp
```diff
@@ -117,34 +118,28 @@ namespace projects {
...
-      } else {
+      } else if (sysBoundaryContainer.existSysBoundary("Ionosphere")) {
+         RP::get("ionosphere.radius", this->ionosphereRadius);
+         RP::get("ionosphere.centerX", this->center[0]);
+         RP::get("ionosphere.centerY", this->center[1]);
+         RP::get("ionosphere.centerZ", this->center[2]);
          RP::get("ionosphere.geometry", this->ionosphereGeometry);
+      } else {
+         if(myRank == MASTER_RANK) {
+            std::cerr<<"Warning in initializing Magnetosphere: Could not find inner boundary geometry!"<<std::endl;
```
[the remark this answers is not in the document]
markusbattarbee: Ok, updated message

markusbattarbee:
> Also, do we want to make use of non-uniform temperatures with tapering to make the test case less homogeneous and more sensitive to stuff?

We could indeed. I put this in the ionospheric one:
```
T = 0.8e6
rho  = 2.0e6
```
and could well put some similar small increases in the other ones if we want?

ykempf:
> and could well put some similar small increases in the other ones if we want?

If we want to catch more readily issues, the more gradients the merrier...

ursg:
Code looks good to me now. Feel free to merge! (Or still change the tests and then merge, at your own discretion.)

PR #674, opened 2022-11-21
Cut-cell precipitation DRO

ursg:
As yet another alternative to the line-cutting precipitation DRO in PR#670, this version implements an implementation where each velocity cell is coverage-tested against the loss cone, allowing for partial coverage to contribute partially.

The concrete implementation defines the loss cone as a signed distance function, which returns the distance to the cone surface for every point in vspace. If this distance is much larger than the vspace cell's diagonal, the cell is assumed to be clearly outside. If it is negative by the same distance, it is clearly inside.
But cells that are closer to the cone boundary then their diagonal are recursively octree-subdivided, and each subcell is again tested in the same way (up to a hardcoded recursion limit of 6).

### DIFF
--- a/datareduction/datareducer.cpp
+++ b/datareduction/datareducer.cpp
@@ -421,6 +421,18 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
          }
          continue;
       }
+      if(lowercase == "populations_precipitationfluxcutcells" || lowercase == "populations_vg_precipitationdifferentialfluxcutcells" || lowercase == "populations_precipitationdifferentialfluxcutcells") {
+         // Per-population precipitation differential flux
+         for(unsigned int i =0; i < getObjectWrapper().particleSpecies.size(); i++) {
+            species::Species& species=getObjectWrapper().particleSpecies[i];
+            const std::string& pop = species.name;
+            outputReducer->addOperator(new DRO::VariablePrecipitationDiffFluxCutCells(i));
+	    std::stringstream conversion;
+	    conversion << (1.0e-4)*physicalconstants::CHARGE;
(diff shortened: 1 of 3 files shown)


ursg:
@maximegrandin, since you apparently already have a nice and friendly test setup for comparing these precipitation DROs (as per your figure in #670), do you want to give this one a spin as well, for comparison?
It's parametrized exactly the same way (with the same parameter names) as the original loss cone DRO, but it's name is ```populations_vg_precipitationdifferentialfluxcutcells```

maximegrandin:
Kostis ran a test with the three precipitation DROs (loss cone, line, and cut cells) altogether, with 16 energy bins logarithmically spaced from 50 eV to 50 keV (loss cone angle of 5° for the first DRO).

I first plotted precipitation at ~2 keV in the Y = 0 slice with each DRO (figure order: loss cone, line, cut cells).
prec8_losscone
prec8_line
prec8_cutcells
We can see that, while the first two DROs give almost identical results, the cut-cell one has a somewhat strange behaviour, with proton "precipitation" at 2 keV being obtained in the whole simulation domain (including in the solar wind).

Then, I plotted the precipitating proton spectra obtained with the three DROs at a selected location in the plasma sheet (–7.5 Re, 0, 2.5 Re).
spectrum
Again, while the first two DROs are in very good agreement in the energy range where both are able to produce data, the cut-cell one overall gives higher fluxes, especially at E < 800 eV and at E > 12 keV. Given that the line and loss-cone DROs are two "extreme cases" (the former underestimating the true loss cone angle, the latter overestimating it), and based on Fig. 15 of Grandin et al. (in review, JSWSC), I would expect that the cut-cell DRO should overall give similar results compared to the other two (in fact, these would likely lie within the thin interval of flux values between the two).

This suggests that there might be an issue in the implementation of the algorithm, as the general idea should in principle work, I think.

ursg:
Uhh, ok. Thanks for testing!
Hmm, this means this PR will require a bit more debugging still, and not make it in for the next big 3D run.

PR #675, opened 2022-11-22
Move the IONOSPHERE_SORTED_SUMS flag to the testpackage flags in Makefile and not in the arch files

ykempf:
All simple and trivial. Except I mistakenly pushed a half-baked LUMI Makefile.


ykempf:
I updated the modules in the accidental LUMI Makefile.

markusbattarbee:
Sadly I fear this won't work - as I'm the one who suggested this approach, I did also try it in the ionosphere tapering fix PR. When going into testpackage, it overwrites CXXFLAGS in one of the recursive inclusions of makefiles. The logic can be a bit confounding.

ykempf:
This... was easy. Thanks to @sandroos or @galfthan originally I guess for laying out these flags in a future-proof way. :)

ursg:
Looks good indeed. I'll quickly sanity-check that this works on all important machines, then let's merge it.

ykempf:
@ursg did you check that?

ursg:
Yups, looking good. This *does* have a trivial merge conflict with #676, but I can handle that one.

PR #679, opened 2022-12-20
Vlsvdiff ionosphere support

ursg:
This code has been lying around for... quite a while already, and I never turned it into a PR. So here's a patch for vlsvdiff so that it can also deal with ionosphere meshes.

This *also* contains some previous changes to AMR grid handling, which I am not so sure about.

### DIFF
--- a/tools/vlsvdiff.cpp
+++ b/tools/vlsvdiff.cpp
@@ -503,6 +503,7 @@ bool convertMesh(vlsvinterface::Reader& vlsvReader,
          cerr << "ERROR, failed to get array info for '" << _varToExtract << "' at " << __FILE__ << " " << __LINE__ << endl;
          return false;
       }
+      std::cerr << "read" << std::endl;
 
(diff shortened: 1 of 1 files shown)


ykempf (commented):
Well this has worked for me so I think we're good. Unless the diff we were surprised by a few minutes ago is actually a vlsvdiff feature?!...

ykempf:
`--diff` segfaults on ionosphere variables.

ursg:
> `--diff` segfaults on ionosphere variables.

Fixed.
This now produces diff output files that visit can plot (and that look reasonable).

markusbattarbee:
Could we activate the Ionosphere_small testpackage variable checks for ionospheric variables for this PR? Then we'd see if they work.

ursg:
Does our reference data already include them? If yes, then this should work right away. Let me try.

ursg:
So, this looks great!

PR #680, opened 2022-12-21
Fixes to LUMI Makefile.

ykempf:
Please someone else inspect for blunders too...


markusbattarbee:
I think you still need
`LIB_PAPI = -lpapi`
when using the papi module.

PR #678, opened 2022-12-20
Remove spatial mesh_data constructs

markusbattarbee:
There was a spatial mesh_data construct which re-used velocity mesh parametrizations, but was not actually used for anything anymore. This PR prunes that code out - I think it must have been an ancient debugging option.

This pruning is performed so that the velocity mesh parametrization can be re-constructed to be portable to GPUs in the future.

PR #668, opened 2022-11-07
Use Eigen Vector3d instead of Agner Vectorclass in the ionosphere

ursg:
Minimally-invasive change of vectorclass inside inosphere.cpp (by doing a couple of defines at the top). This should make the ionosphere compile with fallback vectorclass again.

Compiles, but otherwise untested so far.

Hooray for having three different vector implementations in the code at the same time! So you can always choose!

PR #685, opened 2023-01-02
Cudasiator

markusbattarbee:
Brings Cudasiator up to date with dev; also removes old spatial_mesh constructs and switches the ionosphere and fieldtracer routines to use Eigen instead of vectorclass.


ursg:
Way too big a diff to scroll through all of it, but I'm conceptually approving that everything goes up to date.

markusbattarbee:
Tested that the CUDA version compiles and runs. Made this into a PR instead of directly merging so @hokkanen can see if it breaks things for him

ursg:
Does this now need a rebase, now that #686, #668 and #678 are merged in dev? Or should it keep going as such?

markusbattarbee:
Ho hum. Yeah I think I'll need to do that, and it might be a bit messy as some of those changes were include here, but not necessarily always from the same commit.

PR #686, opened 2023-01-05
Renamed all velocity-related AMR messages and variables to VAMR, clea…

markusbattarbee:
…ned up a few remnant warnings, added sized deallocations.

Tried a bit to compile with VAMR as well but that codepath has indeed gone a bit too stale. Running testpackage now.


ursg:
This seems to have a number of other small things along with it. Not that I disapprove, looks like general cleanup and warnings reduction.

markusbattarbee:
Can confirm from a testpackage run, only numerical level diffs in AMR runs, so this is good.

ursg:
Thumbs up for merging from me then!

Issue #681, opened 2022-12-21
Field tracing: features of flux rope tracing


ykempf:
Performance is not great in early phases and we are tracing some rather not-so-useful regions.
- Include the tracing fraction like the full box tracing
- Trace only if the curvature radius is below a cfg threshold

lassejsc:
This seems to already done and merged, closing. Reopen if necessary

PR #687, opened 2023-01-11
Field tracing fluxrope improvements

ykempf:
For now at least covering #681.


ykempf:
That works, another push coming in when the test is through.

ursg:
So far, so good.

ykempf:
image
Left is up to here, right is the next commit. I think this is OK.

ykempf:
Pending plotting confirmation if I get things to behave, the latter ones now get us where we were planning to go to.

ykempf:
```
| 2   |   Simulation                                              | 1       | 408.8     | 55.67     |
| 3   |     IO                                                    | 1       | 181.4     | 44.38     |
| 4   |       fieldtracing-ionosphere-openclosedTracing           | 1       | 3.467     | 1.911     |
| 4   |       fieldtracing-fullTracing                            | 1       | 40.03     | 22.06     |
| 4   |       fieldtracing-fluxropeTracing                        | 1       | 39.85     | 21.96     |
| 4   |       fieldtracing-fullAndFluxTracing                     | 1       | 52.72     | 29.05     |
```

ykempf:
The last commit is my hack to decrease the size of the arrays going to MPI global reductions. This had a massive improvement at production scale. But it seems I have some small glitches left in the results so some more inspection is due. And then I'll remove the redundant code and debugging output.

fieldtracing/fieldtracing.cpp
```diff
@@ -793,34 +790,40 @@ namespace FieldTracing {
       creal stepSize = min(1000e3, technicalGrid.DX / 2.);
       std::vector<Real> cellFWTracingStepSize(globalDccrgSize, stepSize); // In-flight storage of step size, needed when crossing into next MPI domain
       std::vector<Real> cellBWTracingStepSize(globalDccrgSize, stepSize); // In-flight storage of step size, needed when crossing into next MPI domain
+      
       std::array<int, 3> gridSize = technicalGrid.getGlobalSize();
-      uint64_t maxTracingSteps = 4 * (gridSize[0] * technicalGrid.DX + gridSize[1] * technicalGrid.DY + gridSize[2] * technicalGrid.DZ) / stepSize;
+      creal maxTracingDistance = 4 * (gridSize[0] * technicalGrid.DX + gridSize[1] * technicalGrid.DY + gridSize[2] * technicalGrid.DZ);
```
ykempf, line 795: Do we keep this heuristic (it triggers e.g. in FHA at 1000 s when tracing all cells) or do we want yet another cfg parameter?

fieldtracing/fieldtracing.cpp
```diff
@@ -996,114 +1010,213 @@ namespace FieldTracing {
...
-               cellBWTracingStepSize[n] = reducedCellBWTracingStepSize[n];
             }
             #pragma omp barrier
-         } while(anyCellNeedsTracing && (cellsToDo >= fieldTracingParameters.max_incomplete_lines_fullbox * 2 * globalDccrgSize));
+         } while(!(
+            cellsToDoFullBox <= fieldTracingParameters.fullbox_max_incomplete_lines * 2 * globalDccrgSize
+            && cellsToDoFluxRopes <= fieldTracingParameters.fluxrope_max_incomplete_lines * 2 * globalDccrgSize
+         ));
          
          // Last pass to sort cells that would still continue but won't as we exited as TracingLineEndType::LOOP instead of UNPROCESSED.
```
ykempf, line 1174: This seemed like a good idea, less so now. The larger the leftover fraction we set, the less sensible this is. Conversely with a small fraction we're more likely to have indeed leftover loops mostly. Opinions? Leave unprocessed? Set to loop? Make the choice a cfg option?

ykempf:
This is ready to go but please do have a field day, ask questions and poke holes at what I did. With the deletion of the old functions I just pushed the diff is more useful again.

fieldtracing/fieldtracing.cpp
```diff
@@ -996,114 +1010,213 @@ namespace FieldTracing {
...
-                  MPI_Allreduce(cellBWTracingStepSize.data(), reducedCellBWTracingStepSize.data(), globalDccrgSize, MPI_FLOAT, MPI_MAX, MPI_COMM_WORLD);
+                  MPI_Allreduce(smallCellFWTracingCoordinates.data(), smallSumCellFWTracingCoordinates.data(), 3*smallSizeFW, MPI_FLOAT, MPI_SUM, MPI_COMM_WORLD);
+                  MPI_Allreduce(smallCellBWTracingCoordinates.data(), smallSumCellBWTracingCoordinates.data(), 3*smallSizeBW, MPI_FLOAT, MPI_SUM, MPI_COMM_WORLD);
+                  MPI_Allreduce(smallCellFWTracingStepSize.data(), smallReducedCellFWTracingStepSize.data(), smallSizeFW, MPI_FLOAT, MPI_MAX, MPI_COMM_WORLD);
+                  MPI_Allreduce(smallCellBWTracingStepSize.data(), smallReducedCellBWTracingStepSize.data(), smallSizeBW, MPI_FLOAT, MPI_MAX, MPI_COMM_WORLD);
+                  MPI_Allreduce(smallCellFWRunningDistance.data(), smallReducedCellFWRunningDistance.data(), smallSizeFW, MPI_FLOAT, MPI_MAX, MPI_COMM_WORLD);
+                  MPI_Allreduce(smallCellBWRunningDistance.data(), smallReducedCellBWRunningDistance.data(), smallSizeBW, MPI_FLOAT, MPI_MAX, MPI_COMM_WORLD);
```
ursg, line 1113: These are a stunningly large of number of allreduces here, of course. Would there be any performance merit to accumulating all this data into one struct, do an array of those structs, and then reduce all of that in a single Allreduce call?
(This is not a blocker against this specific pull request, just something to consider as a performance optimization here)

fieldtracing/fieldtracing.cpp
ursg: Hmm, I'm personally in favour of explicitly marking them as UNPROCESSED, just to be able to afterwards quantify how confident we are with the result. In most cases, we'll then afterwards just treat them the same way anyway.

fieldtracing/fieldtracing.cpp
```diff
@@ -996,114 +1010,213 @@ namespace FieldTracing {
...
          }
       } // pragma omp parallel
       phiprof::stop("loop");
       
-      logFile << "(fieldtracing) full box tracing traced in " << itCount << " iterations of the tracing loop with " << cellsToDo << " remaining incomplete field lines (total spatial cells " << globalDccrgSize <<  ")." << endl;
+      logFile << "(fieldtracing) combined flux rope + full box tracing traced in " << itCount
+         << " iterations of the tracing loop with flux rope " << cellsToDoFluxRopes
+         << ", full box " << cellsToDoFullBox
+         << " remaining incomplete field lines (total spatial cells " << globalDccrgSize
+         << ")." << endl;
```
ursg, line 1191: As soon as we are fully confident that the tracing works to our satisfaction, we could put this one behind a "DEBUG" define as well.
But obviously, that is not yet so. :)

fieldtracing/fieldtracing.h
```diff
@@ -93,10 +94,11 @@ namespace FieldTracing {
    * UNPROCESSED: cells inside the ionosphere or outside the outer limits that weren't even processed in the first place
    */
    enum TracingLineEndType {
+      UNPROCESSED, // Keep first for the reductions to work!
```
ursg, line 97: The only disadvantage of having UNPROCESSED first, is that you can't as easily plot "LOOP" and "UNPROCESSED" together in Visit by simply selecting a range of adjacent numbers. But that's not a major issue anyway.

ursg: (Also, doesn't changing the order mid-run of FHA make analysis a bit wonky?)

ursg (approved):
Looking good! My review comments are rather minor here, and not blocking this being merged.

fieldtracing/fieldtracing.cpp
ykempf: Well it's shaken out of a sleeve like "how far do we sensibly go in a box, thinking of not-too-complex IMF+dipole".

fieldtracing/fieldtracing.cpp
ykempf: So I did an attempt last summer of merging the FW and BW arrays into one array of double the size and it took exactly the same amount of time. Now *it is well possible* that that test was botched and there would be more to gain, but I am not willing to invest a lot more into this at this stage since we reached doable performance levels. Also note that not all of them do the same reduction.

fieldtracing/fieldtracing.cpp
ykempf: I think it's useful to have some idea of how it behaves at runtime. If we see insane amounts of iterations we'll want to change parameters.

fieldtracing/fieldtracing.h
ykempf: Up to 1000 s we had 5% leftover and the plots are ugly anyway. It'll be easy to just not look at that earlier bit of the data. Also, we don't ave access to this, only to the TracingPointConnectionType.

fieldtracing/fieldtracing.cpp
ykempf: OK. I'll wait for others' opinion but I agree it'd be best to leave them tagged as UNPROCESSED.

markusbattarbee: I agree - if it's a small fraction, then we can indeed re-evaluate them in post.

ykempf: OK pushed this, removed this last pass.

fieldtracing/fieldtracing.cpp
```diff
@@ -996,114 +1010,213 @@ namespace FieldTracing {
...
-               anyCellNeedsTracing = false;
             }
             #pragma omp barrier
             phiprof::stop("MPI-loop");
             #pragma omp single
             {
-               cellsToDo = 0;
+               cellsToDoFullBox = 0;
+               cellsToDoFluxRopes = 0;
+               cellFWTracingStepSize = reducedCellFWTracingStepSize;
```
markusbattarbee, line 1136: These are vector deep copies. You could do a swap() on them instead to reduce data flow.

A deep copy is needed for the NeedsContinuedXWTracing vectors though, as you apparently use the "old" one to keep track of active tracings still.

fieldtracing/fieldtracing.cpp
```diff
@@ -996,114 +1010,213 @@ namespace FieldTracing {
...
             for(int n=0; n<globalDccrgSize; n++) {
-               if(reducedCellNeedsContinuedFWTracing[n] > 0) {
-                  anyCellNeedsTracing=true;
-                  cellNeedsContinuedFWTracing[n] = 1;
-                  cellsToDo++;
-                  
+               if(cellNeedsContinuedFWTracing[n] > 0) {
+                  cellsToDoFullBox++;
                   // Update that nodes' tracing coordinates
                   cellFWTracingCoordinates[n][0] = sumCellFWTracingCoordinates[n][0] / reducedCellNeedsContinuedFWTracing[n];
```
markusbattarbee, line 1150: What's going on here? Isn't that a division by 1 always (when inside this if)?

fieldtracing/fieldtracing.cpp
markusbattarbee: I was also wondering in case allgather could be used to move around just those tracings which are still active, as there's only ever one "proper active" sender for a trace, but of course for each task, it can have a wide and sparse collection of indices n it's handling at any given time. 

I'm sure there'd be some more elegant object-oriented way to do all this, but in the end, this works, it's a great new feature, and our collaboration has been able to optimize it quite a bit already.

fieldtracing/fieldtracing.cpp
```diff
@@ -996,114 +1010,213 @@ namespace FieldTracing {
...
+                           + (x[2]-(cellInitialCoordinates[n])[2])*(x[2]-(cellInitialCoordinates[n])[2])
+                        );
+                        // ...and if we traced too far from the seed, this is not a flux rope candidate and we do a single +=
+                        if(distance > fieldTracingParameters.fluxrope_max_curvature_radii_extent*cellCurvatureRadius[n]
+                           || cellBWRunningDistance[n] > fieldTracingParameters.fluxrope_max_m_to_trace
+                        ) {
+                           cellBWConnection[n] += TracingLineEndType::N_TYPES;
+                        } else if(cellBWRunningDistance[n] > fieldTracingParameters.fluxrope_max_curvature_radii_to_trace*cellCurvatureRadius[n]) {
+                           // If we're still in the game and reach this limit we have a hit and we do a double +=
+                           cellBWConnection[n] += 2*TracingLineEndType::N_TYPES;
```
markusbattarbee, line 1043: Could you please add some comment description of how this connection type counting is supposed to work?

fieldtracing/fieldtracing.cpp
```diff
@@ -793,34 +790,40 @@ namespace FieldTracing {
...
-      std::vector<int> cellNeedsContinuedFWTracing(globalDccrgSize, 1);                    /*!< Flag, whether tracing needs to continue on another task */
-      std::vector<int> cellNeedsContinuedBWTracing(globalDccrgSize, 1);                    /*!< Flag, whether tracing needs to continue on another task */
-      std::vector<std::array<Real, 3>> cellFWTracingCoordinates(globalDccrgSize);          /*!< In-flight node upmapping coordinates (for global reduction) */
-      std::vector<std::array<Real, 3>> cellBWTracingCoordinates(globalDccrgSize);          /*!< In-flight node upmapping coordinates (for global reduction) */
+      std::vector<int> cellFWConnection(globalDccrgSize, TracingLineEndType::UNPROCESSED); /*!< For reduction of node coupling */
+      std::vector<int> cellBWConnection(globalDccrgSize, TracingLineEndType::UNPROCESSED); /*!< For reduction of node coupling */
+      
+      std::vector<int> cellNeedsContinuedFWTracing(globalDccrgSize, 1); /*!< Flag, whether tracing needs to continue on another task */
+      std::vector<int> cellNeedsContinuedBWTracing(globalDccrgSize, 1); /*!< Flag, whether tracing needs to continue on another task */
```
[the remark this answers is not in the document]
ykempf: Bool won't do, but yep, done.

fieldtracing/fieldtracing.cpp
```diff
@@ -53,11 +53,8 @@ namespace FieldTracing {
       if(fieldTracingParameters.doTraceOpenClosed) {
          traceOpenClosedConnection(technicalGrid, perBGrid, dPerBGrid, nodes);
       }
-      if(fieldTracingParameters.doTraceFullBox) {
-         traceFullBoxConnection(technicalGrid, perBGrid, dPerBGrid, mpiGrid);
-      }
-      if(fieldTracingParameters.doTraceFluxRopes) {
-         traceFluxRopes(technicalGrid, perBGrid, dPerBGrid, mpiGrid);
+      if(fieldTracingParameters.doTraceFullBox || fieldTracingParameters.doTraceFluxRopes) {
```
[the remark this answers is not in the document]
ykempf: Done

fieldtracing/fieldtracing.cpp
```diff
@@ -996,114 +1010,202 @@ namespace FieldTracing {
...
+                        #pragma omp critical
+                        {
+                           warnMaxDistanceExceeded = true;
+                        }
                         break;
                      }
                      
+                     // If we are still in the race for flux rope...
+                     if(cellBWConnection[n] < TracingLineEndType::N_TYPES) {
+                        creal distance = sqrt(
```
markusbattarbee, line 1031: So, as a reminder form our zoom: Distance can be stored in an array to be placed in the fluxrope cellparam, and you only need to do one MPI_ALLREDUCE at the end with MPI_MAX in order to get the maximum followed distance from all mpi-tasks for each cell.

ykempf: Yep, thanks! So do we store the distance in m or in curvature radii? And the max number of curvature radii traced as a VLSV parameter?

fieldtracing/fieldtracing.cpp
```diff
@@ -929,23 +927,51 @@ namespace FieldTracing {
...
                      // Now, after stepping, if it is no longer in our domain, another MPI rank will pick up later.
                      if(fsgridCell[0] == -1) {
                         cellNeedsContinuedFWTracing[n] = 1;
                         cellFWTracingCoordinates[n] = x;
                         break;
                      }
-                  }
+                  } // while true
                } // if FW
                if(cellNeedsContinuedBWTracing[n]) {
```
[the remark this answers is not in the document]
ykempf: yup

fieldtracing/fieldtracing.cpp
ykempf: There was a version of the code where that could be more than 1. I ran a quick test that confirmed this isn't the case anymore.

fieldtracing/fieldtracing.cpp
markusbattarbee: These are still deep copies?

ykempf:
So now I was able to push out of LUMI. Are these comments better, Markus?

fieldtracing/fieldtracing.cpp
```diff
@@ -763,392 +760,191 @@ namespace FieldTracing {
...
+    * this cell is near a fluxrope. This means we store the value of cellMaxExtension[n] into CellParams::FLUXROPE. Otherwise we
+    * store zero.
+    * After that we apply % TracingLineEndType::N_TYPES to recover values UNPROCESSED, CLOSED, OPEN, DANGLING in
+    * cellFWConnection[n] and cellBWConnection[n] so that we can assign the connection types for the full box connection described
+    * above.
+    *
+    * As a freebie since we computed the curvature anyway for flux rope tracing we store that into CellParams::CURVATUREX/Y/Z.
+    *
+    * \sa stepCellAcrossTaskDomain
     */
```
ykempf, line 940: How do you like this explanation attempt?

fieldtracing/fieldtracing.cpp
```diff
@@ -763,392 +760,191 @@ namespace FieldTracing {
...
+    * We are doing two things in one function here as we save a *lot* especially in MPI reductions by combining those. And they are
+    * doing very similar things anyway.
+    *
+    * Firstly we are interested in knowing the ultimate magnetic connection forward and backward from any given location. The
+    * locations are the DCCRG cell centers, there's too many fsgrid cells for practical purposes.
+    * Therefore we step along the magnetic field until we hit an outer box wall (OPEN) or the inner boundary
+    * radius (CLOSED). In rare cases the line might not have reached either of these two termination conditions by the time the
+    * field line has reached a length of maxTracingDistance. In that case we call it DANGLING, it likely ended up in a loop
+    * somewhere (we don't call it "loop" to avoid confusion with the flux rop tracing). As long as we have not hit any of the above
+    * termination conditions, the type is called UNPROCESSED. We allow fieldTracingParameters.fullbox_max_incomplete_lines to remain
```
kstppd, line 883: So this is a bit confusing. How can we not hit one of OPEN,CLOSED or DANGLING  criteria?

fieldtracing/fieldtracing.cpp
```diff
@@ -763,392 +760,191 @@ namespace FieldTracing {
...
+    * for sanity) are near/in a flux rope. As this relies also on tracing forward and backward along the field, we piggyback on the
+    * full box connection algorithm explained above.
+    * We trace along the field up to fieldTracingParameters.fluxrope_max_curvature_radii_to_trace*cellCurvatureRadius[n] and if
+    * within that tracing distance we have't extended further than
+    * fieldTracingParameters.fluxrope_max_curvature_radii_extent*cellCurvatureRadius[n] we are rolled up tightly enough to consider
+    * being close to a flux rope. We'll store the max extension reached into cellMaxExtension[n] for later fine-grained analysis.
+    * The arithmetic idea to avoid using yet more arrays:
+    * We use the cellFWConnection[n]/cellBWConnection[n]/cellConnection[n] (that can only be UNPROCESSED, CLOSED, OPEN or DANGLING,
+    * see TracingLineEndType enum above, ending with N_TYPES). As long as no full box tracing termination condition was reached
+    * we are at UNPROCESSED. If we exceed fieldTracingParameters.fluxrope_max_curvature_radii_to_trace*cellCurvatureRadius[n] OR
```
kstppd, line 921: could we also be dangling here?

fieldtracing/fieldtracing.cpp
kstppd: Mostly understandable but I do have 2 comments which I left above and got a bit confused there

fieldtracing/fieldtracing.cpp
```diff
@@ -763,392 +760,191 @@ namespace FieldTracing {
...
+    *
+    * We are doing two things in one function here as we save a *lot* especially in MPI reductions by combining those. And they are
+    * doing very similar things anyway.
+    *
+    * Firstly we are interested in knowing the ultimate magnetic connection forward and backward from any given location. The
+    * locations are the DCCRG cell centers, there's too many fsgrid cells for practical purposes.
+    * Therefore we step along the magnetic field until we hit an outer box wall (OPEN) or the inner boundary
+    * radius (CLOSED). In rare cases the line might not have reached either of these two termination conditions by the time the
+    * field line has reached a length of maxTracingDistance. In that case we call it DANGLING, it likely ended up in a loop
+    * somewhere (we don't call it "loop" to avoid confusion with the flux rop tracing). As long as we have not hit any of the above
```
kstppd, line 882: rope*

kstppd (approved):

fieldtracing/fieldtracing.cpp
ykempf: This is what 883 sqq. is explaining: with the allowed leftover fraction.

fieldtracing/fieldtracing.cpp
ykempf: What it says on 919? Yes, but very unlikely, `maxTracingDistance` is bigger than `fieldTracingParameters.fluxrope_max_m_to_trace`. See also the check around line 1023.

markusbattarbee: Should there be a check on parameter reading to ensure `maxTracingDistance` is at least as big as `fieldTracingParameters.fluxrope_max_m_to_trace`?

fieldtracing/fieldtracing.cpp
```diff
@@ -1182,34 +1183,28 @@ namespace FieldTracing {
...
-               cellBWTracingStepSize.swap(reducedCellBWTracingStepSize);
-               cellFWRunningDistance.swap(reducedCellFWRunningDistance);
-               cellBWRunningDistance.swap(reducedCellBWRunningDistance);
-               cellFWConnection.swap(reducedCellFWConnection);
-               cellBWConnection.swap(reducedCellBWConnection);
-               // No swap for these as the old ones are used for the small arrays ~70 lines up.
-               cellNeedsContinuedFWTracing = reducedCellNeedsContinuedFWTracing;
-               cellNeedsContinuedBWTracing = reducedCellNeedsContinuedBWTracing;
+               // These are used for the small arrays ~70 lines up.
+               reducedCellNeedsContinuedFWTracing = cellNeedsContinuedFWTracing;
```
markusbattarbee, line 1187: if reductions are only done to the "small" vectors, then should these in fact be renamed to `previousRoundCellNeedsContinuedXWTracing`? or `stored...`?

fieldtracing/fieldtracing.cpp
```diff
@@ -1182,34 +1183,28 @@ namespace FieldTracing {
...
             }
             #pragma omp for schedule(dynamic) reduction(+:cellsToDoFullBox) reduction(+:cellsToDoFluxRopes)
             for(int n=0; n<globalDccrgSize; n++) {
                if(cellNeedsContinuedFWTracing[n]) {
                   cellsToDoFullBox++;
                   // Update that nodes' tracing coordinates
-                  cellFWTracingCoordinates[n][0] = sumCellFWTracingCoordinates[n][0];
-                  cellFWTracingCoordinates[n][1] = sumCellFWTracingCoordinates[n][1];
-                  cellFWTracingCoordinates[n][2] = sumCellFWTracingCoordinates[n][2];
+                  cellFWTracingCoordinates[n][0] = cellFWTracingCoordinates[n][0];
```
markusbattarbee, line 1195: these lines now do nothing?  same at 1205.

fieldtracing/fieldtracing.cpp
```diff
@@ -1252,46 +1247,47 @@ namespace FieldTracing {
             // Handle flux ropes
             mpiGrid[id]->parameters[CellParams::FLUXROPE] = 0;
             const std::array<Real, 3> x = mpiGrid.get_center(id);
-            // Earlier, if we marked nothing (e.g. hit a wall or ionosphere before making a call) reducedCellXWConnection[n] is less than N_TYPES.
+            // We use cellXWConnection as we swapped with reducedCellXWConnection.
+            // Earlier, if we marked nothing (e.g. hit a wall or ionosphere before making a call) cellXWConnection[n] is less than N_TYPES.
```
markusbattarbee, line 1251: I'll try to look in more detail at this soon, but I think one needs to be careful now with doing swaps and then going back and forth. :)

markusbattarbee (commented):
Still need to read the actual commentary on the bus, but some comments here

ykempf:
Implemented your suggestions.

fieldtracing/fieldtracing.cpp
```diff
@@ -1221,287 +1027,271 @@ namespace FieldTracing {
...
-                  anyCellNeedsTracing=true;
-                  cellNeedsContinuedFWTracing[n] = 1;
-                  
-                  // Update that nodes' tracing coordinates
-                  cellFWTracingCoordinates[n][0] = sumCellFWTracingCoordinates[n][0] / reducedCellNeedsContinuedFWTracing[n];
-                  cellFWTracingCoordinates[n][1] = sumCellFWTracingCoordinates[n][1] / reducedCellNeedsContinuedFWTracing[n];
-                  cellFWTracingCoordinates[n][2] = sumCellFWTracingCoordinates[n][2] / reducedCellNeedsContinuedFWTracing[n];
-                  
-                  cellFWRunningDistance[n] = reducedCellFWRunningDistance[n];
+               if(cellNeedsContinuedFWTracing[n]) {
```
markusbattarbee, line 1199: perhaps a nitpick, but this actually doesn't now consider how many cells are unprocessed, but rather, how many half-traces (FW or BW) are still unprocessed.

 So perhaps it should rather check for each cell if
`cellNeedsContinuedBWTracing[n]!=0` and `cellNeedsContinuedFWTracing[n]!=0` for `cellsToDoFullBox++` and inside that if:
 `cellBWConnection[n] == cellFWConnection[n]== TracingLineEndType::UNPROCESSED`
for `cellsToDoFluxRopes++`?

fieldtracing/fieldtracing.cpp
```diff
@@ -763,392 +760,191 @@ namespace FieldTracing {
...
-         cellFWTracingCoordinates.at(n) = mpiGrid.get_center(id);
-         cellBWTracingCoordinates.at(n) = cellFWTracingCoordinates.at(n);
+      std::array<Real, 3> x = cellTracingCoordinates[n];
+      std::array<Real, 3> v({0,0,0});
+      while( true ) {
+         // Check if the current coordinates (pre-step) are in our own domain.
+         std::array<int, 3> fsgridCell = getLocalFsGridCellIndexForCoord(technicalGrid,x);
+         // If it is not in our domain, somebody else takes care of it.
+         if(fsgridCell[0] == -1) {
+            cellNeedsContinuedTracing[n] = 0;
```
markusbattarbee, line 792: Instead of the "needs continued tracing" vectors, you could use (cellXWConnection % TracingLineEndType::N_TYPES == 0) as a check if the line should still continue? In that case, at this check, if the cell is not local then also the connection type should be set to zero. (Would save some MPI reductions)

(You might then want to set anything near boundaries as OPEN_OPEN) or something so they get put aside straight away)

parameters.cpp
```diff
@@ -441,7 +441,8 @@ bool P::addParameters() {
    RP::add("fieldtracing.tracer_max_allowed_error", "Maximum allowed error for the adaptive field line tracers ", 1000);
    RP::add("fieldtracing.tracer_max_attempts", "Maximum allowed attempts for the adaptive field line tracers", 100);
    RP::add("fieldtracing.tracer_min_dx", "Minimum allowed field line tracer step length for the adaptive field line tracers (m)", 100e3);
-   RP::add("fieldtracing.tracer_max_incomplete_fieldlines_fullbox", "Maximum fraction of field lines left incomplete when stopping tracing loop. Defaults to zero to process all, can be slow at scale!", 0);
+   RP::add("fieldtracing.fullbox_max_incomplete_fieldlines", "Maximum fraction of field lines left incomplete when stopping tracing loop. Defaults to zero to process all, will be slow at scale!", 0);
+   RP::add("fieldtracing.fluxrope_max_incomplete_fieldlines", "Maximum fraction of field lines left incomplete when stopping tracing loop. Defaults to zero to process all, will be slow at scale!", 0);
```
markusbattarbee, line 445: These comments should be clarified, now they both say the same thing :)

parameters.cpp
```diff
@@ -441,7 +441,8 @@ bool P::addParameters() {
...
    RP::add("fieldtracing.use_reconstruction_cache", "Use the cache to store reconstruction coefficients. (0: don't, 1: use)", 0);
    RP::add("fieldtracing.fluxrope_max_curvature_radii_to_trace", "Maximum number of seedpoint curvature radii to trace forward and backward from each DCCRG cell to find flux ropes", 10);
```
markusbattarbee, line 447: suggestion: "from each DCCRG cell to find flux ropes" -> "from each DCCRG cell when attempting to identify flux ropes"

markusbattarbee (changes requested):
The description is a lot better now - but some more suggestions for optimization are here.

I was also wondering, do we actually need double-precision values here, or would we get good-enough results with float? That would reduce MPI comms by a lot. Perhaps worth trying. One could even typedef a RealFT for this purpose.

One aspect I still find confusing is the interplay of the various maximum allowed tracing distances. There's two values for flux ropes (one in m, one in curvature radii) and one for all tracing. Inside the "first-loop" (which should be initialization loop, IMO) there's one check which appears to be a heuristic to discard cells which start with large curvature radii? A comment there seems to be outdated? In general, I feel I'm somewhat questioning the use of `fluxrope_max_m_to_trace` anymore.

fieldtracing/fieldtracing.cpp
```diff
@@ -1221,287 +1020,279 @@ namespace FieldTracing {
...
+                     smallCellBWTracingCoordinates.push_back(cellBWTracingCoordinates[n]);
+                     smallCellBWRunningDistance.push_back(cellBWRunningDistance[n]);
+                     smallCellBWTracingStepSize.push_back(cellBWTracingStepSize[n]);
+                     smallCellBWConnection.push_back(cellBWConnection[n]);
+                  }
+               }
+               int smallSizeFW = indicesToReduceFW.size();
+               int smallSizeBW = indicesToReduceBW.size();
+               
+               std::vector<signed char> smallReducedCellNeedsContinuedFWTracing(smallSizeFW), smallReducedCellNeedsContinuedBWTracing(smallSizeBW);
```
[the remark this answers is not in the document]
markusbattarbee: (comments came in the wrong order, hence the "Similarly")

fieldtracing/fieldtracing.cpp
```diff
@@ -763,392 +760,191 @@ namespace FieldTracing {
...
-         #pragma omp for schedule(dynamic)
-         for(int n=0; n<globalDccrgSize; n++) {
-            if(cellNeedsContinuedFWTracing[n] == 1) {
-               cellFWConnection[n] = TracingLineEndType::LOOP;
-            }
-            if(cellNeedsContinuedBWTracing[n] == 1) {
-               cellBWConnection[n] = TracingLineEndType::LOOP;
+         // See the longer comment for the function traceFullBoxConnectionAndFluxRopes for details.
+         // If we are still in the race for flux rope...
+         if(cellConnection[n] < TracingLineEndType::N_TYPES) {
```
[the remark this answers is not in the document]
ykempf: I changed the other one but to keep things readishable I'd rather keep this one as the actual extension. Indeed I don't think this dominates. Unless you really want me to do that one.

fieldtracing/fieldtracing.cpp
ykempf: OK the naming is confusing, but so far the fractions were of traces (2 per cell) and not cells. Should we make it thresholds of fractions of cells that have come to a valid result instead?

ykempf: Or rename the variables to tracesToDo...?

ykempf:
So much for "I'll quickly code that before fetching my morning coffee".

ursg:
Code looks good. Results from today's meeting look good. Everything good. Let's push this button here!

Issue #688, opened 2023-01-16
vlsvextract doesn't look for modern b field names in bulk files


ykempf:
The list of magnetic field variable name combinations vlsvextract goes through when using `--rotate` doesn't include e.g. `vg_b_vol`, which it should these days.

PR #689, opened 2023-01-17
Added support for modern vg_ and fg_ variables in vlsvextract for rotation and plasma frame

ykempf:
What it says on the can. A fix for #688.
Querying fg_b would require quite a bit more code (parsing fsgrid etc). Though now that I think of it, we can't rotate or shift for restarts with this either.

### DIFF
--- a/tools/vlsvextract.cpp
+++ b/tools/vlsvextract.cpp
@@ -492,31 +492,44 @@ void getBulkVelocity(Real* V_bulk,vlsvinterface::Reader& vlsvReader,const string
       exit(1);
    }
    
-   // Read number density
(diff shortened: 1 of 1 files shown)


ykempf:
Fixed hard-coded proton to use popName for plasma frame shift.

ykempf:
So I tested this with an EGI bulk and an EGI restart, works. The only thing that might be confusing to the user is that it prints a VLSV failed message for each unsuccessful read attempt. But in case of success it prints that too and if all fail there is a proper error printed followed by `exit(1)` so I guess we're OK with this now?

PR #690, opened 2023-01-25
Fix compilation on platforms that have non-working openmp

ursg:
This PR mostly removes unnecessary includes of <omp.h> (or puts them behind ```#define _OPENMP```), and provides alternative codepaths for any part of the code that calles omp helper functions directly.


ursg:
Confirmed to not break anything about the building with openmp (which is the important part), and also to compile successfully on a weird architecture with openmp disabled.

PR #691, opened 2023-01-29
A possible fix: Receiving sibling cells sysboundary check

alhom:
Consistent handling of boundary cells on the receiving rank: check if the sibling cells looped over are also `NOT_SYSBOUNDARY`.


markusbattarbee:
Looks sane. And great that it works!
And since they are siblings of a local cell, scell should always be a valid pointer. Still, I'd perhaps take this opportunity to add a check for that as well?

alhom:
Good point - and this is checked on the sending side, along with `nbr != INVALID_CELLID`, so no reason to not be consistent here as well.

alhom:
NB: Also added the testpackage/debug modification of initializing aligned_malloc with nans in https://github.com/fmihpc/vlasiator/pull/691/commits/7c2c6e0c6a408cbb8c244506277bcbf49f55f73d.

ursg (approved):
Yup, I like!

PR #696, opened 2023-01-31
Obsolete conducting option, in add but not get, removed now.

ykempf:
duh


ursg:
Right, these are much more logically located in the Magnetosphere project now, and somehow were still lingering leftovers in the conducting sphere.
Good to remove.

PR #703, opened 2023-02-20
Write fsgrid metadata anyway

ykempf:


ykempf:
Is that good @markusbattarbee?

ykempf:
Merging this, so we have it in the next slot of FHA. #whatcouldpossiblygowrong

PR #704, opened 2023-02-21
Fix reduction for open-closed tracing.

ykempf:
I changed the UNPROCESSED and flipped the reduction for full box + fluxrope the other day, but not in here. So silly.


ursg (approved):
"lol."

PR #676, opened 2022-11-23
Fix carrington and ukko makefile for compilation of dAMR

ursg:
also includes an updated carrington testpackage run script.

PR #705, opened 2023-02-24
Automatic compilation and testpackage run (for Continuous Integration).

ursg:
Github actions automatically get triggered on every PR and commit for the master and dev branches. Basic compilation tests are run on github hosted runners (using a docker VM), testpackage is built and run on the Carrington cluster.

Any newly produced warnings or errors as part of the PR code are highlighted in the PR diff view. Testpackage results are available as a PR check (see "details" for the whole output).

### DIFF
--- a/.github/workflows/github-ci.yml
+++ b/.github/workflows/github-ci.yml
@@ -183,14 +183,14 @@ jobs:
         sbatch -W -o testpackage_run_output.txt ./small_test_carrington_github_ci.sh
         cat testpackage_run_output.txt
         cat $GITHUB_STEP_SUMMARY > $GITHUB_WORKSPACE/testpackage_check_description.txt
+        cd $GITHUB_WORKSPACE
+        tar -czvf testpackage-output.tar.gz testpackage_check_description.txt testpackage_output_variables.txt
     - name: Upload testpackage output
       uses: actions/upload-artifact@v3
       if: always()
       with:
         name: testpackage-output
-        path: |
-          testpackage_check_description.txt
-          testpackage_output_variables.txt
+        path: testpackage-output.tar.gz
       # Note: Testpackage output is further processed in the pr_report.yml workflow
       # (to produce Checks against pull requests)
 
--- a/.github/workflows/github-ci.yml
+++ b/.github/workflows/github-ci.yml
@@ -109,7 +109,9 @@ jobs:
       run: tar --zstd -xvf libraries.tar.zstd
     - uses: ammaraskar/gcc-problem-matcher@master
     - name: Compile vlasiator (Release build)
-      run: VLASIATOR_ARCH=github_actions make -j 3
+      run: |
+        VLASIATOR_ARCH=github_actions make clean
+        VLASIATOR_ARCH=github_actions make -j 3
       
   
   build_testpackage:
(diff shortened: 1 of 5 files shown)


ursg:
I do suggest merging #676 before this, so that the testpackage runs on carrington don't take forever.

ursg:
> reference revision should perhaps be vlasiator_CI_reference_current or something similar, so that it's clear in the Vakka storage which one is being used.

True. Also, I guess it would make sense to again create new reference data.

> Should the build_tools and build_production tasks also include make clean before making? I think so. Good thing boost is available on the github runners... :)

They are being bootstrapped on a "clean" docker image every time anyway, but I suppose it is good practice to run that anyway, just in case it ever changes in the future.

> I saw some references to using scripts like Run ammaraskar/gcc-problem-matcher@master Run LouisBrunner/checks-action@v1.6.0
> 
> How safe is this, sourcing scripts from outsiders? And should we link to an explicit commit from that script, instead of e.g. @master?

Hmm, well. true. But since these are just names of github repositories that their code is pulled from, we could have our own personal forks of them and run with those.

ursg:
> Good thing boost is available on the github runners... :)

The thing they are running on is actually a [custom docker container](https://hub.docker.com/r/ursg/vlasiator_ci) that I buit. It can contain anything we want, so in principle we could also just bake the libraries and reference data in there.

But I think it is nice to continuously check whether the whole build process including dependencies works well.

ursg:
So, this appears to work now. The PR triggers a workflow which builds the code (both on github runners and on carrington). The testpackage runs on carrington, and produces a .tar.gz outputfile. But that workflow does not have write access to the PR for security reasons, because it runs in the context of the branch that the PR originates from.

Completion of that PR however triggers a second workflow which runs in the context of the master branch of the target repository. That ones fetches the testpackage artifact, unpacks it and creates the PR annotation from it. [This](https://github.com/ursg/vlasiator/runs/11625564743) is what it looks like, for example.

Now, in this PR, it is of course not visible yet, because it isn't in fmihpc/master *yet*.

.github/workflows/github-ci.yml
```diff
@@ -109,7 +109,9 @@ jobs:
       run: tar --zstd -xvf libraries.tar.zstd
     - uses: ammaraskar/gcc-problem-matcher@master
     - name: Compile vlasiator (Release build)
-      run: VLASIATOR_ARCH=github_actions make -j 3
+      run: |
+        VLASIATOR_ARCH=github_actions make clean
```
markusbattarbee, line 106: should this make clean be as its own `name` so it shows in the log? (like was added for build_tools)

markusbattarbee (commented):
So looking at 
https://github.com/ursg/vlasiator/pull/60/checks?check_run_id=11554313975
I guess this is what it should look like for PRs in the future?

For some testpackage tests, there are printouts of `stderr: 
(standard_in) 1: syntax error`

Also, the speedup value isn't showing up

ursg:
> So looking at https://github.com/ursg/vlasiator/pull/60/checks?check_run_id=11554313975 I guess this is what it should look like for PRs in the future?
> 
> For some testpackage tests, there are printouts of `stderr: (standard_in) 1: syntax error`
> 
> Also, the speedup value isn't showing up

Ah, I fixed both of those, but apparently haven't done another clean run on my repo after that. So let me repeat that one, just to be sure.

ursg:
There, updated.

ursg:
Since this is not "in the way" of anything, I'm just going to merge it. All we see are a couple of checkmarks on the PRs!

For the testpackage report to work, we'll anyway have to wait until this trickles down to master. (Or do a separate, out-of-order PR of the report.yml file onto there.)

PR #707, opened 2023-03-01
Add github action to report automatic testpackage runs to PR checks

ursg:
This is an out-of-order PR directly onto master, bypassing the regular dev staging.

For security reasons, Github Actions wants this workflow file in the master branch of the repository, and will not otherwise run it. So to get the testpackage result as a nice good/neutral/bad checkmark on pull requests, this file needs to go into master.

(The alternative would be to wait until our next merge-dev-to-master-and-release, which might also be soon)

### DIFF
--- a/.github/workflows/pr_report.yml
+++ b/.github/workflows/pr_report.yml
@@ -1,0 +1,45 @@
+name: Report
+on:
+  workflow_run:
+    workflows: [Github-CI]
+    types: [completed]
+
+permissions:
+  checks: write
+
+jobs:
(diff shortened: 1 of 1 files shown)


markusbattarbee (commented):
It's a valuable addition to our development pipeline and by virtue of being just a github runner file, won't affect the actual codebase as such. My only comments would be in case we'd want to make local safe and stable copies of those runners fron dawidd6 and LouisBrunner...?

markusbattarbee (approved):

PR #709, opened 2023-03-02
visit .xml for plotting

kostahoraites:
created visit variables_3d.xml file, based on a similar file variables_nongyrotropy.xml. Useful template for getting started with VisIt plotting.


kostahoraites:
Also should mention: the old .xml files were out of date and referenced obsolete variable names

PR #710, opened 2023-03-04
Merge ionosphere gpu solver to dev 

MiroPalmu:
Compilation instructions are contained in readme.md but long story short:

while compiling with `IONO_GPU` environment variable set to 1 and linking with cuda runtime `SphericalTriGrid::solve` will use new solver.

Most of the files in this PR are tests for the solver.


markusbattarbee:
Can you please make this point to cudasiator instead of dev?
Also, it should be brought up-to-date with dev in order to include e.g. https://github.com/fmihpc/vlasiator/pull/668/files

PR #711, opened 2023-03-07
Write out full bgb data

ykempf:
Open for comments, seems to work as expected on my laptop. Quite a bit of duplicate code, I'm open for other ways of writing this...

In a second step we can discuss/implement a gridGlue patch that would do the vg versions of all these as well.


ykempf:
Question: do we write them out as differences as is the case now and as is done for all other such DROs, or do we change a) only these BGB guys b) all "derivatives" to be derivatives instead of differences?

markusbattarbee:
I am of the thought that it would be beneficial to actually switch over all derivatives in vlasiator to be proper derivatives - or, alternatively, rename the variables to differences.

ykempf:
So do I add making all relevant DROs actual derivatives to this PR or shall I make a new one after merging this one?

markusbattarbee:
One thing at a time, I think :) But if that's what we want to do next, then indeed let's make this conform already to that style.

ykempf:
Uh, I was only suggesting to make the DROs derivatives, not internally.

ykempf:
Are these first commits going as far back as 2020 just github flukes? Or did I organise a git wreck again?

markusbattarbee:
It shows just 10 commits to me.

ykempf:
Yes but the very first is from the stone age, and the later "merge dev into ykempf/dev" aren't particularly fresh either.

ykempf:
All right hewre is the "lost" PR, I searched for "background" (or actually "ackgr") and didn't find it as it says bgb here...

ykempf:
Merged into #786

PR #712, opened 2023-03-08
Add fieldtracing to allowed groups in check_vlasiator_cfg.sh

ykempf:


ursg:
An automatic CI test whether all DRO names occur in the help test might be nice for PR checks in the future.

ykempf:
But this is a full new group, not a DRO. And DRO means you'd have to grep through and parse all the strings in the ifs in datareducer.cpp.

ykempf:
At least I don't see how else with the current mechanism to get an up to date list of DROs beyond doing manually like I did for Christmas. And that was a PITK.

PR #713, opened 2023-03-08
Dev no double phiprof calls

ykempf:
Found one mistake which crashed TAU, one forgotten renamed stop call, and removed three using namespace phiprof lines that were useless.


ykempf:
Meh and this has the other PR's commit on top. Sorry.

ursg:
Do you want to still rebase this to get rid of the #712 commit in here? Or should I just merge both?

ursg:
I'll just merge both!

PR #716, opened 2023-03-09
Add phiprof::stop to early return calls.

ykempf:


ursg:
Wow, that's actually a lot of them! Thumbs up!

ykempf:
Hmm until other spots bite us when TAUing – should someone go through systematically? Can we make that automatic? Probably not?

ursg:
#721 of makes the type system take care of this for us. So maybe no need to manually go through everything anymore?

ykempf:
True. Great!

PR #718, opened 2023-03-11
Failure of the testpackage in CI should set red check marks.

ursg:
This modifies the main CI workflow to fail the testpackage step if any test fails.

Also, the report step limits the uploaded report size to 60kb.


markusbattarbee (approved):

ursg:
Hum, I tested this in #719, and [it did not give a red check mark](https://github.com/fmihpc/vlasiator/actions/runs/4402710601/jobs/7710223866), so I guess this needs some fixing still.

ursg:
Ah, it's because [the actual run command runs in a subshell](https://github.com/fmihpc/vlasiator/blob/5af965727c1cfc6d5262826717b10a2a01b215dc/testpackage/small_test_carrington_github_ci.sh#L121-L128) and so the ```$?``` variable in the script does *not* actually capture the previous command's return value.

ursg:
Alright, now #719 fails in the correct way, and this here succeeds in the correct way.

PR #720, opened 2023-03-14
LUMI Cray Makefile and one fix

ykempf:


ursg:
So, is this... good to go?

ursg:
Can confirm this fix is also required for building with clang. So merging.

PR #721, opened 2023-03-14
Timer object

lkotipal:
Replace all phiprof timer calls with timer objects as per https://github.com/fmihpc/phiprof/pull/30. Timers are closed by their destructor e.g. when going out of scope, making it easier to avoid bugs caused by dangling timers.

Some guidelines:
- Never use `phiprof::start(args)` or `phiprof::stop(label, args)`, use `phiprof::Timer t {args}` and `t.stop(args)` instead.
- Use `int id = phiprof::initialize(args)` outside a loop and `phiprof::Timer {id}` inside. Initialization is superfluous if done immediately before starting a timer.
- Since timers automatically stop when going out of scope, you may omit `t.stop()` before e.g. return.

### DIFF
--- a/vlasovsolver/vlasovmover.cpp
+++ b/vlasovsolver/vlasovmover.cpp
@@ -77,7 +77,6 @@ void calculateSpatialTranslation(
         Real &time
 ) {
 
-    int trans_timer;
     bool localTargetGridGenerated = false;
     bool AMRtranslationActive = false;
     if (P::amrMaxSpatialRefLevel > 0) AMRtranslationActive = true;
@@ -94,12 +93,12 @@ void calculateSpatialTranslation(
     // ------------- SLICE - map dist function in Z --------------- //
    if(P::zcells_ini > 1){
 
-      phiprof::Timer transTimer_timer {"transfer-stencil-data-z", {"MPI"}};
+      phiprof::Timer transTimer {"transfer-stencil-data-z", {"MPI"}};
       //updateRemoteVelocityBlockLists(mpiGrid,popID,VLASOV_SOLVER_Z_NEIGHBORHOOD_ID);
       SpatialCell::set_mpi_transfer_direction(2);
       SpatialCell::set_mpi_transfer_type(Transfer::VEL_BLOCK_DATA,false,AMRtranslationActive);
       mpiGrid.update_copies_of_remote_neighbors(VLASOV_SOLVER_Z_NEIGHBORHOOD_ID);
-      trans_timer.stop();
+      transTimer.stop();
 
       // bt=phiprof::initializeTimer("barrier-trans-pre-trans_map_1d-z","Barriers","MPI");
       // phiprof::start(bt);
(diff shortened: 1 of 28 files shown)


lkotipal:
aaa

ursg:
This is nice (and apparently works well), but I would suggest that the timers don't get variable names like "mpi", "communication" etc.
Because in its current form, stopping these timers then adds code lines like ```mpi.stop()``` or ```communication.stop()``` that, by themselves, make it in no way apparent that they are dealing with timers.

So either rename all of them to ```mpiTimer```,```communicationTimer``` etc, or change the name of the stop method to ```stopTimer```.

lkotipal:
Normally I'd say the type conveys that information sufficiently but IntelliSense is a bit finicky over remote connections. I'll go through and give the timers better names.

lkotipal:
Yolo'd this in regex, let's see if it compiles

lkotipal:
Automatic testpackage output for pull request #721.

Based on commit 6e9c9be62d160cbecf8dd64189a5748883512849.

Process started by user ```lkotipal``` on host ```turso01.cs.helsinki.fi``` at Mon, 11 Sep 2023 14:59:52 +0300.

:heavy_check_mark: Compilation succeeded with 23 warnings.

:heavy_check_mark: restart_read: Zero diffs.
:large_orange_diamond: Magnetosphere_small: Nonzero diffs: `proton/vg_v_0` has absolute error 32255.9, relative error 0.0639677.
:large_orange_diamond: Magnetosphere_polar_small: Nonzero diffs: `proton/vg_v_nonthermal_0` has absolute error 102622, relative error 0.128336.
:large_orange_diamond: Flowthrough_amr: Nonzero diffs: `proton/vg_rho_0` has absolute error 0.00289569, relative error 1.34239e-09.
:large_orange_diamond: Magnetosphere_3D_small: Nonzero diffs: `proton/vg_v_0` has absolute error 262.342, relative error 0.000349676.
:heavy_check_mark: Ionosphere_small: Zero diffs.

Full testpackage run output is available in `/wrk-vakka/users/lkotipal/testpackage/testpackage_run_6e9c9be62d160cbecf8dd64189a5748883512849.log` on host `turso01.cs.helsinki.fi`

lkotipal:
Turso is having a normal one so here's a manual test

lkotipal:
You're right, my references are probably outdated. I'll generate new ones and re-run.

vlasovsolver/vlasovmover.cpp
```diff
@@ -455,73 +437,70 @@ void calculateAcceleration(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
       for (uint popID=0; popID<getObjectWrapper().particleSpecies.size(); ++popID) {
          adjustVelocityBlocks(mpiGrid, cells, true, popID);
       }
+   } else {
+      // Fairly ugly but no goto
```
ursg, line 441: This part is the only actual change of code semantics, other than phiprof variable reordering.
The testpackage result shows that this is working all as intended, but was this change required for the timer objects, or was it just something that irked you and you wanted to go away?

ursg (approved):
Great stuff! Scrolled through the whole diff just to be sure.
I also approve the occasional formatting fix that you put in.

Your initializer syntax of
```
int someTime {phiprof::initializeTimer("asdf")};
```
made me do a double-take in some places, but that's all ok!

vlasovsolver/vlasovmover.cpp
```diff
@@ -272,9 +256,11 @@ void calculateSpatialTranslation(
    
    // If dt=0 we are either initializing or distribution functions are not translated. 
    // In both cases go to the end of this function and calculate the moments.
-   if (dt == 0.0) goto momentCalculation;
+   if (dt == 0.0) {
+      calculateMoments_R(mpiGrid,localCells,true);
+   }
```
markusbattarbee, line 261: I would add a return here so we don't do a pointless zero-dt translation computation. Then it'll return to the same behaviour we had before.

lkotipal: That was the intention, I think I goofed up the merge there

vlasovsolver/vlasovmover.cpp
lkotipal: iirc that actually didn't compile with the goto since it was jumping over a variable declaration

ursg: Fair enough, that's definitely a sufficient reason to restructure.

PR #722, opened 2023-03-15
Verify restart

alhom:
Drafted a check against corrupt/unfinished restart VLSV files. Should probably go to vlsv lib instead.

Reads 16 last bytes of the vlsv, and checks if the end tag exists in there. If not, there's a good chance the restart file is corrupt. At least, this exits the run in case of a bad file.


alhom:
Heh, draft pull requests trigger the CI agent as well. This will fail on restart_read, though, since there is a hard exit...

grid.cpp
```diff
@@ -224,7 +232,7 @@ void initializeGrids(
       phiprof::start("Read restart");
       if (readGrid(mpiGrid,perBGrid,EGrid,technicalGrid,P::restartFileName) == false) {
          logFile << "(MAIN) ERROR: restarting failed" << endl;
-         exit(1);
+         exit(ExitCodes::RESTART_READ_FAILURE); // Do these have automatically recoverable failure modes?
```
ykempf, line 235: Uhm, like what?

alhom: ... like picking the previous restart file. If the restart file is bust, that would be the recovery, but do we even want to do that automatically?

ykempf: I think that's the kind of think better handled in the job script upon parsing the return code.

alhom: Yup, well, that's why it is here anyway.

alhom:
Rebased to dev, Errorcode definitions and printout works - needs to be hooked to help message.

PR #724, opened 2023-03-17
Pass MPI-IO hints to restart reading.

ykempf:


ioread.cpp
```diff
@@ -1070,9 +1070,22 @@ bool exec_readGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
...
-   MPI_Info mpiInfo = MPI_INFO_NULL;
 
-   if (file.open(name,MPI_COMM_WORLD,MASTER_RANK,mpiInfo) == false) {
+   MPI_Info MPIinfo;
+   if (P::restartReadHints.size() == 0) {
+      MPIinfo = MPI_INFO_NULL;
+   } else {
+      MPI_Info_create(&MPIinfo);
+      
+      for (std::vector<std::pair<std::string,std::string>>::const_iterator it = P::restartReadHints.begin();
```
ursg, line 1080: This would be a prime example for a loop, that would be much more readable as a range-based loop, like
```
for (i : P::restartReadHints) {
   MPI_Info_set(MPIinfo, i.first.c_str(), i.second.c_str());
}
```

ursg (approved):
Looks good (and the performance speaks for itself!).

Just one minor stylistic comment.

ioread.cpp
ykempf: It's copied from the other two such instances. And I hate those `auto`s, must be getting old'n'grumpy. But if you so wish, I shall change the three instances.

ursg: Nah, it's fine.

PR #725, opened 2023-03-21
Added two openmp for loops in LB weight assignment.

ykempf:
I want to see what the testpackage makes of this. CI FTV.

### DIFF
--- a/grid.cpp
+++ b/grid.cpp
@@ -297,6 +297,7 @@ void initializeGrids(
       }
       phiprof::stop("Apply system boundary conditions state");
       
+      #pragma omp parallel for schedule(static)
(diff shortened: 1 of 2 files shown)


ykempf:
So was speedup supposed to be <1 or >1?

lkotipal:
Speedup > 1 is faster. The CI testpackage seems unreliable for measuring speedup, however.

markusbattarbee:
I wouldn't expect any actual benefit here, but indeed why not.

lkotipal:
Might need a more robust performance measurement to be helpful. It's fairly straightforward to run the testpackage through `tau_exec`. Maybe we should look into using tau tools to automate the processing of performance data?

ursg:
Yup.

PR #726, opened 2023-03-21
Cudasiator unified ongoing

markusbattarbee:
Acceleration has been fully ported to GPU, including block adjustments. 

Single-thread/stream speed is >10x faster than singe-thread CPU version (when going to WID8 but maintaining v-space resolution).

Requires this branch from hashinator:
https://github.com/kstppd/hashinator/tree/stream_attach

When going to 32 threads/streams, comparing against 32 CPU threads, speed improvement is lost - reason is yet unclear, possibly related to lack of `cudaMemAdviseSetPreferredLocation` calls or overhead from stream attaches. Also, the hashinator stream attaching is not yet working and may be a culprit.

Hipifying / converting to ROCm has not been tested yet on this version.


markusbattarbee:
Brought it up to date with dev. 

Fixed issues with the hashinator stream attach - required changes to hashinator repo are yet to be committed. Speed actually seemingly went down. Further analysis required.

Other improvements to be tested:
* move all, even miniscule cudaMemcpy calls, to use pinned host memory.
* Implement a second higher-priority stream for the few serial kernels
* optimize clear / re-allocation of vectors
* move blockdata and parameters back to non-unified memory

markusbattarbee:
Updated, now requires (as of today) Hashinator branch
https://github.com/kstppd/hashinator/tree/non_blocking_experimental

Multi-stream speed issues have not been resolved yet. Blocking operations (due to unified memory) inside Hashinator were not the (only?) cause, so next up is evaluating page faults.

PR #727, opened 2023-03-21
Make differences derivatives

ykempf:
So this will go together with #711, I'll make that consistent with this present branch.


ykempf:
The big things are weeded out it seems, but I get too large diffs in the magnetosphere with ionosphere test. More to follow. Not tonight though.

ykempf:
Ah and not to self, check the coefficients of the Balsara reconstruction.

ykempf:
OK that's better but some diffs are still scary. More to investigate.

ykempf:
Meh. So I found out that something landed a mass density of 1e-38 or so instead of zero, making high Alfvén speed in the electric field solver. Hence my suggested fix to `divideIfNotZero`, is it similar to your approach elsewhere @alhom? I'll proceed with the next tests to see what causes their bugs, possibly I should protect some more of the division calls (assuming folks like this approach).

ykempf:
So https://github.com/fmihpc/vlasiator/actions/runs/4561406741/jobs/8047792342 wasn't completely satisfactory, therefore I replaced the fixed `divideIfNonZero` in the moments calculation with the equivalent to the previous, `divideIfNonNegative`.

ursg:
Those testpackage diffs are weird. Most of them seem to be at numerics level, but then acctest_4_helium hos a macroscopic one, even though that one is just a single cell like the other acctests, and shouldn't even have any influence of derivatives.

ykempf:
My view after not looking at/thinking of it during the weekend. I spent already way too much time on this attempt. I'll cherry-pick into #711 and have that then write out derivatives as derivatives instead of differences, and leave the division if non zero/negative/positive to another occasion.

ykempf:
Uh I canceled to not rerun but of course it doesn't show the results right here now. Meh. I'll relaunch them. Now there's nothing that should cause diffs so I suppose the remaining are the "accepted" diffs?!

ykempf:
Well these are pretty big for a no-change-in-the-solvers testpackage run.

ykempf:
How do I trigger a rerun of the CI tests without pushing a blank commit?

ursg:
In the "Actions" tab, there's a button for it:
example

PR #728, opened 2023-03-22
Testpackage profiling

lkotipal:
Run testpackage with `tau_exec` and collect results into TAUdb.


lkotipal:
aaa

lkotipal:
On hold for workunit support in Tau (https://github.com/UO-OACISS/tau2/issues/9)

PR #729, opened 2023-03-24
Fix generate_version

lkotipal:
Changes hex codes in `generate_version.sh` (and by proxy `version.cpp`) to character literals instead of integers. This fixes an error where non-ascii characters in the diff cause a failure to compile, as any one-byte character literal can be converted to `char`, but integers > 128 emit a narrowing conversion error.


ursg:
Looks good. If it looks stupid, but it works, it is not stupid. :D

PR #730, opened 2023-03-27
Greatly simplify Makefile

ursg:
the Makefile now makes use of gcc's automatic dependency generation (-MMD), so we no longer need to explicitly list dependencies for each source file, many of which were outdated anyway.

Also, most compilation is now done through generic pattern rules per-directory, and only a few files still need their own rules.

And to make things more pretty, the lengthy compiler output is now supressed by default, instead just printing "[CC] sourcefile.cpp". This can be toggled by building with "make V=1".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,14 +17,21 @@ endif
 include MAKE/Makefile.${ARCH}
 
 
+# For more silent make output (without echoing all commands)
+# Buiild with "make V=1" to get all verbose output
+SILENT_0 := @
+SILENT_1 :=
+V := 0
+SILENT = $(SILENT_$(V))
 
-
+# Automatically let the compiler build makefile dependencies
+COMPFLAGS += -MMD
 
 #set a default archive utility, can also be set in Makefile.arch
 AR ?= ar
 
 #londrillo_delzanna (no other options)
-FIELDSOLVER ?= londrillo_delzanna
+FIELDSOLVER ?= ldz_main
 #Add -DFS_1ST_ORDER_SPACE or -DFS_1ST_ORDER_TIME to make the field solver first-order in space or time
 # COMPFLAGS += -DFS_1ST_ORDER_SPACE
 # COMPFLAGS += -DFS_1ST_ORDER_TIME
@@ -41,15 +48,16 @@ COMPFLAGS += -DNDEBUG
 # COMPFLAGS += -DDEBUG_SOLVERS
 # COMPFLAGS += -DDEBUG_IONOSPHERE
 
+
 #Set order of semilag solver in velocity space acceleration
-#  ACC_SEMILAG_PLM 	2nd order	
-#  ACC_SEMILAG_PPM	3rd order 
+#  ACC_SEMILAG_PLM 	2nd order
+#  ACC_SEMILAG_PPM	3rd order
 #  ACC_SEMILAG_PQM      5th order (use this one unless you are testing)
 #Set order of semilag solver in spatial translation
 #  TRANS_SEMILAG_PLM 	2nd order	
 #  TRANS_SEMILAG_PPM	3rd order (for production use, use unless testing)
 #  TRANS_SEMILAG_PQM	5th order (significantly slower due to larger stencil)
-COMPFLAGS += -DACC_SEMILAG_PQM -DTRANS_SEMILAG_PPM 
+COMPFLAGS += -DACC_SEMILAG_PQM -DTRANS_SEMILAG_PPM
 
 #Add -DCATCH_FPE to catch floating point exceptions and stop execution
 #May cause problems
@@ -63,13 +71,13 @@ COMPFLAGS += -DACC_SEMILAG_PQM -DTRANS_SEMILAG_PPM
 #//////////////////////////////////////////////////////
 
 #will need profiler in most places..
-COMPFLAGS += ${INC_PROFILE} 
+COMPFLAGS += ${INC_PROFILE}
 
 #use jemalloc
-COMPFLAGS += ${INC_JEMALLOC} 
+COMPFLAGS += ${INC_JEMALLOC}
 
 #define precision
-COMPFLAGS += -D${FP_PRECISION} 
+COMPFLAGS += -D${FP_PRECISION}
 
 #define precision for the distribution function
 COMPFLAGS += -D${DISTRIBUTION_FP_PRECISION}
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -230,266 +183,77 @@ data:
...
+	$(SILENT)${CMP} ${CXXFLAGS} ${FLAGS} ${MATHFLAGS} -c $< ${INC_DCCRG} ${INC_FSGRID} ${INC_BOOST} ${INC_ZOLTAN} ${INC_EIGEN}
+
+# for all files in the projects/ dir
+%.o: projects/%.cpp projects/%.h
+	@echo [CC] $<
+	$(SILENT)${CMP} ${CXXFLAGS} ${FLAGS} ${MATHFLAGS} -c $< ${INC_DCCRG} ${INC_ZOLTAN} ${INC_BOOST} ${INC_EIGEN} ${INC_FSGRID} ${INC_VECTORCLASS}
+
+# (Second, more complex rules for the subdirectories of projects/)
+.SECONDEXPANSION:
+%.o: projects/$$*/$$*.cpp projects/$$*/$$*.h projects/projectTriAxisSearch.h
```
ykempf, line 231: Is this going to try to build the unsupported ones?

Makefile
```diff
@@ -504,19 +268,19 @@ DEPS_PARTICLES = particles/particles.h particles/particles.cpp particles/field.h
 OBJS_PARTICLES = particles/physconst.o particles/particles.o particles/readfields.o particles/particleparameters.o particles/distribution.o readparameters.o version.o particles/scenario.o particles/histogram.o
 
 vlsvextract: ${DEPS_VLSVREADER} ${DEPS_VLSVREADERINTERFACE} tools/vlsvextract.h tools/vlsvextract.cpp ${OBJS_VLSVREADER} ${OBJS_VLSVREADERINTERFACE}
-	${CMP} ${CXXFLAGS} ${FLAGS} -c tools/vlsvextract.cpp ${INC_BOOST} ${INC_DCCRG} ${INC_EIGEN} ${INC_VLSV} -I$(CURDIR) 
+	${CMP} ${CXXFLAGS} ${FLAGS} -c tools/vlsvextract.cpp ${INC_BOOST} ${INC_DCCRG} ${INC_EIGEN} ${INC_VLSV} -I$(CURDIR)
 	${LNK} -o vlsvextract_${FP_PRECISION} vlsvextract.o  ${OBJS_VLSVREADERINTERFACE} ${LIB_BOOST} ${LIB_DCCRG}  ${LIB_VLSV} ${LDFLAGS}
 
 vlsv2silo:  ${DEPS_VLSVREADERINTERFACE} tools/vlsv2silo.cpp  ${OBJS_VLSVREADERINTERFACE}
```
ykempf, line 274: I think this is as good as dead, has anyone maintained it since the last umpteen changes to our VLSV format we did? I actually just dropped the mention on our installation guide the other day.

Makefile
ursg: It's only going to build the ones that the vlasiator target has as dependencies. So no.

Makefile
ursg: I think I used it semi-recently still, but I agree it's a bit left n the dust. Might soon make sense to just dump it altogether and just use python where required.

ykempf:
...and could be looked up in the repo's history without much hassle.
________________________________

Makefile
```diff
@@ -117,77 +125,16 @@ LIBS += ${LIB_ZOLTAN}
...
-
-#DEPS_CPU_TRANS_MAP_AMR = ${DEPS_COMMON} ${DEPS_CELL} grid.h vlasovsolver/vec.h vlasovsolver/cpu_trans_map.hpp vlasovsolver/cpu_trans_map.cpp vlasovsolver/cpu_trans_map_amr.hpp vlasovsolver/cpu_trans_map_amr.cpp
-
-DEPS_VLSVMOVER = ${DEPS_CELL} vlasovsolver/vlasovmover.cpp vlasovsolver/cpu_acc_map.hpp vlasovsolver/cpu_acc_intersections.hpp \
-	vlasovsolver/cpu_acc_intersections.hpp vlasovsolver/cpu_acc_semilag.hpp vlasovsolver/cpu_acc_transform.hpp \
-	vlasovsolver/cpu_moments.h vlasovsolver/cpu_trans_map.hpp vlasovsolver/cpu_trans_map_amr.hpp
-
-DEPS_VLSVMOVER_VAMR = ${DEPS_CELL} vlasovsolver_amr/vlasovmover.cpp vlasovsolver_amr/cpu_acc_map_amr.hpp vlasovsolver_amr/cpu_acc_intersections.hpp \
+
+DEPS_VLSVMOVER_VAMR = vlasovsolver_amr/vlasovmover.cpp vlasovsolver_amr/cpu_acc_map_amr.hpp vlasovsolver_amr/cpu_acc_intersections.hpp \
```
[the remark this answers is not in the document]
ursg: True true. But that's a matter for a separate PR.

Makefile
markusbattarbee: I used this just last December to check on a specific VDF of FHA which was reporting hitting the v-space wall. Downloading a single VDF .vlsv for a quick diagnostic was a lot easier than downloading a full restart file... So let's definately keep it in.

markusbattarbee:
Looks like both NVCC and CLANG support "-MMD", but I was unable to discern if HIPCC supports it? Any other compilers we'd need to make sure we support?

ursg:
Since crayCC is also llvm-based, I expect it will also provide the -MMD option. But it probably makes sense to check.

Either way, in case we get a platform where the compiler does not support dependency file generation, compiling will still work just fine without it -- it just requires more ```make clean``` in between, which most people anyway do all the time since our hand-curated dependencies were outdated more often than they were correct.

Makefile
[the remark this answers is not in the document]
ykempf: Well, I would still accept it as part of this Makefile Easter cleaning, but your call.

Makefile
ykempf: Add a comment to that effect maybe?

Makefile
ykempf: Sorry I was not clear and the comment interface isn't showing where I actually clicked: I meant ditching vlsv2silo, definitely not vlsvextract, I'm a big fan of the latter!

ursg: Ah, true. I don't believe anyone has used that silo export in the last 5 (10?) years. But let's make that a separate PR.

markusbattarbee:
> Either way, in case we get a platform where the compiler does not support dependency file generation, compiling will still work just fine without it -- it just requires more `make clean` in between, which most people anyway do all the time since our hand-curated dependencies were outdated more often than they were correct.

Ok, that's a good thing to know. (And I guess it'll then also require editing the Makefile to remove that -MMD flag)

ursg:
I just checked on Perlmutter (which is basically the same architecture as LUMI): both crayCC and hipcc are clang-derived, and both support -MMD.

ursg:
The only other relevant C-compiler for HPC is IBM xlc (if we ever target something as weird as a BlueGene system again), and that [supports -MMD too](https://www.ibm.com/docs/en/xl-c-and-cpp-aix/16.1?topic=descriptions-qmakedep-m).

PR #731, opened 2023-03-31
Print both relative and abso error variables in testpackage CI output

ursg:
Before, only the maximum absolute error determined which variable would be lifted to the tests' one-line status.


ursg:
Works, apparently!

PR #737, opened 2023-04-06
Squelch warning about uninitialized neighborhood variable.

ursg:
In faceNeighbor calculation, we always trust that the dimension values we get back from DCCRG are sane.
This also introduces a safety check, which will hopefully never trigger.

This PR is part of my "Remove one compiler warning per day" initiative.


ykempf:
Shouldn't that return an appropriate error code? Or was that unmerged yet?

ursg:
I would claim that the defaut abort code 255 ("Something went total bananas, better have a good look at what happens") is totally appropriate here.

PR #744, opened 2023-04-13
Updated profiling in fieldsolver, corrected samples/Magnetosphere3d i…

markusbattarbee:
…nflow magnetic field, made lumi makefile system library includes use -isystem


fieldsolver/derivatives.cpp
```diff
@@ -364,25 +364,29 @@ void calculateDerivativesSimple(
    
    phiprof::stop(timer);
 
-   timer=phiprof::initializeTimer("Compute cells");
-   phiprof::start(timer);
+   //timer=phiprof::initializeTimer("Compute cells");
+   //phiprof::start(timer);
```
ykempf, line 368: Why comment and not delete (same for a few others further down)?

fieldsolver/derivatives.cpp
```diff
@@ -364,25 +364,29 @@ void calculateDerivativesSimple(
...
-      for (int j=0; j<gridDims[1]; j++) {
-         for (int i=0; i<gridDims[0]; i++) {
-            if (technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE) continue;
-            if (RKCase == RK_ORDER1 || RKCase == RK_ORDER2_STEP2) {
-               calculateDerivatives(i,j,k, perBGrid, momentsGrid, dPerBGrid, dMomentsGrid, technicalGrid, sysBoundaries, RKCase);
-            } else {
-               calculateDerivatives(i,j,k, perBDt2Grid, momentsDt2Grid, dPerBGrid, dMomentsGrid, technicalGrid, sysBoundaries, RKCase);
+   #pragma omp parallel
+   {
+      phiprof::start("FS derivatives compute cells");
```
ykempf, line 373: Why not call this guy just outside of the threaded region?

fieldsolver/ldz_hall.cpp
```diff
@@ -790,7 +790,7 @@ void calculateHallTerm(
  * \param sysBoundaries System boundary condition functions.
  * \param RKCase Element in the enum defining the Runge-Kutta method steps
  * \param communicateMomentsDerivatives whether to communicate derivatves with the neighbour CPUs
- * 
+ *
```
ykempf, line 793: Hooray tens of whitespace lines.

ykempf (commented):
So yeah I was wondering why you now everywhere open the omp parallel regions, then start the timers, then have the omp for loops.

fieldsolver/derivatives.cpp
markusbattarbee: Because if the phiprof call is outside the threaded region, we don't get any threading balance information in phiprof. At least that's how I understand it.

ursg: also, only if we open them like this do we see the thread imbalances properly in tau traces

ykempf:
Approved, testpackage is clean.

PR #746, opened 2023-04-13
Threading backgroundfield init

ykempf:
phiprof calls left in there for now, to be investigated also at scale – does it work better now? Posting for CI output too.

### DIFF
--- a/backgroundfield/backgroundfield.cpp
+++ b/backgroundfield/backgroundfield.cpp
@@ -203,7 +203,7 @@ void setPerturbedField(
                   surfaceAverage(valueFunction,
                      (coordinate)fComponent,
                                  accuracy,
-                                 start,
+                                 start.data(),
                                  dx[faceCoord1[fComponent]],
                                  dx[faceCoord2[fComponent]]
                                 );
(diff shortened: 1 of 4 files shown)


backgroundfield/backgroundfield.cpp
```diff
@@ -41,93 +42,94 @@ void setBackgroundField(
...
+      faceCoord1[0]=1;
+      faceCoord2[0]=2;
+      faceCoord1[1]=0;
+      faceCoord2[1]=2;
+      faceCoord1[2]=0;
+      faceCoord2[2]=1;
+      
+      auto localSize = BgBGrid.getLocalSize();
+      
+      // TODO: These can be threaded now that the dipole field is threadsafe
```
ursg, line 62: I suppose this comment is now moot?

ursg (approved):
CI output looks good, too.

markusbattarbee:
The only significant diff is from nonthermal moments, which is a small small subpopulation anyway and sensitive to close-to-numerical changes, so I wouldn't be too worried about that.

ykempf:
In this goes?

PR #748, opened 2023-04-18
Squelch more unused variable warnings.

ursg:
These ones are pretty straightforward.

This PR is part of my "remove one compiler warning per day" initiative.

### DIFF
--- a/datareduction/datareductionoperator.cpp
+++ b/datareduction/datareductionoperator.cpp
@@ -1426,7 +1426,9 @@ namespace DRO {
       const uint vectorSize = 3;
       //Input data into buffer
       const char* ptr = reinterpret_cast<const char*>(&PTensor);
-      for (uint i = 0; i < 3*sizeof(Real); ++i) buffer[i] = ptr[i];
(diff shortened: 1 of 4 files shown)


datareduction/datareductionoperator.cpp
```diff
@@ -1423,7 +1423,6 @@ namespace DRO {
       const bool calculateNonthermal = true;
       //Calculate and save:
       PTensorOffDiagonalNonthermalCalculations( cell, calculateNonthermal, averageVX, averageVY, averageVZ, popID, PTensor );
-      const uint vectorSize = 3;
       //Input data into buffer
       const char* ptr = reinterpret_cast<const char*>(&PTensor);
       for (uint i = 0; i < 3*sizeof(Real); ++i) buffer[i] = ptr[i];
```
markusbattarbee, line 1428: Actually, for these I'd recommend we rather use vectorSize in this loop, because we can indeed have datareducers which have different length vectors.

ursg: Alright, updated. vectorSize is now being used (and I properly formatted the loops with braces, too!)

PR #755, opened 2023-04-26
Remove "#pragma ivdep" warnings by replacing with "#pragma omp simd"

ursg:
Hooray for this thing actually being standardized in OpenMP 4!

Compare, for example https://blog.rwth-aachen.de/hpc_import_20210107/attachments/28344675/30867475.pdf

(This might warrant a specific performance check, to see that it actually produces code of the same speed)

This PR is part of my "remove one compiler warning per day" initiative.

### DIFF
--- a/vlasovsolver/cpu_1d_pqm.hpp
+++ b/vlasovsolver/cpu_1d_pqm.hpp
@@ -99,11 +99,7 @@ inline void filter_pqm_monotonicity(Vec *values, uint k, Vec &fv_l, Vec &fv_r, V
       //todo store and then load data to avoid inserts (is it beneficial...?)
       
 //serialized the handling of inflexion points, these do not happen for smooth regions
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wunknown-pragmas"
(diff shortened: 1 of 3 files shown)


markusbattarbee:
Can confirm that Mahti NVCC also does not complain of this pragma!

ursg:
Hmm, the automatic testpackage output indiid looks like it is a little bit slower than the reference data. Might be a random fluctuation of the cluster though?

Should we maybe test this at scale?

markusbattarbee:
At least we need new reference data, hard to say from those 1,6xx speedup values

lkotipal:
Yes, in my experience the speedup on the automatic testpackage output doesn't seem to have anything to do with real results.

ursg:
Looking with objdump at least shows that the assembly output is not identical, which I had hoped. Alright, doing this properly then.

ursg:
Now I reran with updated reference data, and it does indeed look like it is a tiny bit slower (to the extent that we can trust carrington's performance results). So hum, should we leave this PR dangling for now? It is at least useful to have this change available when we need it.

PR #767, opened 2023-05-13
Amr translation cleanup

markusbattarbee:
This PR moves all AMR translation pencil functionality into its own file. Any metadata which remains the same between LB calls is stored. All pencil construction and splitting functions have been re-reviewed to work correctly. The logic of storing ids for a pencil is slightly different as now it also stores VLASOV_STENCIL_WIDTH cells at each end. 

The cell widths (dz) in translation are now passed as Realf because it's always constant for all blocks and Vecs within a cell.

The looping structure of AMR translation was heavily modified, so that instead of performing 4 write operations (read into buffer, write translated value into 1st and 2nd intermediate buffer, finally back into cell) per translated cell per cycle, there is only one read/write into ordered source data and one write back into the cell.

Main benefit is simplification of memory management and preparation for heterogeneous architectures. Performance improvements may occur if previous approach had sub-optimal caching etc. I'll run tests on Carrington to see.

Things not modified in this PR, but which could be reviewed in the future: remote neighbour contribution, consolidating the neighbour finding functions in pencil construction.


markusbattarbee:
Did the sample Magnetosphere config on Carrington, 8 nodes. Appears that cache accessing has been good with the old version as well, potential improvement falls within random deviation. Mind you, this was still RCB, not RIB.
```
----------------------------------------------------------------------------------------------------------------------------------------------------------
 |                                               | Count  |         Process time          |   Thread imbalances    | Workunits                 | 
------------------------------------------------------------------------------------------------------------------------------------------------------
 | Id  | Lvl | Grp | Name                        | Avg    | Avg (s)   | Time % | Imb %    | No | Avg %   | Max %   | Avg                       | 
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Dev
 | 183 | 3   |     |     Propagate               | 801    | 1.218e+04 | 88.31  | 9.899    | 1  |         |       | 6.835e+06 Cells/s         | 
 | 184 | 4   |     |       Spatial-space         | 801    | 3130      | 25.69  | 0.1054   | 1  |         |       | 2.66e+07 Cells/s          | 
 | 220 | 4   |     |       Update boundaries tra | 801    | 1035      | 8.493  | 56.29    | 1  |         |       |                           | 
 | 230 | 4   |     |       Propagate Fields      | 801    | 4151      | 34.06  | 8.029    | 1  |         |       | 1126 SpatialCells/s       | 
 | 254 | 4   |     |       Velocity-space        | 801    | 1428      | 11.72  | 48.89    | 1  |         |       | 5.833e+07 Cells/s         | 
 | 268 | 4   |     |       Update boundaries acc | 801    | 2440      | 20.03  | 0.3463   | 1  |         |       |                           | 
Opt
 | 184 | 3   |     |     Propagate               | 801    | 1.269e+04 | 89.27  | 7.102    | 1  |          |         | 6.567e+06 Cells/s         | 
 | 185 | 4   |     |       Spatial-space         | 801    | 3070      | 24.19  | 0.1617   | 1  |          |         | 2.715e+07 Cells/s         | 
 | 227 | 4   |     |       Update boundaries tra | 801    | 1331      | 10.49  | 42.22    | 1  |          |         |                           | 
 | 237 | 4   |     |       Propagate Fields      | 801    | 4125      | 32.5   | 9.544    | 1  |          |         | 1110 SpatialCells/s       | 
 | 261 | 4   |     |       Velocity-space        | 801    | 1493      | 11.77  | 41.16    | 1  |          |         | 5.581e+07 Cells/s         | 
 | 275 | 4   |     |       Update boundaries acc | 801    | 2671      | 21.05  | 0.7445   | 1  |          |         |                           |

**Comment by markusbattarbee** [comment]
Decided to also look in detail at the actual time spent in translation-mapping+store, as of course the imbalances can mask true performance. And then we do see a 10% increase in performance! (compare the compute-mapping-sum values)
```
------------------------------------------------------------------------------------------------------------------------------------------------
 |                                               | Count  |         Process time          |   Thread imbalances    | Workunits                 | 
------------------------------------------------------------------------------------------------------------------------------------------------
 | Id  | Lvl | Grp | Name                        | Avg    | Avg (s)   | Time % | Imb %    | No | Avg %   | Max %   | Avg                       | 
------------------------------------------------------------------------------------------------------------------------------------------------
Dev
 | 183 | 3   |     |     Propagate               | 801    | 1.218e+04 | 88.31  | 9.899    | 1  |         |         | 6.835e+06 Cells/s         | 
 | 184 | 4   |     |       Spatial-space         | 801    | 3130      | 25.69  | 0.1054   | 1  |         |         | 2.66e+07 Cells/s          | 
 | 190 | 7   |     |         compute-mapping-z   | 801    | 529.8     | 17.09  | 18.29    | 1  |         |         |                           | 
 | 200 | 7   |     |         compute-mapping-x   | 801    | 369.3     | 11.91  | 21.21    | 1  |         |         |                           | 
 | 210 | 7   |     |         compute-mapping-y   | 801    | 455.2     | 14.68  | 14.97    | 1  |         |         |                           | 
 |.    | 7   |     |         compute-mapping-sum | 801    | 1354.3
 | 220 | 4   |     |       Update boundaries tra | 801    | 1035      | 8.493  | 56.29    | 1  |         |         |                           | 
 | 230 | 4   |     |       Propagate Fields      | 801    | 4151      | 34.06  | 8.029    | 1  |         |         | 1126 SpatialCells/s       | 
 | 254 | 4   |     |       Velocity-space        | 801    | 1428      | 11.72  | 48.89    | 1  |         |         | 5.833e+07 Cells/s         | 
 | 268 | 4   |     |       Update boundaries acc | 801    | 2440      | 20.03  | 0.3463   | 1  |         |         |                           | 
Opt
 | 184 | 3   |     |     Propagate               | 801    | 1.269e+04 | 89.27  | 7.102    | 1  |         |         | 6.567e+06 Cells/s         | 
 | 185 | 4   |     |       Spatial-space         | 801    | 3070      | 24.19  | 0.1617   | 1  |         |         | 2.715e+07 Cells/s         | 
 | 191 | 7   |     |         compute-mapping-z   | 801    | 466.4     | 15.34  | 26.88    | 1  |         |         |                           | 
 | 203 | 7   |     |         compute-mapping-x   | 801    | 346.9     | 11.41  | 30.23    | 1  |         |         |                           | 
 | 215 | 7   |     |         compute-mapping-y   | 801    | 407.5     | 13.41  | 25.52    | 1  |         |         |                           | 
 |     | 7   |     |         compute-mapping-sum | 801    | 1220,8    | 
 | 227 | 4   |     |       Update boundaries tra | 801    | 1331      | 10.49  | 42.22    | 1  |         |         |                           | 
 | 237 | 4   |     |       Propagate Fields      | 801    | 4125      | 32.5   | 9.544    | 1  |         |         | 1110 SpatialCells/s       | 
 | 261 | 4   |     |       Velocity-space        | 801    | 1493      | 11.77  | 41.16    | 1  |         |         | 5.581e+07 Cells/s         | 
 | 275 | 4   |     |       Update boundaries acc | 801    | 2671      | 21.05  | 0.7445   | 1  |         |         |                           | 
```
I'm also currently running a variant where I reverted DZ to use `Vec` to see if that has any effect.

**Comment by markusbattarbee** [comment]
So the DZ-Vec change had pretty much negligible effect. 
Using Realf might help a tiny bit with memory management, and it wasn't at least slower, so I'm happy with this PR in its current form. 

After this gets merged, I can update the other amr PR #609 to work with this.

**Comment by markusbattarbee** [comment]
Something that now occurred to me: the load balance evaluation on line
https://github.com/markusbattarbee/vlasiator-public/blob/f5c3ba9c73bd8a98f84a3a24ac3f29a95a5a3400/vlasovsolver/cpu_trans_map_amr.cpp#LL296C10-L296C10
is now different from before. 

It used to be that it calculates how many times each local cell is the centrepoint of translation.
 
In its current form, it calculates how many times the data from each cell needs to be read in order to perform translation.

Opinions?

File: vlasovsolver/cpu_trans_pencils.cpp [hunk 1]
```diff
@@ -0,0 +1,1170 @@
...
+      for (const auto& [neighbor, dir] : mpiGrid.get_face_neighbors_of(celli) ) {
+         if ( dir == -((int)dimension + 1) ) {
+            // Check that the neighbor is not across a periodic boundary by calculating
+            // the distance in indices between this cell and its neighbor.
+            auto nbrIndices = mpiGrid.mapping.get_indices(neighbor);
+
+            // If a neighbor is across a periodic boundary, non-local, or
+            // in non-periodic boundary layer 2
+            // then we use the current cell as a seed for pencils
+            if ( (myIndices[dimension] < nbrIndices[dimension]) ||
```
markusbattarbee, line 672: So this point should fix the issue with a 2-cell wide periodic domain.

It used to read along the lines of:
```
	    if ( abs ( (int64_t)(myIndices[dimension] - nbrIndices[dimension]) ) >
		 pow(2,mpiGrid.get_maximum_refinement_level()) ||
```
but as in this section we are always looking at negative direction neighbours, I think a simple greater-than check should be enough. Thoughts? @ursg @ykempf

ykempf: So this checks if my index is larger than my left neighbour's, correct? And that means my left neighbour is across the periodic wall?

markusbattarbee: Well, smaller than, but yes. Left neighbours usually have a smaller index, so a larger index would mean it's actually further along in the grid, which realistically should mean it's across a periodic boundary.

ykempf: smaller, of course – I agree with this then.

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -0,0 +1,208 @@
...
+   // Split one pencil into up to four pencils covering the same space.
+   // dx and dy are the dimensions of the original pencil.
+   void split(const uint myPencilId, const Real dx, const Real dy) {
+      auto myIds = this->getIds(myPencilId);
+
+      // Find paths that members of this pencil may have in other pencils (can happen)
+      // so that we don't add duplicates.
+      std::vector<int> existingSteps;
+
+#pragma omp parallel for
```
ykempf, line 124: No big difference with a schedule here I suppose?

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -713,51 +35,53 @@ bool check_skip_remapping(Vec* values) {
...
-      
-   }
-   
-   // Go from 0 to length here to propagate all the cells in the pencil
-   for (uint i = 0; i < lengthOfPencil; i++){
-      
-      // The source array is padded by VLASOV_STENCIL_WIDTH on both sides.
-      uint i_source   = i + VLASOV_STENCIL_WIDTH;
-      
+   // const uint nTargetNeighborsPerPencil = 1;
```
ykempf, line 60: Is this comment still needed?

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -772,284 +96,94 @@ void propagatePencil(
...
          //    std::exit(1);
          // }
-         
-         for (uint planeVector = 0; planeVector < VEC_PER_PLANE; planeVector++) {   
 
+         // Loop over Vec's in current plance
+         for (uint planeVector = 0; planeVector < VEC_PER_PLANE; planeVector++) {
             // Check if all values are 0:
-            if (check_skip_remapping(values + i_trans_ps_blockv_pencil(planeVector, k, i-VLASOV_STENCIL_WIDTH, lengthOfPencil))) continue;
+            if (check_skip_remapping(values + i_trans_ps_blockv_pencil(planeVector, k, i, lengthOfPencil))) continue;
```
ykempf, line 103: Just to be sure, the dropping of this `-VLASOV_STENCIL_WIDTH` is intentional?

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1451,12 +258,23 @@ bool trans_map_1d_amr(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+
+   // Vector with all cell ids
+   vector<CellID> allCells(localPropagatedCells);
+   allCells.insert(allCells.end(), remoteTargetCells.begin(), remoteTargetCells.end());
+
+   // Vectors of pointers to the cell structs
+   std::vector<SpatialCell*> allCellsPointer(allCells.size());
 
+   // Initialize allCellsPointer
+   #pragma omp parallel for
```
ykempf, line 274: Add a schedule?

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1953,36 +672,36 @@ void update_remote_mapping_contribution_amr(
...
 
                   // Allocate memory for one sibling at recvIndex.
-                  
+
                   recvIndex = get_sibling_index(mpiGrid,nbr);
 
                   ncell->neighbor_number_of_blocks.at(recvIndex) = ccell->get_number_of_velocity_blocks(popID);
                   ncell->neighbor_block_data.at(recvIndex) =
-                     (Realf*) aligned_malloc(ncell->neighbor_number_of_blocks.at(recvIndex) * WID3 * sizeof(Realf), 64);
+                     (Realf*) aligned_malloc(ncell->neighbor_number_of_blocks.at(recvIndex) * WID3 * sizeof(Realf), WID3);
```
ykempf, line 684: If someone gets odd ideas, will this crash and burn?

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1991,39 +710,39 @@ void update_remote_mapping_contribution_amr(
...
                         && ncell->neighbor_number_of_blocks.at(i_sib) != scell->get_number_of_velocity_blocks(popID)
                         && scell->sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY) {
-                  
-                        
+
+
                         ncell->neighbor_number_of_blocks.at(i_sib) = scell->get_number_of_velocity_blocks(popID);
                         ncell->neighbor_block_data.at(i_sib) =
-                           (Realf*) aligned_malloc(ncell->neighbor_number_of_blocks.at(i_sib) * WID3 * sizeof(Realf), 64);
+                           (Realf*) aligned_malloc(ncell->neighbor_number_of_blocks.at(i_sib) * WID3 * sizeof(Realf), WID3);
```
ykempf, line 717: ditto

vlasovsolver/cpu_trans_pencils.cpp
```diff
@@ -0,0 +1,1170 @@
...
+std::array<std::unordered_set<CellID>,3> DimensionTargetCells;
+
+//Is cell translated? It is not translated if DO_NO_COMPUTE or if it is sysboundary cell and not in first sysboundarylayer
+bool do_translate_cell(SpatialCell* SC){
+   if(SC->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE ||
+      (SC->sysBoundaryLayer != 1 && SC->sysBoundaryFlag != sysboundarytype::NOT_SYSBOUNDARY))
+      return false;
+   else
+      return true;
+}
```
ykempf, line 18: This wants some PADDINGton? :teddy_bear:

vlasovsolver/cpu_trans_pencils.cpp
```diff
@@ -0,0 +1,1170 @@
...
+
+void flagSpatialCellsForAmrCommunication(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
+                                         const vector<CellID>& localPropagatedCells) {
+
+   // Only flag/unflag cells if AMR is active
+   if (mpiGrid.get_maximum_refinement_level()==0) return;
+
+   // return if there's no cells to flag
+   if(localPropagatedCells.size() == 0) {
+      std::cerr<<"No cells!"<<std::endl;
```
ykempf, line 70: Should we take this opportunity and make this more explicit? Every now and then I run into this message during some tests and I'm worried and dig up this part of the code to check.

vlasovsolver/cpu_trans_pencils.cpp
```diff
@@ -0,0 +1,1170 @@
...
+      return;
+   }
+
+   for (int dimension=0; dimension<3; dimension++) {
+      // These neighborhoods now include the AMR addition beyond the regular vlasov stencil
+      int neighborhood = getNeighborhood(dimension,VLASOV_STENCIL_WIDTH);
+
+      // Set flags: loop over local cells
+#pragma omp parallel for
```
ykempf, line 79: schedule?

vlasovsolver/cpu_trans_pencils.cpp
```diff
@@ -0,0 +1,1170 @@
...
+                                        std::vector<uint> path,
+                                        Realf* sourceDZ,
+                                        Realf* targetRatios
+                                        ){
+
+   // These neighborhoods now include the AMR addition beyond the regular vlasov stencil
+   int neighborhood = getNeighborhood(dimension,VLASOV_STENCIL_WIDTH);
+   stringstream ss;
+   for (uint j = 0; j < L; ++j) {
+      ss<< ids[j] << " ";
```
ykempf, line 208: Is it worth putting this behind some DEBUG flag? Maybe not as it's a rare thing anyway?

vlasovsolver/cpu_trans_pencils.cpp
```diff
@@ -0,0 +1,1170 @@
...
+                vector<CellID> &seedIds) {
+
+   const bool debug = false;
+   int myRank;
+   if (debug) MPI_Comm_rank(MPI_COMM_WORLD,&myRank);
+
+   // These neighborhoods now include the AMR addition beyond the regular vlasov stencil
+   int neighborhood = getNeighborhood(dimension,VLASOV_STENCIL_WIDTH);
+
+#pragma omp parallel for
```
ykempf, line 637: Same question, scheduling?

vlasovsolver/cpu_trans_pencils.cpp
```diff
@@ -0,0 +1,1170 @@
...
+   int neighborhood = getNeighborhood(dimension,VLASOV_STENCIL_WIDTH);
+
+   int myRank;
+   if(debug) {
+      MPI_Comm_rank(MPI_COMM_WORLD,&myRank);
+   }
+
+   std::vector<CellID> pencilIdsToSplit;
+
+// Thread this loop here
```
ykempf, line 793: Need this comment?
(and scheduling? :P)

ykempf (commented):
I didn't go through the actual contents, I am trusting that if we get clean diffs from a reasonable 3D AMR test we'll be good.

vlasovsolver/cpu_trans_pencils.hpp
markusbattarbee: I would not schedule this. The cost is not suitable for static, but going to (1) would be overkill as it's still a very cheap loop, and not done often.

vlasovsolver/cpu_trans_map_amr.cpp
markusbattarbee: ok, can clean.

vlasovsolver/cpu_trans_map_amr.cpp
markusbattarbee: No, it just aligns the data based on the block size, which is now correct. We still want alignment even if we change WID3.

vlasovsolver/cpu_trans_pencils.cpp
markusbattarbee: no, as Padding type is only for FSgrid cells.

vlasovsolver/cpu_trans_pencils.cpp
markusbattarbee: nope, I think the regular dynamic schedule here is just fine.

vlasovsolver/cpu_trans_pencils.cpp
markusbattarbee: ok, can make it more wordy.

vlasovsolver/cpu_trans_pencils.cpp
markusbattarbee: It should never happen, is a fallthrough error.

vlasovsolver/cpu_trans_pencils.cpp
markusbattarbee: This is again a lightweight loop but with non-constant computational cost, so I think regular dynamic schedule is fine.

vlasovsolver/cpu_trans_pencils.cpp
markusbattarbee: Deleted the comment. This is also a lightweight but non-constant loop so dynamic is good.

markusbattarbee:
Now with a recent CI run, and up-to-date reference data, we see a bit more diffs in AMR testpackage runs that should come from pure summation order changes. One potential source of diffs is the change to weight calculation I listed.

In order to be able to evaluate what changes actually took place in each function when they were split into different files, I have made a dummy draft PR which contains the old functions but placed in the new files:
https://github.com/markusbattarbee/vlasiator-public/pull/14/files?diff=unified&w=1

It's still a lot of diffs, but should be at least theoretically possible to evaluate.

vlasovsolver/cpu_trans_pencils.cpp
markusbattarbee: correction: regular static is good enough here.

ursg (approved):

ykempf (approved):

PR #701, opened 2023-02-16
AMR Update

lkotipal:
- Remove infinities and NANs from J/B_perp
- Use total B instead of perturbed when calculating alpha
- Report amount of cells set to refine
- ~~Optional threshold `AMR.least_cells_to_refine` for minimum amount of cells to refine as a fraction of grid.~~
- Don't scale constituents of alpha

PR #769, opened 2023-05-19
Remove two unused functions from derivatives.cpp

ursg:
(They were trivial anyway, and thus probably never been taken into use)

This PR is part of my "remove one compiler warning per day" initiative


ykempf:
And if next week @lkotipal needs to add a check on full B or E differences as part of getting the FHB dAMR tests going, he'll have to rewrite this? I'm not sure this is particularly useful right now.

lkotipal:
`getPerB` is switched to `getB` in #701. `getE` is indeed unused as `U` is approximated to just the kinetic energy and magnetic field contribution.

ykempf:
So? Can be merged?

PR #775, opened 2023-05-29
Shared Fsgrid coupling, recalculated on LB

ursg:
This moves the fsgrid coupling information into the shared coupling object (which remained woefully unused so far).

Also, coupling is now only being recalculated on LB, and not multiple times every timestep.

Untested. Let's see what the CI does with this.

(This requires [the fsgrid branch that actually contains these structures](https://github.com/fmihpc/fsgrid/tree/couplingData), which the CI script automatically checks out already)

### DIFF
--- a/fieldsolver/gridGlue.cpp
+++ b/fieldsolver/gridGlue.cpp
@@ -39,27 +39,22 @@ int getNumberOfCellsOnMaxRefLvl(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geomet
 */
 
 template <typename T, int stencil> void computeCoupling(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
-							const std::vector<CellID>& cells,
-							FsGrid< T, stencil>& momentsGrid,
-							std::map<int, std::set<CellID> >& onDccrgMapRemoteProcess,
-							std::map<int, std::set<CellID> >& onFsgridMapRemoteProcess,
-							std::map<CellID, std::vector<int64_t> >& onFsgridMapCells
-							) {
-    
+      const std::vector<CellID>& cells,
+      FsGrid< T, stencil>& momentsGrid,
+      FsGridCouplingInformation* coupling) {
+
   //sorted list of dccrg cells. cells is typicall already sorted, but just to make sure....
   std::vector<CellID> dccrgCells = cells;
   std::sort(dccrgCells.begin(), dccrgCells.end());
 
   //make sure the datastructures are clean
-  onDccrgMapRemoteProcess.clear();
-  onFsgridMapRemoteProcess.clear();
-  onFsgridMapCells.clear();
-  
-  
+  coupling->onDccrgMapRemoteProcess.clear();
+  coupling->onFsgridMapRemoteProcess.clear();
+  coupling->onFsgridMapCells.clear();
+
   //size of fsgrid local part
   const std::array<int, 3> gridDims(momentsGrid.getLocalSize());
-  
- 
+
   //Compute what we will receive, and where it should be stored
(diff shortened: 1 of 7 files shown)


ykempf:
It throws
> System boundary 6 is invalid

in all tests.

ykempf:
But 6 is `N_SYSBOUNDARY_CONDITIONS` so there's something dodgy there for sure.

ykempf:
We're probably missing one initial `computeCoupling()` call early in, as now it's only called once in the `balanceLoad()` call of `vlasiator.cpp`.

fieldsolver/gridGlue.hpp
```diff
@@ -186,5 +186,62 @@ template< unsigned int numFields > void getFieldDataFromFsGrid(
...
+
+  onDccrgMapRemoteProcess   maps fsgrid processes (key) => set of dccrg cellIDs owned by current rank that map to  the fsgrid cells owned by fsgrid process (val)
+
+  onFsgridMapRemoteProcess  maps dccrg processes  (key) => set of dccrg cellIDs owned by dccrg-process that map to current rank fsgrid cells 
+  onFsgridMapCells          maps remote dccrg CellIDs to local fsgrid cells
+*/
+
+template <typename T, int stencil> void computeCoupling(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
+      const std::vector<CellID>& cells,
+      FsGrid< T, stencil>& momentsGrid,
```
markusbattarbee, line 201: Here this should be renamed to technicalgrid as that's the one it now gets.
Do we even want to pass an explicit coupling as a parameter, if it's included in the technicalgrid anyway?

vlasiator.cpp
```diff
@@ -454,6 +454,7 @@ int main(int argn,char* args[]) {
    );
    
    const std::vector<CellID>& cells = getLocalCells();
+   computeCoupling(mpiGrid, cells, momentsGrid, &gridCoupling);
```
markusbattarbee, line 457: This is now in conflict with how `balanceLoad()` calls the `computeCoupling()` with `technicalGrid` and `technicalGrid.coupling`

fieldsolver/gridGlue.cpp
```diff
@@ -565,12 +487,12 @@ void feedBoundaryIntoFsGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
   std::vector<MPI_Request> receiveRequests;
   
   //computeCoupling
-  computeCoupling(mpiGrid, cells, technicalGrid, onDccrgMapRemoteProcess, onFsgridMapRemoteProcess, onFsgridMapCells);
+  //computeCoupling(mpiGrid, cells, technicalGrid, technicalGrid.coupling);
```
markusbattarbee, line 490: Let's add a comment here that this is now recomputed at load balance

fieldsolver/gridGlue.cpp
```diff
@@ -425,23 +352,23 @@ void getFieldsFromFsGrid(
...
   }
-  
+
   MPI_Waitall(receiveRequests.size(), receiveRequests.data(), MPI_STATUSES_IGNORE);
 
 
   //Aggregate receives, compute the weighted average of these
   ii=0;
-  for (auto const &rcv : onDccrgMapRemoteProcess){
+  for (auto const &rcv : BgBGrid.coupling->onDccrgMapRemoteProcess){
```
markusbattarbee, line 371: this (and all above references to `BgBGrid.coupling` or `momentsGrid.coupling`) should instead access `technicalGrid.coupling`, I wager

markusbattarbee:
This is deprecated now, right @ursg?

ursg:
Yup, I suppose it was obviated by #1012

PR #777, opened 2023-05-31
Update PAPI github url in CI build script.

ursg:
Since the old repository is now empty, this was breaking the CI build process.


ursg:
This has compiled succesfully, so I'll merge it before it even had a chance to run the testpackage.

PR #779, opened 2023-06-08
Enable compiler autovectorization diagnostics in CI.

ursg:
This PR is mostly to check whether the resulting flood of compiling outputs shows anything reasonable in the check annotations.


ursg:
Turns out, the vector notification outputs of the compiler are not automatically turned into github source annotations by the scripts we have, so that makes this PR a bit pointless. Well, it was worth a try.

PR #782, opened 2023-06-08
Build testpackage definition automatically from directories.

ursg:
Every test should now have a test_spec.sh shell fragment that adds to the corresponding shell arrays (the preexisting ones have been added appropriately).

Small_test_definitions.sh then, instead of having the hardcoded list of tests, iterates through all tests/ directories and builds the array from those.

Alternatively, the user can still supply a list of test names to run in the small_test_definitinos.sh tests array.

### DIFF
--- a/testpackage/small_test_definitions.sh
+++ b/testpackage/small_test_definitions.sh
@@ -22,7 +22,7 @@ tests=()
 
 # If no tests are specified, run all tests/ subdirectories.
 if [ "${#tests[@]}" -eq "0" ]; then
-   tests=($(ls -d tests/*))
+   tests=($(/bin/ls -d tests/*))
    echo "No run_tests list specified, using all ${#tests[@]} subdirectories of tests/"
 fi;
 


testpackage/small_test_definitions.sh
```diff
@@ -18,138 +18,28 @@ run_dir="run"
...
-# acceleration test
-test_name[1]="acctest_2_maxw_500k_100k_20kms_10deg"
-comparison_vlsv[1]="fullf.0000001.vlsv"
-comparison_phiprof[1]="phiprof_0.txt"
-variable_names[1]="proton/vg_rho proton/vg_v proton/vg_v proton/vg_v proton"
-variable_components[1]="0 0 1 2"
-single_cell[1]=1
+# If no tests are specified, run all tests/ subdirectories.
+if [ "${#tests[@]}" -eq "0" ]; then
+   tests=($(ls -d tests/*))
```
ykempf: Use `/usr/bin/ls` to avoid messups with default `--color`.

ykempf:
Right so restart write should come before read unlike the alphabet, so these will need to either be redesigned to be one test in one go, or sequential.

markusbattarbee:
Hmm. It would perhaps make sense to place the actual launch commands for each test in a file inside that directory. 

Then each launch file could include several launch commands, and eg. the read-write-test could be a single one.

If we want to compare different rank/thread combos, then we could also trigger those from those files.

PR #783, opened 2023-06-08
Bailout when fg cells are not cubic

kstppd:
One way to not continue the run when fsgrid cells are not cubic.


ykempf:
Do we actually want to bailout? It's not wrong in itself to run with non-cubic. Though no one has gone through fully systematically to check we can run completely sanely, I think we're supporting it, or almost fully.

kstppd:
> Do we actually want to bailout? It's not wrong in itself to run with non-cubic. Though no one has gone through fully systematically to check we can run completely sanely, I think we're supporting it, or almost fully.

Posting here as well:

I think it is plain wrong. I believe there are many places in the code where ```dx==dy==dz``` is considered true. Of course up to you

PR #784, opened 2023-06-08
Remove projections from ionosphere and conductingsphere perturbed B computations

ykempf:
So this is the working state. I guess we can even ditch the whole projection functions now that no boundary is using that?

### DIFF
--- a/sysboundary/copysphere.cpp
+++ b/sysboundary/copysphere.cpp
@@ -20,14 +20,14 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-/*!\file conductingsphere.cpp
- * \brief Implementation of the class SysBoundaryCondition::Conductingsphere to handle cells classified as sysboundarytype::CONDUCTINGSPHERE.
+/*!\file copysphere.cpp
+ * \brief Implementation of the class SysBoundaryCondition::Copysphere to handle cells classified as sysboundarytype::COPYSPHERE.
  */
 
 #include <cstdlib>
 #include <iostream>
 
-#include "conductingsphere.h"
+#include "copysphere.h"
 #include "../projects/project.h"
 #include "../projects/projects_common.h"
 #include "../vlasovmover.h"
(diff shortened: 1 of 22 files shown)


ykempf:
If we agree on this, shall we ditch the projection function altogether? Also, shall we rename conductingsphere to... something else?

ykempf:
That might break the testpackage diffing? Let's see.

sysboundary/copysphere.cpp
```diff
@@ -40,65 +40,65 @@
...
-         Readparameters::add(pop + "_conductingsphere.VY0", "Bulk velocity of conductospheric distribution function in X direction (m/s)", 0.0);
-         Readparameters::add(pop + "_conductingsphere.VZ0", "Bulk velocity of conductospheric distribution function in X direction (m/s)", 0.0);
-         Readparameters::add(pop + "_conductingsphere.fluffiness", "Inertia of boundary smoothing when copying neighbour's moments and velocity distributions (0=completely constant boundaries, 1=neighbours are interpolated immediately).", 0);
+         Readparameters::add(pop + "_copysphere.rho", "Number density of the copysphere (m^-3)", 0.0);
+         Readparameters::add(pop + "_copysphere.T", "Temperature of the copysphere (K)", 0.0);
+         Readparameters::add(pop + "_copysphere.VX0", "Bulk velocity of conductospheric distribution function in X direction (m/s)", 0.0);
+         Readparameters::add(pop + "_copysphere.VY0", "Bulk velocity of conductospheric distribution function in X direction (m/s)", 0.0);
+         Readparameters::add(pop + "_copysphere.VZ0", "Bulk velocity of conductospheric distribution function in X direction (m/s)", 0.0);
```
ursg, line 71: These are still saying "conductospheric"

ykempf:
We're good with this one I suppose?

ursg (approved):
Yup

markusbattarbee:
Diffs exist in conducting/copysphere simulations but are expected due to a change to the boundary condition handling. Can't find any references to conductingsphere via grepping the branch. I guess it's good indeed.

PR #785, opened 2023-06-09
add "BOUNDARY_PADDING" boundarytype, for non-coupled DO_NOT_COMPUTE.

ursg:
This type only gets assigned to fsgrid cells that are left outside the actual simulation domain after dccrg refinement. They should not be computed or coupled into dccrg, as not to distort the volume averages.

Not that there is special treatment near the ionosphere, where *do* want to couple fsgrid cells.

### DIFF
--- a/fieldsolver/derivatives.cpp
+++ b/fieldsolver/derivatives.cpp
@@ -373,7 +373,9 @@ void calculateDerivativesSimple(
       for (int j=0; j<gridDims[1]; j++) {
          for (int i=0; i<gridDims[0]; i++) {
             if (technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE ||
-                 technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::BOUNDARY_PADDING) continue;
+                 technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::BOUNDARY_PADDING) {
+               continue;
+            }
             if (RKCase == RK_ORDER1 || RKCase == RK_ORDER2_STEP2) {
                calculateDerivatives(i,j,k, perBGrid, momentsGrid, dPerBGrid, dMomentsGrid, technicalGrid, sysBoundaries, RKCase);
             } else {
--- a/fieldsolver/derivatives.cpp
+++ b/fieldsolver/derivatives.cpp
@@ -372,7 +372,8 @@ void calculateDerivativesSimple(
    for (int k=0; k<gridDims[2]; k++) {
       for (int j=0; j<gridDims[1]; j++) {
          for (int i=0; i<gridDims[0]; i++) {
-            if (technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE) continue;
+            if (technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE ||
+                 technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::BOUNDARY_PADDING) continue;
             if (RKCase == RK_ORDER1 || RKCase == RK_ORDER2_STEP2) {
                calculateDerivatives(i,j,k, perBGrid, momentsGrid, dPerBGrid, dMomentsGrid, technicalGrid, sysBoundaries, RKCase);
             } else {
@@ -513,7 +514,8 @@ void calculateBVOLDerivativesSimple(
    for (int k=0; k<gridDims[2]; k++) {
       for (int j=0; j<gridDims[1]; j++) {
          for (int i=0; i<gridDims[0]; i++) {
-            if (technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE) {
+            if (technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE ||
+                 technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::BOUNDARY_PADDING) {
                continue;
             }
             calculateBVOLDerivatives(volGrid,technicalGrid,i,j,k,sysBoundaries);
@@ -658,7 +660,8 @@ void calculateCurvatureSimple(
    for (int k=0; k<gridDims[2]; k++) {
       for (int j=0; j<gridDims[1]; j++) {
          for (int i=0; i<gridDims[0]; i++) {
-            if (technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE) {
+            if (technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE ||
+                 technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::BOUNDARY_PADDING) {
                continue;
             }
             calculateCurvature(volGrid,bgbGrid,technicalGrid,i,j,k,sysBoundaries);
--- a/fieldsolver/derivatives.cpp
+++ b/fieldsolver/derivatives.cpp
@@ -373,7 +373,7 @@ void calculateDerivativesSimple(
       for (int j=0; j<gridDims[1]; j++) {
          for (int i=0; i<gridDims[0]; i++) {
             if (technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE ||
-                 technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::BOUNDARY_PADDING) {
+                 technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::OUTER_BOUNDARY_PADDING) {
                continue;
             }
             if (RKCase == RK_ORDER1 || RKCase == RK_ORDER2_STEP2) {
@@ -517,7 +517,7 @@ void calculateBVOLDerivativesSimple(
       for (int j=0; j<gridDims[1]; j++) {
          for (int i=0; i<gridDims[0]; i++) {
             if (technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE ||
-                 technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::BOUNDARY_PADDING) {
+                 technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::OUTER_BOUNDARY_PADDING) {
                continue;
             }
             calculateBVOLDerivatives(volGrid,technicalGrid,i,j,k,sysBoundaries);
@@ -663,7 +663,7 @@ void calculateCurvatureSimple(
       for (int j=0; j<gridDims[1]; j++) {
          for (int i=0; i<gridDims[0]; i++) {
             if (technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE ||
-                 technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::BOUNDARY_PADDING) {
+                 technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::OUTER_BOUNDARY_PADDING) {
                continue;
             }
             calculateCurvature(volGrid,bgbGrid,technicalGrid,i,j,k,sysBoundaries);
--- a/fieldsolver/gridGlue.cpp
+++ b/fieldsolver/gridGlue.cpp
@@ -390,6 +390,13 @@ void getFieldsFromFsGrid(
          for (auto const fsgridCell: fsgridCells){
             //loop over fsgrid cells for which we compute the average that is sent to dccrgCell on rank remoteRank
             if(technicalGrid.get(fsgridCell)->sysBoundaryFlag == sysboundarytype::BOUNDARY_PADDING) {
+               // We skip boundary padding cells on the outer boundaries here,
+               // because their fields anyway don't contribute anything
+               // meaningful (as there are never properly updated).
+               //
+               // Note we do *NOT* skip DO_NOT_COMPUTE cells, because we need
+               // the bg vol fields to contribute to the innermost simulation
+               // cell's DCCRG volume averages.
                continue;
             }
             std::array<Real, fsgrids::volfields::N_VOL> * volcell = volumeFieldsGrid.get(fsgridCell);
--- a/fieldsolver/gridGlue.cpp
+++ b/fieldsolver/gridGlue.cpp
@@ -388,10 +388,10 @@ void getFieldsFromFsGrid(
          //loop over dccrg cells to which we shall send data for this remoteRank
          auto const &fsgridCells = onFsgridMapCells[dccrgCell];
          for (auto const fsgridCell: fsgridCells){
-         //loop over fsgrid cells for which we compute the average that is sent to dccrgCell on rank remoteRank
-//        if(technicalGrid.get(fsgridCell)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE) {
-//           continue;
-//        }
+            //loop over fsgrid cells for which we compute the average that is sent to dccrgCell on rank remoteRank
+            if(technicalGrid.get(fsgridCell)->sysBoundaryFlag == sysboundarytype::BOUNDARY_PADDING) {
+               continue;
+            }
             std::array<Real, fsgrids::volfields::N_VOL> * volcell = volumeFieldsGrid.get(fsgridCell);
             std::array<Real, fsgrids::bgbfield::N_BGB> * bgcell = BgBGrid.get(fsgridCell);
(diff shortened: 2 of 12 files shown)


sysboundary/sysboundary.cpp
```diff
@@ -636,8 +636,11 @@ bool SysBoundary::classifyCells(dccrg::Dccrg<spatial_cell::SpatialCell, dccrg::C
...
 
-                     if (layer > 2 && technicalGrid.get(x, y, z)->sysBoundaryFlag != sysboundarytype::NOT_SYSBOUNDARY) {
+                     if (layer > 2 && (technicalGrid.get(x,y,z)->sysBoundaryFlag == sysboundarytype::IONOSPHERE || 
+                                       technicalGrid.get(x,y,z)->sysBoundaryFlag == sysboundarytype::CONDUCTINGSPHERE)) {
                         technicalGrid.get(x, y, z)->sysBoundaryFlag = sysboundarytype::DO_NOT_COMPUTE;
+                     } else if (layer > 2 && technicalGrid.get(x, y, z)->sysBoundaryFlag != sysboundarytype::NOT_SYSBOUNDARY) {
+                        technicalGrid.get(x, y, z)->sysBoundaryFlag = sysboundarytype::BOUNDARY_PADDING;
                      }
                   }
                }
```
markusbattarbee, line 646: I recommend adding a comment to line 647/650 that the ghost update is required becaues belongsToLayer() requires up-to-date ghost information.

Also a comment to line 660/663 that because the layers count is increased so much for FSgrid cells, outer boundary cells should all be handled by the outer loop. Do they? Or we could add the PADDING setting there as well just to be sure.

markusbattarbee (changes requested):
Found also FSgrid accesses in sysboundarycondition.cpp L886
`   Real SysBoundaryCondition::fieldBoundaryCopyFromSolvingNbrMagneticField()` where the padding type should I think be added

fieldsolver/ldz_volume.cpp
```diff
@@ -47,7 +47,8 @@ void calculateVolumeAveragedFields(
    for (int k=0; k<gridDims[2]; k++) {
       for (int j=0; j<gridDims[1]; j++) {
          for (int i=0; i<gridDims[0]; i++) {
-            if(technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE) continue;
```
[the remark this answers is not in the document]
ykempf: Well no as there's an encounter of the *third* type (BOUNDARY_PADDING) if I got this right?

fieldsolver/gridGlue.cpp
```diff
@@ -388,10 +388,10 @@ void getFieldsFromFsGrid(
          //loop over dccrg cells to which we shall send data for this remoteRank
          auto const &fsgridCells = onFsgridMapCells[dccrgCell];
          for (auto const fsgridCell: fsgridCells){
-         //loop over fsgrid cells for which we compute the average that is sent to dccrgCell on rank remoteRank
-//        if(technicalGrid.get(fsgridCell)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE) {
-//           continue;
-//        }
+            //loop over fsgrid cells for which we compute the average that is sent to dccrgCell on rank remoteRank
+            if(technicalGrid.get(fsgridCell)->sysBoundaryFlag == sysboundarytype::BOUNDARY_PADDING) {
```
ykempf, line 392: Well this is exactly wrong, isn't it? You want to include padding, exclude `DO_NOT_COMPUTE`.

fieldsolver/gridGlue.hpp
```diff
@@ -168,7 +168,8 @@ template< unsigned int numFields > void getFieldDataFromFsGrid(
       int nCellsToSum = 0;
       
       for(int iCell = 0; iCell < nCells; ++iCell) {
-         if ((transferBufferPointerTechnical[i] + iCell)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE) {
+         if ((transferBufferPointerTechnical[i] + iCell)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE ||
+              (transferBufferPointerTechnical[i] + iCell)->sysBoundaryFlag == sysboundarytype::BOUNDARY_PADDING) {
```
ykempf, line 172: And this is also wrong, methinks, you *want* to have the `BOUNDARY_PADDING` cells in the averaging.

sysboundary/sysboundary.cpp
```diff
@@ -636,8 +636,11 @@ bool SysBoundary::classifyCells(dccrg::Dccrg<spatial_cell::SpatialCell, dccrg::C
 
                      technicalGrid.get(x, y, z)->sysBoundaryLayer = layer;
 
-                     if (layer > 2 && technicalGrid.get(x, y, z)->sysBoundaryFlag != sysboundarytype::NOT_SYSBOUNDARY) {
+                     if (layer > 2 && (technicalGrid.get(x,y,z)->sysBoundaryFlag == sysboundarytype::IONOSPHERE || 
+                                       technicalGrid.get(x,y,z)->sysBoundaryFlag == sysboundarytype::CONDUCTINGSPHERE)) {
                         technicalGrid.get(x, y, z)->sysBoundaryFlag = sysboundarytype::DO_NOT_COMPUTE;
+                     } else if (layer > 2 && technicalGrid.get(x, y, z)->sysBoundaryFlag != sysboundarytype::NOT_SYSBOUNDARY) {
+                        technicalGrid.get(x, y, z)->sysBoundaryFlag = sysboundarytype::BOUNDARY_PADDING;
```
ykempf, line 643: I think I disagree with this logic.
At this stage (just before the loop) we have got the DCCRG types fed into `technicalGrid` at L514, so we can just use that and be done. No need for padding, upkeep the cell type and make sure the rest of the code is always checking layer as well as boundary type if we don't want to process cells that have boundary type && layer > 2.

ykempf: We just need to move L514 feeding to L590 after the mpiGrid update which includes the DO_NOT_COMPUTE assignments on the inside of the inner boundary.

ykempf (changes requested):
I do not approve, see comments, unless it's a Monday effect I believe this is not doing what we want in the way we want.

fieldsolver/gridGlue.cpp
markusbattarbee: That depends on whether we decide the PADDING type to be the one at the ionosphere or the one at the outer boundaries. I understood that PADDING is at the outer boundaries. And sysboundary.cpp L639 Urs' version puts PADDING at the outer boundaries.

sysboundary/sysboundary.cpp
ykempf: Or well indeed there was the original thing with the upstream wall – so move 514 to 590, and add a loop or modify the last loop, cells of type IONOSPHERE or CONDUCTINGSPHERE that have layer > 2 become PADDING and the other boundary types that are layer > 2 become DO_NOT_COMPUTE? I feel the confusion level rising...

ursg:
> Found also FSgrid accesses in sysboundarycondition.cpp L886 ` Real SysBoundaryCondition::fieldBoundaryCopyFromSolvingNbrMagneticField()` where the padding type should I think be added

True. fixed in c17e2821.

fieldsolver/gridGlue.cpp
```diff
@@ -388,10 +388,10 @@ void getFieldsFromFsGrid(
          //loop over dccrg cells to which we shall send data for this remoteRank
          auto const &fsgridCells = onFsgridMapCells[dccrgCell];
          for (auto const fsgridCell: fsgridCells){
-         //loop over fsgrid cells for which we compute the average that is sent to dccrgCell on rank remoteRank
-//        if(technicalGrid.get(fsgridCell)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE) {
-//           continue;
-//        }
+            //loop over fsgrid cells for which we compute the average that is sent to dccrgCell on rank remoteRank
```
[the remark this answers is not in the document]
ursg: Done in e77d24aa

fieldsolver/ldz_volume.cpp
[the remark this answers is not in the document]
ursg: Hmm, I agree in principle. But I am confused by the fact that we were [already not doing them before](https://github.com/fmihpc/vlasiator/blob/dev/fieldsolver/ldz_volume.cpp#L50).
So what was it we were averaging into the inner boundary edge cells, if not a volume field we calculated here?

fieldsolver/gridGlue.hpp
```diff
@@ -168,7 +168,8 @@ template< unsigned int numFields > void getFieldDataFromFsGrid(
       int nCellsToSum = 0;
       
       for(int iCell = 0; iCell < nCells; ++iCell) {
-         if ((transferBufferPointerTechnical[i] + iCell)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE) {
```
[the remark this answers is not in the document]
ursg: Right, it is unused, and I guess it should be removed as part of #775.

fieldsolver/gridGlue.hpp
ursg: Right, this doesn't make a lot of sense, neither in its current, nor in the previous version. But it seems this function never gets called anyway, so I suggest just chopping the whole function away.

fieldsolver/ldz_volume.cpp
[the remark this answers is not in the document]
ykempf: It's mostly the BGBVOL of course that we're interested in/which dominates. And there is anyway no perB nor perBvol in DNC or padding cells.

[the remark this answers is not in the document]
ykempf: And TBH even if some neighbour might have some non-zero derivatives, I wouldn't want to use those to compute a perBvol – perfectly OK imo to not compute any perBvol outside of boundaries.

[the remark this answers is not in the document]
markusbattarbee: So this function only looks at perBvol. Which indeed should be zero in DNC and boundary padding cells. Now we just need to find where BGBvol is calculated... :)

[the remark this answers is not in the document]
ykempf: Well that's done everywhere by the backgroundfield code, or we wouldn't have got the right stuff from this previous fix.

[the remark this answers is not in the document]
ykempf: https://github.com/fmihpc/vlasiator/blob/dev/backgroundfield/backgroundfield.cpp#L117
And no skipping in the parent loop over fsgrid cells.

markusbattarbee:
> > Found also FSgrid accesses in sysboundarycondition.cpp L886 ` Real SysBoundaryCondition::fieldBoundaryCopyFromSolvingNbrMagneticField()` where the padding type should I think be added
> 
> True. fixed in [c17e282](https://github.com/fmihpc/vlasiator/commit/c17e2821b899393583d2fa06a545eb97a95c29e8).

You also need to add it to the second loop...

sysboundary/sysboundary.cpp
ursg: I think I'm just going to build a small testcase first, to see if this doesn't already do what we want now.

ursg:
> You also need to add it to the second loop...

Whoops. Yep. Fixed in 77510dae.

fieldsolver/derivatives.cpp
```diff
@@ -372,7 +372,10 @@ void calculateDerivativesSimple(
    for (int k=0; k<gridDims[2]; k++) {
       for (int j=0; j<gridDims[1]; j++) {
          for (int i=0; i<gridDims[0]; i++) {
-            if (technicalGrid.get(i,j,k)->sysBoundaryFlag == sysboundarytype::DO_NOT_COMPUTE) continue;
```
markusbattarbee, line 375: So question: Is there any harm in calculating the derivatives for these cells as well? Inside the inner boundary they should be zero, but near the outer boundaries they might be nonzero. And then if we want to evaluate volumetric fields correctly there, we might need them. Removing a per-cell check could allow more vectorization, with very little actual cost?

Ditto for lines 519 and 665.

fieldsolver/gridGlue.cpp
```diff
@@ -388,10 +388,17 @@ void getFieldsFromFsGrid(
...
+            if(technicalGrid.get(fsgridCell)->sysBoundaryFlag == sysboundarytype::OUTER_BOUNDARY_PADDING) {
```
markusbattarbee, line 392: Oh dear. Another can-of-worms-idea. How will this impact time-varying boundaries? Should we in fact ensure the fields get updated and communicated there as well, so that the volB grows in sync with the actual boundary?

markusbattarbee (commented):
Some more thoughts for discussion

fieldsolver/derivatives.cpp
ykempf: We won't be using them though, right?

fieldsolver/gridGlue.cpp
ykempf: They will be as we only skip padding here?

fieldsolver/derivatives.cpp
markusbattarbee: Well, in constructing a volumetric value, there's a function `reconstructionCoefficients` which actually can use the derivative from the neighbouring cell. So... I'm not sure?

fieldsolver/gridGlue.cpp
ursg: Yep, I think so too. Isn't this already handled properly with the introduction of padding?

fieldsolver/derivatives.cpp
ursg: This brings us to the uncomfortable question: Is there *any* mechanism in the code that could propagate information from these cells into the domain proper? Because I had assumed that no (so they are really just dead ballast, and can be ignored completely).

ykempf: No, reconstructionCoefficients is purely local?

ykempf: So I think in 2D in L2 boundary cells (outer edge of the wall) we probably set derivatives to zero so we don't access out of domain.
In 3D now we thus need to compute derivatives in the DNC/PADDING regions just in case, but only out to the penultimate layer? That would be sufficient once at startup, right? Until we go time-varying at inflow walls, there I suppose ideally we need the proper inflowing spatial profile/derivatives?? Ouch.

markusbattarbee: Interestingly that `reconstructionCoefficients` https://github.com/fmihpc/vlasiator/blob/ae70882b72cd22ea9b8bcccca170757c1efaade0/fieldsolver/fs_common.cpp#L95 has a built-in way of managing right-at-the-edge cells, so it can also calculate derivatives there, it seems.

ykempf:
So how's this coming along? We can't start FIB without this one.

fieldsolver/derivatives.cpp
ykempf: So I do indeed agree that we should be able to compute derivatives throughout. And indeed we introduce the effect that the hinterland of the (inflow) boundary thus affects the volume averaged fields at the boundary (*not* inside-cells at least as long as the refinement is uniform along the wall, and the boundary is planar). Though this also means affecting at copy-boundaries, where the hinterland will also have initial values we don't want to keep affecting. That is in essence the same issue as with the time-varying inflow.
Therefore: I would advocate for zero-derivatives in DO_NOT_COMPUTE and OUTER_BOUNDARY_PADDING so we don't instead have "nastier" derivatives affecting us.

ykempf: Except that indeed as Markus was pointing out to me, we need to upkeep derivatives because of the "compensation term" in type 4 setups (vector potential modified dipole) storing a significant part in the perturbed field.

ykempf: However if we blindly remove this check, we'll run off the domain edges. We shouldn't call `calculateDerivatives()` for i,j,k == 0 or at the other end.

markusbattarbee: Yep. Perhaps we could replicate the `reconstructionCoefficients` approach there?

fieldsolver/gridGlue.cpp
markusbattarbee: Yeah I guess it'll be fine.

sysboundary/sysboundary.cpp
markusbattarbee: I had for a bit wondered why this only handles ionosphere/copysphere, and not the outer wall boundary types, but the reason is that any DNC/BP cells at the outer walls are already handled by the previous loop. And that is exactly because the value `MAX_NUMBER_OF_BOUNDARY_LAYERS` is so large for fsgrid.

sysboundary/sysboundary.cpp
ykempf: I'll resolve this conversation, we ran several tests and my issues with the logic were that I had the opposite concept of which cells to flag differently. Works well like so too.

fieldsolver/derivatives.cpp
ykempf: Did that in my branch, will be put online soon.

ykempf:
I branched off to my fork to test some of the things discussed above and then merge into an FIB branch.

ursg:
This has been completely absorbed in, and superseded by, #791

PR #786, opened 2023-06-13
Write out full bgb data x make differences derivatives

ykempf:
This combines #711 and #727 to make sure they are consistent.

### DIFF
--- a/common.h
+++ b/common.h
@@ -446,7 +446,7 @@ namespace sysboundarytype {
       IONOSPHERE,       /*!< Ionospheric current model */
       OUTFLOW,          /*!< No fixed conditions on the fields and distribution function. */
       SET_MAXWELLIAN,   /*!< Set Maxwellian boundary condition, i.e. set fields and distribution function. */
-      CONDUCTINGSPHERE, /*!< A perfectly conducting sphere as the simple inner boundary */
+      COPYSPHERE,       /*!< A sphere with copy-condition for perturbed B as the simple inner boundary */
       N_SYSBOUNDARY_CONDITIONS
    };
 }
--- a/common.h
+++ b/common.h
@@ -368,12 +368,15 @@ namespace fsgrids {
       dBGBydz,     /*!< Derivative of background face-averaged By in z-direction. */
       dBGBzdx,     /*!< Derivative of background face-averaged Bz in x-direction. */
       dBGBzdy,     /*!< Derivative of background face-averaged Bz in y-direction. */
+      dBGBXVOLdx,  /*!< Derivative of background volume-averaged Bx in x-direction. */
       dBGBXVOLdy,  /*!< Derivative of background volume-averaged Bx in y-direction. */
       dBGBXVOLdz,  /*!< Derivative of background volume-averaged Bx in z-direction. */
       dBGBYVOLdx,  /*!< Derivative of background volume-averaged By in x-direction. */
+      dBGBYVOLdy,  /*!< Derivative of background volume-averaged By in y-direction. */
       dBGBYVOLdz,  /*!< Derivative of background volume-averaged By in z-direction. */
       dBGBZVOLdx,  /*!< Derivative of background volume-averaged Bz in x-direction. */
       dBGBZVOLdy,  /*!< Derivative of background volume-averaged Bz in y-direction. */
+      dBGBZVOLdz,  /*!< Derivative of background volume-averaged Bz in z-direction. */
       N_BGB
    };
    
--- a/datareduction/datareducer.cpp
+++ b/datareduction/datareducer.cpp
@@ -41,8 +41,8 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
         it++) {
 
       /* Note: Each data reducer generation should be followed by a call to setUnitMetaData
-	 with the following arguments:
-	 unit, unit in LaTeX formulation, variable in LaTeX formulation, conversion factor
+         with the following arguments:
+         unit, unit in LaTeX formulation, variable in LaTeX formulation, conversion factor
       */
 
       // Sidestep mixed case errors
@@ -79,9 +79,9 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
                   }
                }
                return retval;
-	 }
-	 ));
-         outputReducer->addMetadata(outputReducer->size()-1,"T","$\\mathrm{T}$","$B$","1.0");
+            }
+         ));
+         outputReducer->addMetadata(outputReducer->size()-1,"T","$\\mathrm{T}$","$B_\\mathrm{fg}$","1.0");
          if(!P::systemWriteAllDROs) {
             continue;
          }
@@ -115,11 +115,44 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
                return retval;
          }
          ));
-	 outputReducer->addMetadata(outputReducer->size()-1,"T","$\\mathrm{T}$","$B_\\mathrm{bg}$","1.0");
+         outputReducer->addMetadata(outputReducer->size() - 1, "T", "$\\mathrm{T}$", "$B_\\mathrm{bg,fg}$", "1.0");
          if(!P::systemWriteAllDROs) {
             continue;
          }
       }
+      if(P::systemWriteAllDROs || lowercase == "fg_backgroundbvol" || lowercase == "backgroundbvol" || lowercase == "fg_b_background_vol") { // Static (typically dipole) magnetic field part, volume-averaged
+         outputReducer->addOperator(new DRO::DataReductionOperatorFsGrid("fg_b_background_vol",[](
+            FsGrid< std::array<Real, fsgrids::bfield::N_BFIELD>, FS_STENCIL_WIDTH> & perBGrid,
+            FsGrid< std::array<Real, fsgrids::efield::N_EFIELD>, FS_STENCIL_WIDTH> & EGrid,
+            FsGrid< std::array<Real, fsgrids::ehall::N_EHALL>, FS_STENCIL_WIDTH> & EHallGrid,
+            FsGrid< std::array<Real, fsgrids::egradpe::N_EGRADPE>, FS_STENCIL_WIDTH> & EGradPeGrid,
+            FsGrid< std::array<Real, fsgrids::moments::N_MOMENTS>, FS_STENCIL_WIDTH> & momentsGrid,
+            FsGrid< std::array<Real, fsgrids::dperb::N_DPERB>, FS_STENCIL_WIDTH> & dPerBGrid,
+            FsGrid< std::array<Real, fsgrids::dmoments::N_DMOMENTS>, FS_STENCIL_WIDTH> & dMomentsGrid,
+            FsGrid< std::array<Real, fsgrids::bgbfield::N_BGB>, FS_STENCIL_WIDTH> & BgBGrid,
+            FsGrid< std::array<Real, fsgrids::volfields::N_VOL>, FS_STENCIL_WIDTH> & volGrid,
+            FsGrid< fsgrids::technical, FS_STENCIL_WIDTH> & technicalGrid)->std::vector<double> {
+
+               std::array<int32_t,3>& gridSize = technicalGrid.getLocalSize();
+               std::vector<double> retval(gridSize[0]*gridSize[1]*gridSize[2]*3);
+
+               // Iterate through fsgrid cells and extract total BVOL
+               for(int z=0; z<gridSize[2]; z++) {
+                  for(int y=0; y<gridSize[1]; y++) {
+                     for(int x=0; x<gridSize[0]; x++) {
+                        retval[3*(gridSize[1]*gridSize[0]*z + gridSize[0]*y + x)] =     (*BgBGrid.get(x,y,z))[fsgrids::BGBXVOL];
+                        retval[3*(gridSize[1]*gridSize[0]*z + gridSize[0]*y + x) + 1] = (*BgBGrid.get(x,y,z))[fsgrids::BGBYVOL];
+                        retval[3*(gridSize[1]*gridSize[0]*z + gridSize[0]*y + x) + 2] = (*BgBGrid.get(x,y,z))[fsgrids::BGBZVOL];
+                     }
+                  }
+               }
+               return retval;
+            }
+         ));
+         outputReducer->addMetadata(outputReducer->size()-1,"T","$\\mathrm{T}$","$B_\\mathrm{bg,vol,fg}$","1.0");
+         continue;
+      }
+
       if(P::systemWriteAllDROs || lowercase == "fg_perturbedb" || lowercase == "perturbedb" || lowercase == "fg_b_perturbed") { // Fluctuating magnetic field part
          outputReducer->addOperator(new DRO::DataReductionOperatorFsGrid("fg_b_perturbed",[](
                       FsGrid< std::array<Real, fsgrids::bfield::N_BFIELD>, FS_STENCIL_WIDTH> & perBGrid,
@@ -149,7 +182,7 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
                return retval;
          }
          ));
-	 outputReducer->addMetadata(outputReducer->size()-1,"T","$\\mathrm{T}$","$B_\\mathrm{per}$","1.0");
+         outputReducer->addMetadata(outputReducer->size()-1,"T","$\\mathrm{T}$","$B_\\mathrm{per,fg}$","1.0");
          if(!P::systemWriteAllDROs) {
             continue;
          }
@@ -183,7 +216,7 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
                return retval;
          }
          ));
-	 outputReducer->addMetadata(outputReducer->size()-1,"V/m","$\\mathrm{V}\\,\\mathrm{m}^{-1}$","$E$","1.0");
+         outputReducer->addMetadata(outputReducer->size()-1,"V/m","$\\mathrm{V}\\,\\mathrm{m}^{-1}$","$E$","1.0");
          if(!P::systemWriteAllDROs) {
             continue;
          }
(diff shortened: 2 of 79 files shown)


datareduction/datareducer.cpp
```diff
@@ -35,6 +35,575 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
+               for(int y=0; y<gridSize[1]; y++) {
+                  for(int x=0; x<gridSize[0]; x++) {
+                     retval[gridSize[1]*gridSize[0]*z + gridSize[0]*y + x] = BgBGrid.get(x,y,z)->at(fsgrids::bgbfield::dBGBxdx) / technicalGrid.DX;
+                  }
+               }
+            }
+            return retval;
+         }
+      ));
+      outputReducer->addMetadata(outputReducer->size()-1,"T/m","$\\mathrm{T}\\,\\mathrm{m}^{-1}$","$B_{X,\\mathrm{bg,fg}} (\\Delta X)^{-1}$","1.0");
```
alhom, line 125: "$\\Delta B_{X,\\mathrm{bg,fg}} (\\Delta X)^{-1}$", idem below

ykempf: Good now, did I catch'em all?

iowrite.cpp
```diff
@@ -1453,12 +1452,14 @@ bool writeGrid(
...
       return false;
    }
-
-   //Write Ionosphere Grid
-   if( writeIonosphereGridMetadata( vlsvWriter ) == false ) {
-      phiprof::stop("writeGrid-reduced");
-      phiprof::stop("metadataIO");
-      return false;
+   
+   if( !P::writeFullBGB ) {
```
ursg, line 1456: This one might need a little explanatory comment. Why *wouldn't* we also write tthe ionosphere metadata here?

ursg: (Because we really only use it do write *only* BGB, but that isn't obvious from the code here)

fieldsolver/ldz_hall.cpp
```diff
@@ -460,10 +460,10 @@ void calculateEdgeHallTermXComponents(
...
-      EXHall = (Bz*((BgBGrid.get(i,j,k)->at(fsgrids::bgbfield::dBGBxdz)+dPerBGrid.get(i,j,k)->at(fsgrids::dperb::dPERBxdz))/technicalGrid.DZ -
-                     (BgBGrid.get(i,j,k)->at(fsgrids::bgbfield::dBGBzdx)+dPerBGrid.get(i,j,k)->at(fsgrids::dperb::dPERBzdx))/technicalGrid.DX) -
-                  By*((BgBGrid.get(i,j,k)->at(fsgrids::bgbfield::dBGBydx)+dPerBGrid.get(i,j,k)->at(fsgrids::dperb::dPERBydx))/technicalGrid.DX-
-                     ((BgBGrid.get(i,j,k)->at(fsgrids::bgbfield::dBGBxdy)+dPerBGrid.get(i,j,k)->at(fsgrids::dperb::dPERBxdy))/technicalGrid.DY)));
+      EXHall = Bz*((BgBGrid.get(i,j,k)->at(fsgrids::bgbfield::dBGBxdz)+dPerBGrid.get(i,j,k)->at(fsgrids::dperb::dPERBxdz)) / technicalGrid.DZ -
+                  (BgBGrid.get(i,j,k)->at(fsgrids::bgbfield::dBGBzdx)+dPerBGrid.get(i,j,k)->at(fsgrids::dperb::dPERBzdx)) / technicalGrid.DX) -
+               By*((BgBGrid.get(i,j,k)->at(fsgrids::bgbfield::dBGBydx)+dPerBGrid.get(i,j,k)->at(fsgrids::dperb::dPERBydx)) / technicalGrid.DX -
```
ursg, line 465: These formatting and parenthesation changes look scary, but they do look correct, so :+1:

ursg (approved):
Thumbs up for this. I guess this will somewhat collide with PR #787, but those conflicts should be straightforward to fix.

fieldsolver/ldz_hall.cpp
ykempf: Yeah I went through when attempting to make them derivatives thorughout the code, and while here I noticed there were too many parenthesis pairs so I "cleaned up" even though it's still scary.

iowrite.cpp
ykempf: I *think* this was so early in now that the ionosphere metadata wasn't actually properly there. Do you want me to test this?

ykempf:
The ionosphere test is sensitive. Here I assume that we got one or more cells that were (not) coupled in (or ditched), probably/hopefully due to numerics, and that shows up as a O(10¹⁰ m²) diff in upmapped area. Bummer.

ykempf:
For the record here: upon closer inspection I see that the diffs of order 10¹⁰ are actually relative O(10⁻⁴) and the diffs in the coordinates mapped are up to maybe 15 km, with the dx being 7500 km in Ionosphere_small. So this comes from differences in the tracing and maybe from the many iterations of the ionosphere solver (every step). The actual upmapped area is identical to the eye and to 4 digits in VisIt.

grid.cpp
```diff
@@ -266,7 +266,9 @@ void initializeGrids(
    // Update technicalGrid
    technicalGrid.updateGhostCells(); // This needs to be done at some point
 
-   if (!P::isRestart) {
+   if (!P::isRestart && !P::writeFullBGB) {
```
markusbattarbee, line 269: I can't very easily brain this triple-negative with an exception. Suggestion for a new comment:

If we are starting a new regular simulation, we need to prepare all cells with their initial state.

grid.cpp
```diff
@@ -329,8 +331,13 @@ void initializeGrids(
       phiprof::stop("Init moments");
       */
 
+   } else if (P::writeFullBGB) {
```
markusbattarbee, line 334: Suggestion for new comment:

If, instead of starting a regular simulation, we are only writing out the background field, it is enough to set a dummy load balance value of 1 here.

vlasiator.cpp
```diff
@@ -331,9 +331,17 @@ int main(int argn,char* args[]) {
...
-   if (logFile.open(MPI_COMM_WORLD,MASTER_RANK,"logfile.txt",P::isRestart) == false) {
-      if(myRank == MASTER_RANK) cerr << "(MAIN) ERROR: Logger failed to open logfile!" << endl;
-      exit(1);
+   if(!P::writeFullBGB) {
+      if (logFile.open(MPI_COMM_WORLD,MASTER_RANK,"logfile.txt",P::isRestart) == false) {
+         if(myRank == MASTER_RANK) cerr << "(MAIN) ERROR: Logger failed to open logfile!" << endl;
+         exit(1);
+      }
+   } else {
+      // If we are out to write the full background field and derivatives, we don't want to overwrite the exisitng run's logfile.
```
markusbattarbee, line 340: typos here

parameters.h
```diff
@@ -104,6 +104,7 @@ struct Parameters {
 
    static bool writeInitialState; /*!< If true, initial state is written. This is useful for debugging as the restarts
                                      are always written out after propagation of 0.5dt in real space.*/
+   static bool writeFullBGB; /*!< If true, write full BGB components and derivatives in a dedicated file.*/
```
markusbattarbee, line 107: add  ", then exit" to the comment

parameters.cpp
```diff
@@ -230,6 +230,8 @@ bool P::addParameters() {
    RP::add("io.write_initial_state",
            "Write initial state, not even the 0.5 dt propagation is done. Do not use for restarting. ", false);
 
+   RP::add("io.write_full_bgb_data", "Write a dedicated file containing all BGB components and first derivatives.", false);
```
markusbattarbee, line 233: add ", then exit" to the comment

iowrite.cpp
markusbattarbee: Could be tested. And if the ionosphere data isn't there, then let's add a comment along the lines of "If only writing out BGB data, ionosphere metadata isn't valid and should not be output."

vlasiator.cpp
```diff
@@ -469,6 +477,71 @@ int main(int argn,char* args[]) {
    // Free up memory:
    readparameters.~Readparameters();
 
+   if(P::writeFullBGB) {
+      logFile << "Writing out full BGB components and derivatives." << endl << writeVerbose;
```
markusbattarbee, line 481: " and exiting"

datareduction/datareducer.cpp
```diff
@@ -35,6 +35,575 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
 {
    typedef Parameters P;
 
+   if(P::writeFullBGB) {
```
markusbattarbee, line 38: Add a comment: "Special case, only construct all required background magnetic field datareducers, then return" or even more verbose.

datareduction/datareducer.cpp
```diff
@@ -35,6 +35,575 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
+            }
+            return retval;
+         }
+      ));
+      outputReducer->addMetadata(outputReducer->size()-1,"T/m","$\\mathrm{T}\\,\\mathrm{m}^{-1}$","$\\Delta B_{Z,\\mathrm{bg,vol,fg}} (\\Delta Z)^{-1}$","1.0");
+
+      // end of the write full BGB set
+      return;
+   }
+
```
markusbattarbee, line 606: And add comment: "Now we actually prepare the required datareducers for a regular run."

markusbattarbee (changes requested):
Urgh, this datareducer.cpp diff is hideous. 

And, what you've now got is a huge amount of duplication, as it has basically a giant construction at the beginning for this one special case of BGB writing.

Wouldn't it be possible that if one is requesting the bgb writeout, the initialization would replace the list of datareducers read from the config with one provided (which has only the bgb values, derivatives, etc)?

The reason I'm also thinking about this is that merging this with Jaro's much more elegant and re-using labmda approach will be a nightmare, whereas if the bgb writing only tampers with the list of reducers to make, it'll be much more straightforward.

datareduction/datareducer.cpp
ykempf: Will become obsolete with change of approach.

datareduction/datareducer.cpp
ykempf: Will similarly become obsolete.

iowrite.cpp
ykempf: OK initialising the ionosphere communicator instead to get that written.

datareduction/datareducer.cpp
```diff
@@ -79,10 +79,10 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
                   }
                }
                return retval;
-	 }
-	 ));
-         outputReducer->addMetadata(outputReducer->size()-1,"T","$\\mathrm{T}$","$B$","1.0");
-         continue;
+            }
+         ));
+         outputReducer->addMetadata(outputReducer->size()-1,"T","$\\mathrm{T}$","$B_\\mathrm{bg,fg}$","1.0");
```
markusbattarbee, line 84: This one is still all B, not just bgb. Feel free to add the fg descriptor, but the bg,fg pair would go to line 116.

markusbattarbee (commented):
Oh that is so much easier to diff-analyse! Thank you! Still a lot of changes to go through though, I'll try to check them all tomorrow.

Want to check about volumetric d's though: here they aren't being divided by the dx, dy or dz. But looking at derivatives.cpp they also look they are differences, not derivatives?

datareduction/datareducer.cpp
```diff
@@ -2175,256 +2208,510 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
-                      FsGrid< std::array<Real, fsgrids::moments::N_MOMENTS>, FS_STENCIL_WIDTH> & momentsGrid,
-                      FsGrid< std::array<Real, fsgrids::dperb::N_DPERB>, FS_STENCIL_WIDTH> & dPerBGrid,
-                      FsGrid< std::array<Real, fsgrids::dmoments::N_DMOMENTS>, FS_STENCIL_WIDTH> & dMomentsGrid,
-                      FsGrid< std::array<Real, fsgrids::bgbfield::N_BGB>, FS_STENCIL_WIDTH> & BgBGrid,
-                      FsGrid< std::array<Real, fsgrids::volfields::N_VOL>, FS_STENCIL_WIDTH> & volGrid,
-                      FsGrid< fsgrids::technical, FS_STENCIL_WIDTH> & technicalGrid)->std::vector<double> {
+	 outputReducer->addMetadata(outputReducer->size()-1,"T/m","$\\mathrm{T}\\mathrm{m}^{-1}$","$\\Delta B_{Z,\\mathrm{per,vol,fg}} (\\Delta Z)^{-1}$","1.0");
+
+         continue;
+      } // fg_derivs
```
markusbattarbee, line 2224: so a bigger comment here perhaps, signifying end of fg_derivs, and noting that they are only for the perturbed field. And a starting comment for the fg background field derivatives (regular and volumetric).

datareduction/datareducer.cpp
```diff
@@ -2175,256 +2208,510 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
+                  }
+               }
+               return retval;
+            }
+         ));
+         outputReducer->addMetadata(outputReducer->size()-1,"T/m","$\\mathrm{T}\\,\\mathrm{m}^{-1}$","$\\Delta B_{Z,\\mathrm{bg,vol,fg}} (\\Delta Z)^{-1}$","1.0");
 
          continue;
-      }
+      } // fg_derivs_b_background
```
markusbattarbee, line 2714: another big comment line here please to aid readability

datareduction/datareducer.cpp
```diff
@@ -871,17 +904,17 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
                std::array<int32_t,3>& gridSize = technicalGrid.getLocalSize();
                std::vector<double> retval(gridSize[0]*gridSize[1]*gridSize[2]);
 
-               // Iterate through fsgrid cells and extract boundary flag
```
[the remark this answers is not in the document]
markusbattarbee: Also, we have the xdx, ydy, zdz terms for all other fields, should we also thus include fg_dbxdx, fg_dbydy, fg_dbzdz?

markusbattarbee (changes requested):

markusbattarbee:
Oh, one more thing, `"g_b_background_vol` and `fg_derivs_b_background` should be added to the allowed output parameters list, since they are new.

common.h
```diff
@@ -362,18 +362,24 @@ namespace fsgrids {
...
+      dBGBzdz,     /*!< Derivative of background face-averaged Bz in z-direction. */
+      dBGBXVOLdx,  /*!< Derivative of background volume-averaged Bx in x-direction. */
       dBGBXVOLdy,  /*!< Derivative of background volume-averaged Bx in y-direction. */
       dBGBXVOLdz,  /*!< Derivative of background volume-averaged Bx in z-direction. */
       dBGBYVOLdx,  /*!< Derivative of background volume-averaged By in x-direction. */
+      dBGBYVOLdy,  /*!< Derivative of background volume-averaged By in y-direction. */
       dBGBYVOLdz,  /*!< Derivative of background volume-averaged By in z-direction. */
       dBGBZVOLdx,  /*!< Derivative of background volume-averaged Bz in x-direction. */
       dBGBZVOLdy,  /*!< Derivative of background volume-averaged Bz in y-direction. */
+      dBGBZVOLdz,  /*!< Derivative of background volume-averaged Bz in z-direction. */
```
ykempf, line 382: So would you prefer these be inserted at the tail end of this fsgrid instead of interspersed, to improve data access for the ones actually used in the code?

ykempf: Or should we have a separate fsgrid even? Those I think we should let the code compute in `backgroundfield.cpp` and extract them in `datareducer.cpp`, I wouldn't want to have them computed at IO time. Although now that I write this, I suppose it wouldn't be a big issue although then it wouldn't be a lambda-based DRO any more?

datareduction/datareducer.cpp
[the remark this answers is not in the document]
ykempf: So these are not computed anywhere. I suppose we could write a DRO that would compute them at IO, instead of having them computed in the normal `calculateDerivatives()` every step or through some contraption checking whether IO is coming up, and storing them an cluttering more some fsgrid?

ykempf:
Summary of our discussion with Markus, so I remember what to code after lunch:
- Adding some more bells, whistles and hoops to get fg_dperbidi out is inconvenient now, and realistically won't be of much use;
- Thence we can ditch computing and storing fg_dbgbidi components;
- However we'll keep the fg_dbgbivoldi computation, storage and one-off IO as well as their vg averages;
- And I'll repair the git wreck that caused me to actually revert what was computing the fg_dbgbidi to not do it, when we want it, in backgroundfield.cpp.

alhom:
Oh, since there will be a big bunch of outputs now, should we actually use instead of separate `fg_dperbxdz` some construct like `fg_dperbidj/fg_dperbxdz` to group the outputs? Or even store them as 9-component vectors (... which VisIt handles as tensors directly)?

datareduction/datareducer.cpp
ykempf: I added those.

datareduction/datareducer.cpp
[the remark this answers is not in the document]
ykempf: As discussed and summarised here too; now we only compute and write the background volume derivatives, to go along with the perturbed volume derivatives' full jacobian output. more later and/or as part of the GPU rehaul.

ykempf:
Right so this has two bugs in the merge to get all variables written out in `datareducer.cpp`. I have a fix ready to push. These bugs cause what we see in this PR's CI tests under Ionosphere_small, namely that variables were not found except for `fg_b` as that DROs `continue` was not put in an `if` for the option writing all variables.
This however exposes an issue in the reporting (here of `Ionosphere_small`: only `fg_b` `0`, `1` and `2` had output for the diff, yet the report is shouting about `fg_e` having diffs, though it prints the empty value there as no value was detected. I think a test should be reported as properly failed when vlsvdiff doesn't find a variable.
So I have a fix ready to push, I can hold it up until the CI testing is fixed. As discussed on Zulip, the diffs reported don't match those obtained by running the testpackage manually.

ykempf:
Hooray

ykempf:
So, merge this?

PR #787, opened 2023-06-13
Write all DROs

ykempf:
We talked about it during the hackathon last week. Here's a suggestion from me how to bypass the current logic and select all DROs in one go. This is done for all file types, hence the flag making it skip fg variables for select file types has higher precedence and one can get all non-fg or all including fg.

Do we want to include this in one of the testpackage tests? I ran one Magnetosphere_3D_small-based test and it did seem to behave as expected, and unlike what I expected **did not crash**.

### DIFF
--- a/datareduction/datareducer.cpp
+++ b/datareduction/datareducer.cpp
@@ -49,7 +49,7 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
       std::string lowercase = *it;
       for(auto& c : lowercase) c = tolower(c);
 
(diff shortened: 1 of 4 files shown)


ursg:
This is starting to be so much code repetition that we could consider building a template or doing some funky metaprogramming to generate all these ever-same DRO definitions. But that, of course, is not for this PR.

Functionally approved!

ykempf:
Just wait for the CI to finish as I added this onw to the testpackage (though the testpackage is not checking whether that succeeded).

PR #791, opened 2023-07-03
FIB production branch

ykempf:
This is (so far) a merged branch of
- today's `dev`
- `ykempf/write_out_full_bgb_data_X_make_differences_derivatives` (itself a merger of both these branches, see #786)
- `ykempf/boundaryPadding`, an update off `ursg/boundaryPadding`, see #785

which are the intended new features and fixes going into the FIB branch. Not to be merged until after the summer break, but to collect what we're doing in FIB.


ykempf:
So I wanted to post the merge commit so it runs the testpackage hopefully overnight. But it'll definitely need careful checking, not to be merged just yet, I might well have made some blunders in merging the timer changes to the PADDING fixes.

ykempf:
Meh, this also failed to upload the report. Here it is. As expected, there's bigger diffs in the Magnetospheres and Ionosphere due to the rejig with boundary padding.
[testpackage_run_output.txt](https://github.com/fmihpc/vlasiator/files/12722271/testpackage_run_output.txt)

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -1410,6 +1410,8 @@ bool trans_map_1d_amr(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
       return false;
    }
 
+   phiprof::start("setup");
```
ykempf, line 1413: Dude! Who committed this?

ykempf: Anyway, gone.

markusbattarbee (changes requested):
Unused function still lingering here

ykempf:
So this is not merged nor is it in FID. And I assume neither is it in FIC. Should that have been the case? Argh.

sysboundary/setbyuser.cpp
```diff
@@ -102,33 +102,11 @@ namespace SBC {
...
-            result = templateB[i][component];
-            break; // This effectively sets the precedence of faces through the order of faces.
-         }
-      }
-      return result;
+      // There are projects that have non-uniform and non-zero perturbed B, e.g. Magnetosphere with dipole type 4.
+      // We cannot take a value from the templateCell, we need a copy of the value from initialization which has
+      // been set in perBGrid as well as perBDt2Grid and isn't touched as we are in boundary cells for components
+      // that aren't solved.
+      return bGrid.get(i,j,k)->at(fsgrids::bfield::PERBX+component);
```
ykempf, line 109: This file is gone in dev with time-varying BCs now. That's the conflict which is hampering this PR right now. But I'm uncertain how this should be handled now that we have the time-varying in.

markusbattarbee: Rebase this branch onto dev and see what happens?
(I had also completely missed that these fixes weren't already in dev)

markusbattarbee: I now had another look at this: the current setmaxwellian.cpp which declares the maxwellian boundary condition inherits from inflow.cpp (so, if I understand correctly, the old setbyuser became the new inflow). 
This makes a bit more sense now.

However, this opens up yet another new can of worms!

If we have a maxwellian time-varying inflow, where we need to overwrite the local perB values, we would need to have some way to retain the initialization-time perB as well. This is because of the hackyness of type 4 dipoles using perB to store the difference to a regular dipole... ecch.

markusbattarbee: And no, we can't store the increment in the template as the template is uniform over the whole inflow boundary, whereas the $\Delta B$ value is different for each cell as it needs to cancel out the BGB.

ykempf:
I am satisfied this is ported into #866. Even with the diff view of both PRs was made arbitrarily inconsistent by github because why not.

PR #793, opened 2023-07-04
Shorten the max distance for full box tracing.

ykempf:
For production runs we often hit this max tracing distance of >1500 RE in FHA, now shortened to a quarter.

Presumably we'll want to make this a cfg parameter? Up for discussion and inclusion in the FIB_production_branch #791 for now, to be discussed and concluded in August.


ykempf:
Checks successful.

markusbattarbee (approved):

ykempf:
Meanwhile, this is also in FIB_production and was approved, so merge?

PR #795, opened 2023-07-10
Makefile and a small fix to build on jetsonTX Arm cluster at BSC.

ursg:
Note that the type change in ionosphere.cpp is also required to build on RISC-V, so I suppose this is relevant for all architectures with some alignment constraints. Not a performance issue either, so why not just fix it for all.

This has been tested to compile and run at least the Magnetosphere_small test from the testpackage on the cluster in Barcelona.

### DIFF
--- a/MAKE/Makefile.bsc_jetsontx
+++ b/MAKE/Makefile.bsc_jetsontx
@@ -1,0 +1,57 @@
+# Makefile for the jetson-tx cluster at BSC
+#
+# Note:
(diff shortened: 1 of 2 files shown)


ykempf:
So, what about the other tests? Please report, we have been finding pretty old but stinking bugs e.g. with threading of phiprof by switching compilers on other platforms, so any unexpected behaviour on new platforms and compilers should be kept track of!

PR #797, opened 2023-07-12
Ionosphere_small.cfg test: coupling radius > ionosphere radius.

ykempf:


markusbattarbee:
Is it anymore a meaningful ionospheric test if it's being run at almost 11 RE?

ursg:
I mean, it still runs a bunch of ionosphere codepaths for sanity-checking, so it's better than nothing.
But do you have a better suggestion for a "small" ionosphere test?

ykempf:
Anyway running 3 s of an ionosphere won't do much, but this explains why there was only microscopic potentials, nothing is coupled. Can we decrease the dipole strength so we get longer dt and can run a few (tens of?) seconds quickly and actually get the beginning of the steep CPCP increase we've seen in FIA/FIB?

ykempf:
Ah heh I looked at my own diffs again and of course at that coupling radius nothing will couple, we should pull in the inner radius instead.

markusbattarbee:
Also need to allow FS subcycling to get longer timesteps.

markusbattarbee:
I've made a test run at `/wrk-vakka/users/markusb/KTH/Ionosphere_intermediate` with 30 seconds of simulation, FS subcycling, the inner boundary at 25e6 metres (just under 4 RE), and a dipole scaling factor of 0.001. This ran in 756 seconds on Carrington so is still relatively expensive for CI. I'm also a bit surprised that FS subcycles were consistently at the maximum allowed value. Up for discussion and adjustment.

PR #798, opened 2023-08-10
LUMI makefile fixes

lkotipal:
- Fix OpenMP for Cray environment. Using `-lgomp` breaks the function `omp_get_thread_num()` causing errors in phiprof output.
- Use modules for Eigen and Boost.
- Use `-isystem` to avoid warnings from outside libraries.

### DIFF
--- a/MAKE/Makefile.lumi_cray
+++ b/MAKE/Makefile.lumi_cray
@@ -2,7 +2,7 @@ CMP = CC
 LNK = CC
 
 # Modules loaded
-# module --force purge ; module load LUMI/22.08 ; module load cpeCray ; module load papi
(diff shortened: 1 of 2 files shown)


ykempf:
So in some of the last messages I exchanged with LUMI support in July it seemed that some issues I was seeing/reporting were probably due to my stubborn use of `module --force purge`. Should we/can you test a set of module commands to use from a pristine module environment, and add that in the Makefiles next to or instead of my purge-based ones?

lkotipal:
So that's just the same module commands without purge, right? Running right now, I expect it to work just fine.

markusbattarbee (approved):

markusbattarbee:
> So that's just the same module commands without purge, right? Running right now, I expect it to work just fine.

So was it fine?

lkotipal:
cpeGNU was but cpeCray had an MPI crash, I'll rerun to see if that was a fluke.

PR #799, opened 2023-08-10
Vlasiator gpu

markusbattarbee:
Implemented integrated CUDA/HIP porting mechanism via defines
Integrated @hokkanen Jaro's ARCH-looping method for moments, datareducers, and timestep evaluation
Now also compiles and runs on LUMI-G


kstppd:
yehaaaa

Issue #698, opened 2023-02-06
OpenMPI 4 IO bug


henry2004y:
I happen to have a minimal working example for reproducing the OpenMPI4 VLSV bug.

Here is a simple flowthrough test configuration file:

<details><summary>Flowthrough_1d_sin.cfg</summary>
<p>

```shell
project = Flowthrough
propagate_field = 0
propagate_vlasov_acceleration = 0
propagate_vlasov_translation = 1
dynamic_timestep = 1

ParticlePopulations = proton

[io]
write_initial_state = 0

system_write_t_interval = 20.0
system_write_file_name = bulk
system_write_distribution_stride = 0
system_write_distribution_xline_stride = 1
system_write_distribution_yline_stride = 0
system_write_distribution_zline_stride = 0

[variables]
output = populations_vg_v
output = populations_vg_rho
diagnostic = populations_vg_blocks

[gridbuilder]
x_length = 10
y_length = 1
z_length = 1
x_min = -10.
x_max = 10.
y_min = -1.0
y_max = 1.0
z_min = -1.0
z_max = 1.0
t_max = 20.0
dt = 0.00625

[proton_properties]
mass = 1
mass_units = PROTON
charge = 1

[proton_vspace]
vx_min = +0.5
vx_max = +1.5
vy_min = -1.0
vy_max = +1.0
vz_min = -1.0
vz_max = +1.0
vx_length = 160
vy_length = 10
vz_length = 10

[proton_sparse]
minValue = -1.0e-20

[boundaries]
periodic_x = yes
periodic_y = yes
periodic_z = yes

[Flowthrough]
densityModel = Sinewave
densityWidth = 10.0
Bx = 1e-9
By = 0e-9
Bz = 0e-9

[proton_Flowthrough]
T = 1.0e-7
rho = 0.5
rhoBase = 0.25
VX0 = 1.0
VY0 = 0
VZ0 = 0
nSpaceSamples = 1
nVelocitySamples = 1

[bailout]
velocity_space_wall_block_margin = 0
```

</p>
</details>

When `system_write_distribution_stride = 0`, then the output VLSV files are good. However, when `system_write_distribution_stride = 1`, the output VLSV files lose the metadata at the end.

The current workaround is to disable OpenMPI’s I/O implementation, `export OMPI_MCA_io=^ompio`.

PR #800, opened 2023-08-16
MPI_T MCA disabling mpiio

ykempf:
Here's the prototype, as this has highest precedence, do we want to check whether other ways of setting this parameter are active? Or are we happy with this printf warning/reminder? This also seems to throw an MPI error at the end of execution, at least on my laptop, needs to be investigated. Though it is whenthe job is exiting after successful completion of the simulation, so not a huge problem it seems.

```
--------------------------------------------------------------------------
A system call failed during shared memory initialization that should
not have.  It is likely that your MPI job will now either abort or
experience performance degradation.

  Local host:  lx6-flafo-03
  System call: unlink(2) /dev/shm/vader_segment.lx6-flafo-03.22f0001.1
  Error:       No such file or directory (errno 2)
--------------------------------------------------------------------------
[lx6-flafo-03:20092] *** Process received signal ***
[lx6-flafo-03:20092] Signal: Segmentation fault (11)
[lx6-flafo-03:20092] Signal code: Address not mapped (1)
[lx6-flafo-03:20092] Failing at address: 0x30
--------------------------------------------------------------------------
A system call failed during shared memory initialization that should
not have.  It is likely that your MPI job will now either abort or
experience performance degradation.

  Local host:  lx6-flafo-03
  System call: unlink(2) /dev/shm/vader_segment.lx6-flafo-03.22f0001.0
  Error:       No such file or directory (errno 2)
--------------------------------------------------------------------------
[lx6-flafo-03:20091] *** Process received signal ***
[lx6-flafo-03:20091] Signal: Segmentation fault (11)
[lx6-flafo-03:20091] Signal code: Address not mapped (1)
[lx6-flafo-03:20091] Failing at address: 0x540
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
mpirun noticed that process rank 1 with PID 0 on node lx6-flafo-03 exited on signal 11 (Segmentation fault).
--------------------------------------------------------------------------
```


ykempf:
Related to/workaround for #698 so the users don't have to scratch their head and then set the environment or mpirun MCA flag.

markusbattarbee:
so should we first check what type of MPI is being used? IIRC this was only required for OpenMPI, not MPICH or whatever the third somewhat common implementation was.

Strange that error message, but might be a cubbli-only thing?

ykempf:
Probably a cubbli issue my error message, nothing suspicious on Mahti.

vlasiator.cpp
```diff
@@ -277,6 +309,10 @@ int main(int argn,char* args[]) {
    MPI_Comm comm = MPI_COMM_WORLD;
    MPI_Comm_rank(comm,&myRank);
    
+   if (myRank == MASTER_RANK) {
```
markusbattarbee, line 312: I'd put this inside the `if`-clause on line 272, right at the end after the `#endif`

vlasiator.cpp
```diff
@@ -259,9 +259,41 @@ int main(int argn,char* args[]) {
...
+   stringstream mpiioMessage;
+
+   if(versionstr.find("Open MPI") != std::string::npos) {
+      #ifdef VLASIATOR_ALLOW_MCA_OMPIO
+         mpiioMessage << "We detected OpenMPI but the compilation flag VLASIATOR_ALLOW_MCA_OMPIO was set so we do not override the default MCA io flag." << endl;
+      #else
+         overrideMCAompio = true;
+         int index, count;
+         char io_value[64];
+
```
markusbattarbee, line 279: I'd condense this a bit by removing empty lines

markusbattarbee (approved):

vlasiator.cpp
ykempf: Or can I move that before `MPI_Init_thread()`??

markusbattarbee: Ah right, I was focusing just on where the mpiioMessage is generated. I guess it's fine as is then, not worth adding a check for the length of the mpiioMessage or a comment about mpiioMessages as it's easy enough to see what it's associated with.

ykempf: I just saw that we have two calls to MPI_Comm_rank (well the second one is in the unlikely case `if (required > provided){` – can't remember that ever happening), I simplified that in the upcoming commit.

ykempf: Now it's simpler.

ykempf:
So where is the report of this supposedly successful testpackage? I'd want to see it before hitting this merge button...

ykempf:
That says "Invalid request. Only 65535 characters are allowed; 77087 were supplied."

ykempf:
OK so testpackage report fails with that 64k limit. So here's the output file salvaged from Carrington.
[testpackage_run_output.txt](https://github.com/fmihpc/vlasiator/files/12718605/testpackage_run_output.txt)

ykempf:
And so thereby I declare the testpackage passed and shall merge.

PR #802, opened 2023-08-17
ig_inplanecurrent at restart

ykempf:
Recompute full sigma tensor in ionosphere at restart so that ig_inplanecurrent is sane in output occurring between a restart and the next ionosphere solution step.


ykempf:
The diff looks massive, but it's because of a second commit where I swapped regular `//` for `/*!< ... */`. And the effective change is just to call the tensor construction at restart without recomputing precipitation or height-integration, but inserting an `if` block indents it and the diff is jumbled. Sorry!

ykempf:
Is the reference data up to date?

PR #816, opened 2023-09-06
Fix timer

lkotipal:
Another one


markusbattarbee:
Would you want to migrate all cudasiator branch phiprof timers to objects?

lkotipal:
Sure, I can see how well my dev PR merges and take it from there

PR #820, opened 2023-09-15
FSGrid checkups: Removed duplicate code, use the fsgrid header functions instead

alhom:
Removing duplicate code from vlsvdiff.


alhom:
Well, that was an old `dev` pull. Rebasing.

PR #823, opened 2023-09-18
Debug timers on testpackage runs

lkotipal:
Compile testpackage with `DEBUG_PHIPROF_TIMERS` and check for errors in CI.


markusbattarbee:
But I think this one won't actually make the debug phiprof version available?
https://github.com/fmihpc/vlasiator/blob/8f77053299620866b9dfd0aa03bb280466fe4603/.github/workflows/github-ci.yml#L37

lkotipal:
The actual testpackage run should work on the team's libraries iirc

PR #829, opened 2023-09-29
Submit script for jobs

lkotipal:
Add script `submit.sh` to check Vlasiator config before submitting a batch job.

First, the job script `JOB` must define the Vlasiator binary and config file as `BIN` and `CFG` respectively. Then run `submit.sh JOB` from the work directory. This script performs some checks and then runs the job script if these checks succeed unless flag `-t` or `--test` is used. The checks are as follows:
- Check that both `BIN` and `CFG` exist
- Check that `check_vlasiator_cfg.sh` finds no invalid options
- Check that Maxwellian input file exists if defined
- Check that atmospheric model file exist if defined
- Check that system and restart write paths exist, and output their striping
- Output disk quota for workspace associated with the account the job is run on

### DIFF
--- a/tools/check_vlasiator_cfg.sh
+++ b/tools/check_vlasiator_cfg.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-mpirun_cmd=$(which aprun > /dev/null && echo "aprun" || echo "mpirun")
+mpirun_cmd=$(which aprun &> /dev/null && echo "aprun" || echo "mpirun")
 vlasiator=$1
 cfg=$2
 
--- a/tools/submit.sh
+++ b/tools/submit.sh
@@ -2,7 +2,8 @@
 
 if [[ $# < 1 ]]
 then
-    echo "Usage: ./submit.sh JOB, where JOB is the job script."
+    echo "Usage: ./submit.sh JOB [-t, --test], where JOB is the job script."
+    echo "Use option -t or --test to test without submitting job."
     exit 1
 fi
 
@@ -112,6 +113,9 @@ if [[ $ERR -gt 0 ]]
 then
     echo "Errors found, not submitting job."
     exit 1
+elif [[ $2 == "-t" ]] || [[ $2 == '--test' ]]
+then
+    echo "Test successful"
 else
     echo "Submitting $1"
     sbatch $1
--- a/tools/submit.sh
+++ b/tools/submit.sh
@@ -17,7 +17,7 @@ then
    exit 1
 fi
 
-if ! [[ -f $BIN ]]
+if ! [[ -f $CFG ]]
 then
    echo "Config $CFG not found!"
    exit 1
@@ -34,17 +34,20 @@ fi
 
 SW=$(sed -n s/^file_..\\s*=\\s*//p < $CFG)
 
-if ! test -f $SW
-then
-    echo "Solar wind file $SW not found!"
-    echo ""
-    ERR=1
-fi
+for i in $SW
+do
+    if ! test -f $i
+    then
+        echo "Solar wind file $i not found!"
+        echo ""
+        ERR=1
+    fi
+done
 
 MF=$(sed -n s/^atmosphericModelFile\\s*=\\s*//p < $CFG)
 if [[ -z $MF ]]
 then
-    MF=NRLMSIS.dat
+    MF=$(sed -n "s/ionosphere.atmosphericModelFile (=\(.*\))/\1/p" < check.txt)
 fi
 
 if ! test -f $MF
@@ -60,15 +63,18 @@ then
     IO=.
 fi
 
-if ! test -d $IO
-then
-    echo "System write path $IO not found!"
-    echo ""
-    ERR=1
-else
-    echo $IO
-    lfs getstripe -d $IO
-fi
+for i in $IO
+do
+    if ! test -d $i
+    then
+        echo "System write path $i not found!"
+        echo ""
+        ERR=1
+    else
+        echo $i
+        lfs getstripe -d $i
+    fi
+done
 
 RW=$(sed -n s/^restart_write_path\\s*=\\s*//p < $CFG)
 if [[ -z $RW ]]
--- a/tools/submit.sh
+++ b/tools/submit.sh
@@ -1,0 +1,112 @@
+#!/bin/bash
+
+if [[ $# < 1 ]]
+then
+    echo "Usage: ./submit.sh JOB, where JOB is the job script."
+    exit 1
+fi
+
+ERR=0
+
+BIN=$(sed -n 's/BIN\s*=\s*//p' < $1)
+CFG=$(sed -n 's/CFG\s*=\s*//p' < $1)
+
+if ! [[ -f $BIN ]]
+then
+   echo "Binary $BIN not found!"
+   exit 1
+fi
+
+if ! [[ -f $BIN ]]
+then
+   echo "Config $CFG not found!"
+   exit 1
+fi
+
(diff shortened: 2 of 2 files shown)


tools/submit.sh
```diff
@@ -0,0 +1,112 @@
...
+BIN=$(sed -n 's/BIN\s*=\s*//p' < $1)
+CFG=$(sed -n 's/CFG\s*=\s*//p' < $1)
+
+if ! [[ -f $BIN ]]
+then
+   echo "Binary $BIN not found!"
+   exit 1
+fi
+
+if ! [[ -f $BIN ]]
```
markusbattarbee, line 20: should check for $CFG

tools/submit.sh
```diff
@@ -0,0 +1,112 @@
...
+./check_vlasiator_cfg.sh $BIN $CFG > check.txt
+
+if grep -q "Invalid options" check.txt 
+then
+    echo "Invalid options in config file, check check.txt"
+    echo ""
+    ERR=1
+fi
+
+SW=$(sed -n s/^file_..\\s*=\\s*//p < $CFG)
```
markusbattarbee, line 35: So does this check for several sw files? e.g. if we use multipop or have numerous different inflow boundaries.

tools/submit.sh
```diff
@@ -0,0 +1,112 @@
...
+then
+    echo "Solar wind file $SW not found!"
+    echo ""
+    ERR=1
+fi
+
+MF=$(sed -n s/^atmosphericModelFile\\s*=\\s*//p < $CFG)
+if [[ -z $MF ]]
+then
+    MF=NRLMSIS.dat
```
markusbattarbee, line 47: What if there's no ionosphere present in the run? In that case we don't want to check for this file.

tools/submit.sh
```diff
@@ -0,0 +1,112 @@
...
+fi
+
+if ! test -d $IO
+then
+    echo "System write path $IO not found!"
+    echo ""
+    ERR=1
+else
+    echo $IO
+    lfs getstripe -d $IO
```
markusbattarbee, line 70: Excellent to check for striping :)

tools/submit.sh
```diff
@@ -0,0 +1,112 @@
...
+fi
+
+if ! test -f $MF
+then
+    echo "Atmospheric model file $MF not found!"
+    echo ""
+    ERR=1
+fi
+
+IO=$(sed -n s/^system_write_path\\s*=\\s*//p < $CFG)
```
markusbattarbee, line 57: Same, does this check for several IO directories?

markusbattarbee (changes requested):
promising

tools/submit.sh
lkotipal: Fixed to iterate over every sw file

tools/submit.sh
lkotipal: Now defaults to what `check_vlasiator_cfg.sh` gives, which is empty (always passes `test`) for non-ionosphere runs

markusbattarbee: Sorry to nitpick, but the ionospheric boundary does default to using `NRLMSIS.dat`. So What this should do is a) check if an ionospheric boundary is contained in the config, and then either verify that a configured atmospheric model file exists, or if one isn't configured, that the default file NRLMSIS.dat exists.

lkotipal: `check_vlasiator_cfg.sh` gives under unused options `ionosphere.atmosphericModelFile (=NRLMSIS.dat)` on ionosphere runs, which is parsed by the script. In non-ionosphere runs the unused option is absent and the script `sed`s an empty string for `$MF`.

tools/submit.sh
lkotipal: Fixed to iterate over every IO path

tools/submit.sh
markusbattarbee: Ah, ok, good then!

markusbattarbee (approved):

tools/check_vlasiator_cfg.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-mpirun_cmd=$(which aprun > /dev/null && echo "aprun" || echo "mpirun")
+mpirun_cmd=$(which aprun &> /dev/null && echo "aprun" || echo "mpirun")
```
ykempf, line 3: Ah good addition of `&`! Now many times I drop `mpirun` even as e.g. on frontends or my laptop the direct serial running is sufficient to get the output. So should we change this line altogether, if we're fiddling with it anyway?

tools/submit.sh
```diff
@@ -0,0 +1,122 @@
...
+then
+   echo "Binary $BIN not found!"
+   exit 1
+fi
+
+if ! [[ -f $CFG ]]
+then
+   echo "Config $CFG not found!"
+   exit 1
+fi
```
ykempf, line 25: How about passing those as arguments instead?

tools/submit.sh
```diff
@@ -0,0 +1,122 @@
...
+then
+    MF=$(sed -n "s/ionosphere.atmosphericModelFile (=\(.*\))/\1/p" < check.txt)
+fi
+
+if ! test -f $MF
+then
+    echo "Atmospheric model file $MF not found!"
+    echo ""
+    ERR=1
+fi
```
ykempf, line 59: OK I see there's more of these environment-variable-based guys. So I suggest either of the below
- pass all input on the command line
- have the error messages print both the environment variable name and its contents – as I read this atm it'll print a blank if a variable is not set, it would be more helpful if it would tell the user what to set to define it. Me dumb user won't remember all these variables and I'll go sequentially setting them until it stops complaining.

If you pass them all on the command line as options, then you can check the count at the beginning and print what the options should be, in order.

If you want them as environment variables, how to skip one if it's not needed, e.g. in a 2D run no atmospheric model file is needed?

tools/submit.sh
```diff
@@ -0,0 +1,122 @@
...
+
+IO=$(sed -n s/^system_write_path\\s*=\\s*//p < $CFG)
+if [[ -z $IO ]]
+then
+    IO=.
+fi
+
+for i in $IO
```
ykempf, line 67: How do you pass several here? Will `IO="bulk bulk1 bulk5"` be enough?

ykempf: Why not parse from the cfg like you do for RW below?

tools/submit.sh
ykempf: doops

ykempf (commented):
Hupsis I must have been a bit cross-eyed yesterday indeed.
Now some/many of these options can also be legally passed to vlasiator on the run line in the jobscript, but let's not maybe ask you to foolproofly parse those out of the job script and be happy with this for now. (Up to the one open comment I still have.)

markusbattarbee:
The two things we do usually pass from the command line are the configuration file (which is being checked for) and a restart file. However, the restart file is sometimes taken from parsing the list of restarts in a given directory and taking the latest one, so I don't really see how the existence of such a file could be verified in a check script like this. So not actually requesting anything, just discussing.

What this now does require is that the job script to be checked *must* use `BIN` and `CFG` to declare the binary and the config file, but I guess we can live with that level of standardization.

lkotipal:
Using latest restart file could be a config parameter as well I suppose, instead of always copypasting the oneliner from previous job scripts.

tools/check_vlasiator_cfg.sh
lkotipal: Probably yes. I'm pretty sure `check_vlasiator_cfg` is broken on mpirun as well, I've just run it serially since it's just reading the parameters.

PR #830, opened 2023-10-12
Introduces a parameter to limit the allowed maximum spatial refinement level

ykempf:
Up for discussion and CI.

### DIFF
--- a/parameters.cpp
+++ b/parameters.cpp
@@ -433,7 +433,7 @@ bool P::addParameters() {
            (Realf)0.5);
    // Spatial Refinement parameters
    RP::add("AMR.max_spatial_level", "Maximum absolute spatial mesh refinement level", (uint)0);
-   RP::add("AMR.max_allowed_spatial_level", "Maximum currently allowed spatial mesh refinement level", (uint)0);
+   RP::add("AMR.max_allowed_spatial_level", "Maximum currently allowed spatial mesh refinement level", (uint)42);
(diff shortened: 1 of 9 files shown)


ykempf:
Eh. Let me change that already.

ykempf:
So the terminology of the new flag is: we have the already-existing max refinement level and I don't want to rename this, and we have the new one that tells what the actually allowed maximum is, and that should be smaller than the other pre-existing flag. But I see how this _allowed_ terminology might be confusing, so suggestions welcome to make this better. And the next push will allow for not setting the new flag in configs so the users don't have surprises...

parameters.cpp
```diff
@@ -157,6 +157,7 @@ Realf P::vamrCoarsenLimit = 0.5;
 string P::vamrVelRefCriterion = string("");
 
 uint P::amrMaxSpatialRefLevel = 0;
+uint P::amrMaxAllowedSpatialRefLevel = 0;
```
lkotipal, line 160: Rather have this as a signed integer, with e.g. negative values for automatic behavior

ykempf: Ah, *this* is why it wasn't merged yet. :)

ykempf: Better?

lkotipal: Perfect

ykempf:
Yay! A pass from CI! Merge?

PR #831, opened 2023-10-12
Fix subcycles resetting on refinement

lkotipal:


sysboundary/sysboundary.cpp
```diff
@@ -490,7 +490,10 @@ bool SysBoundary::classifyCells(dccrg::Dccrg<spatial_cell::SpatialCell, dccrg::C
             // Here for debugging since boundarytype should be fed from MPIGrid
             technicalGrid.get(x, y, z)->sysBoundaryFlag = sysboundarytype::N_SYSBOUNDARY_CONDITIONS;
             technicalGrid.get(x, y, z)->sysBoundaryLayer = 0;
-            technicalGrid.get(x, y, z)->maxFsDt = numeric_limits<Real>::max();
+            // Don't screw with maximum dt when refining
+            if (P::tstep == 0) {
```
ykempf, line 494: Does this need additional handling when restarting?

lkotipal: `technicalGrid` is read on restart after boundary cell classification

ykempf: https://github.com/fmihpc/vlasiator/blob/dev/iowrite.cpp#L1659
No we only write and re-read fg_perturbed_b and fg_e.

lkotipal: Changed to use `tstep_min`

vlasiator.cpp
```diff
@@ -985,16 +985,15 @@ int main(int argn,char* args[]) {
...
             logFile << "(AMR): Adapting refinement!"  << endl << writeVerbose;
             if (!adaptRefinement(mpiGrid, technicalGrid, sysBoundaryContainer, *project))
                continue;   // Refinement failed and we're bailing out
 
             // Calculate new dt limits since we might break CFL when refining
-            phiprof::Timer computeDtimer {"compute-dt"};
+            phiprof::Timer computeDtimer {"compute-dt-amr"};
             calculateSpatialTranslation(mpiGrid,0.0);
-            calculateAcceleration(mpiGrid,0.0);      
+            calculateAcceleration(mpiGrid,0.0); // Might be unnecessary
```
ykempf, line 996: Though as Markus said, the bvol and derivatives for the Hall term in acceleration will changed, right?

lkotipal: Yeah. Thought I'd leave the comment for later testing but that might be more misleading after all

projects/Magnetosphere/Magnetosphere.cpp
```diff
@@ -702,8 +702,16 @@ namespace projects {
...
       int myRank;       
       MPI_Comm_rank(MPI_COMM_WORLD,&myRank);
-      if (myRank == MASTER_RANK)
+      if (myRank == MASTER_RANK) {
          std::cout << "Maximum refinement level is " << mpiGrid.mapping.get_maximum_refinement_level() << std::endl;
+      }
+      
+      if (!P::useAlpha && !P::useJPerB) {
+         if (myRank == MASTER_RANK) {
+            std::cout << "WARNING All refinement indices disabled" << std::endl;
```
ykempf, line 711: Tell the user of the future this means no refinement will take place, or only static if activated? Or will this function not be called for static, so just tell that no refinement is expected to occur?

lkotipal: This code path is only if you have adaptive refinement enabled with both refinement indices disabled. Thinking more on it, I think that should be an early return as well

ykempf: Well it is a return just below this.

lkotipal: So it is, just with a wrong wrong return type (now fixed)

fieldsolver/derivatives.cpp
```diff
@@ -700,9 +700,10 @@ static std::array<Real, 3> getMomentumDensity(SpatialCell* cell)
  */
 static Real calculateU(SpatialCell* cell)
 {
+   Real rho = cell->parameters[CellParams::RHOM];
    std::array<Real, 3> p = getMomentumDensity(cell);
    std::array<Real, 3> B = getBVol(cell);
-   return (pow(p[0], 2) + pow(p[1], 2) + pow(p[2], 2)) / (2.0 * cell->parameters[CellParams::RHOM]) + (pow(B[0], 2) + pow(B[1], 2) + pow(B[2], 2)) / (2.0 * physicalconstants::MU_0);
+   return rho > EPS ? (pow(p[0], 2) + pow(p[1], 2) + pow(p[2], 2)) / (2.0 * cell->parameters[CellParams::RHOM]) : 0.0 + (pow(B[0], 2) + pow(B[1], 2) + pow(B[2], 2)) / (2.0 * physicalconstants::MU_0);
```
ykempf, line 706: I appreciate your skill but this is approaching a level of one-liner complexity I'm not sure we want to aim for.

lkotipal: Yeah, I'll add line breaks to make it easier to swallow at least

ykempf: I *assume* and **hope** that you know the precedence rules. I don't know those in particular. :)

lkotipal: Nice catch, I usually err on the side of parens but forgot them here

ykempf: Ahaa. Oho!

fieldsolver/derivatives.cpp
```diff
@@ -766,19 +777,21 @@ void calculateScaledDeltas(
    std::array<Real, 3> myJ = {dBZdy - dBYdz, dBXdz - dBZdx, dBYdx - dBXdy};
    Real BdotJ {0.0};
    Real Bsq {0.0};
+   Real J {0.0};
    for (int i = 0; i < 3; ++i) {
       BdotJ += myB[i] * myJ[i];
       Bsq += myB[i] * myB[i];
+      J += myJ[i] * myJ[i];
    }
+   J = std::sqrt(J);
```
ykempf, line 786: Is this *really* certain never to be 0 by "chance", "bad luck" or "design", e.g. at t=0?

lkotipal: No, but sqrt(0) would just be 0 no?

alhom: Might it be `sqrt(-0)`?

lkotipal: Shouldn't be negative. I'm going to be upset if IEEE floats allow a sum of square numbers to be...

fieldsolver/derivatives.cpp
```diff
@@ -766,19 +777,21 @@ void calculateScaledDeltas(
...
 
    Real Bperp {0.0};
-   Real J {0.0};
-   for (int i = 0; i < 3; ++i) {
-      Bperp += std::pow(myB[i] * (1 - BdotJ / Bsq), 2);
-      J += myJ[i] * myJ[i];
+   if (Bsq > EPS) {
+      for (int i = 0; i < 3; ++i) {
```
markusbattarbee, line 790: so here we should use perB + BGB to evaluete Bperp

parameters.cpp
```diff
@@ -680,13 +685,30 @@ void Parameters::getParameters() {
...
+      cerr << "Alpha coarsening threshold not set, using half of refine threshold" << endl;
+      P::alphaCoarsenThreshold = P::alphaRefineThreshold / 2.0;
+   }
+   if (P::useAlpha && P::alphaRefineThreshold < 0) {
+      cerr << "WARNING using alpha without refine threshold set" << endl;
+   }
+   RP::get("AMR.use_J_per_B",P::useJPerB);
+   RP::get("AMR.jperb_refine_threshold",P::jperbRefineThreshold);
+   RP::get("AMR.jperb_coarsen_threshold",P::jperbCoarsenThreshold);
+   if (P::jperbCoarsenThreshold < 0) {
```
markusbattarbee, line 701: same comments for jperb than for alpha

markusbattarbee (changes requested):
There's also the question of whether the defaults for refine and coarsen thresholds should be -1, or perhaps some tested default good-enough values?

parameters.cpp
```diff
@@ -680,13 +685,30 @@ void Parameters::getParameters() {
    RP::get("AMR.refine_on_restart",P::refineOnRestart);
    RP::get("AMR.force_refinement",P::forceRefinement);
    RP::get("AMR.should_filter",P::shouldFilter);
-   RP::get("AMR.refine_threshold",P::refineThreshold);
-   RP::get("AMR.unrefine_threshold",P::unrefineThreshold);
+   RP::get("AMR.use_alpha",P::useAlpha);  // Now should this just be P::refineThreshold > 0?
+   RP::get("AMR.alpha_refine_threshold",P::alphaRefineThreshold);
+   RP::get("AMR.alpha_coarsen_threshold",P::alphaCoarsenThreshold);
+   if (P::alphaCoarsenThreshold < 0) {
```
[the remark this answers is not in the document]
lkotipal: Oh yeah, that would give the nag and change the coarsening threshold without actually doing anything useful

parameters.cpp
```diff
@@ -437,13 +440,15 @@ bool P::addParameters() {
    RP::add("AMR.refine_on_restart","If true, re-refine vlasov grid on restart", false);
    RP::add("AMR.force_refinement","If true, refine/unrefine the vlasov grid to match the config on restart", false);
    RP::add("AMR.should_filter","If true, filter vlasov grid with boxcar filter on restart",false);
-   RP::add("AMR.refine_threshold","Determines the minimum value of the refinement parameter to refine cells", 1.0);
-   RP::add("AMR.unrefine_threshold","Determines the maximum value of the refinement parameter to unrefine cells", 0.0);
+   RP::add("AMR.use_alpha","Use alpha as a refinement index", false);
```
ykempf, line 443: Write out to some extent what "alpha" means, this is for the end user. So at least something about differences to neighbors.

parameters.cpp
```diff
@@ -437,13 +440,15 @@ bool P::addParameters() {
...
+   RP::add("AMR.alpha_refine_threshold","Determines the minimum value of alpha to refine cells", -1.0);
+   RP::add("AMR.alpha_coarsen_threshold","Determines the maximum value of alpha to unrefine cells", -1.0);
+   RP::add("AMR.use_J_per_B","Use J/B_perp as a refinement index", false);
+   RP::add("AMR.jperb_refine_threshold","Determines the minimum value of jperb to refine cells", -1.0);
+   RP::add("AMR.jperb_coarsen_threshold","Determines the maximum value of jperb to unrefine cells", -1.0);
```
ykempf, line 448: Should the defaults be sensible values instead?

lkotipal: For coarsens, that defaults to half the refine threshold. I'm personally leaning more towards removing `use_whatever` and replacing that with having the refine threshold set, but either works I suppose

ykempf: Easier to switch on and off with a separate line in a cfg, but either way works.

lkotipal: You could just switch it off by commenting/deleting the refine threshold since that defaults to -1

projects/Magnetosphere/Magnetosphere.cpp
```diff
@@ -747,7 +747,7 @@ namespace projects {
                   ++refined_neighbors;
                } else if (neighborRef < refLevel) {
                   ++coarser_neighbors;
-               } else if ((P::useAlpha ? mpiGrid[neighbor]->parameters[CellParams::AMR_ALPHA] > P::alphaRefineThreshold : false) || (P::useJPerB ? neighborAlphaTwo > P::jperbRefineThreshold : true)) {
+               } else if ((P::useAlpha ? mpiGrid[neighbor]->parameters[CellParams::AMR_ALPHA] > P::alphaRefineThreshold : false) || (P::useJPerB ? neighborAlphaTwo > P::jperbRefineThreshold : false)) {
                   refined_neighbors += 4;
```
markusbattarbee, line 751: Could you add a comment about why this increments with 4 please?

ykempf:
Now that we're test-firing and fire-testing this...
We ran into the built-in memory bailout limit vs. planned memory needs exit on 256 nodes for FID on LUMI. I was surprised, but I now think it's expectable. Could we instead do it in 2 steps, if the planned memory footprint exceeds the bailout limit, do a load balance based on that target memory footprint/block count, and estimate again whether it would go bust, if yes, exit with a restart, if no proceed with refinement? Or do we run the risk of running OOM during that intermediate LB step in cases where a large region would be unrefining, thus with a light target footprint but still big before unrefining?

vlasiator.cpp
```diff
@@ -992,7 +992,14 @@ int main(int argn,char* args[]) {
             logFile << "(AMR): Adapting refinement!"  << endl << writeVerbose;
             if (!adaptRefinement(mpiGrid, technicalGrid, sysBoundaryContainer, *project)) {
                // OOM, rebalance and try again
+               for (auto id : mpiGrid.get_local_cells_to_refine()) {
+                  mpiGrid[id]->parameters[CellParams::LBWEIGHTCOUNTER] *= 8.0;
+               }
                balanceLoad(mpiGrid, sysBoundaryContainer);
+               for (auto id : mpiGrid.get_local_cells_to_refine()) {
```
markusbattarbee, line 999: Now this has me wondering. Should we in fact leave the LBWEIGHTCOUNTER value intact? Because then it would write those values into the restart file, and they would be used by the new more-tasks run for initial LB!

grid.cpp
```diff
@@ -1403,20 +1404,22 @@ void mapRefinement(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
...
       refines = project.adaptRefinement(mpiGrid);
    }
 
    int cells = getLocalCells().size();
    MPI_Allreduce(MPI_IN_PLACE, &refines, 1, MPI_INT, MPI_SUM, MPI_COMM_WORLD);
    MPI_Allreduce(MPI_IN_PLACE, &cells, 1, MPI_INT, MPI_SUM, MPI_COMM_WORLD);
-   double ratio = static_cast<double>(refines) / static_cast<double>(cells);
-   logFile << "(AMR) Refining " << refines << " cells, " << 100.0 * ratio << "% of grid" << std::endl;
+   double ratio_refines = static_cast<double>(refines) / static_cast<double>(cells);
+   logFile << "(AMR) Refining " << refines << " cells, " << 100.0 * ratio_refines << "% of grid" << std::endl;
```
ykempf, line 1422: What will be more useful, comparing to exisitng cells or comparing to the max theoretical/fsgrid cell count? Or add the latter to the line as well, so make it print ratio to existing cells, existing cell count and fsgrid cell count?

ykempf: Here or as part of the below print, no need to have that info on both I suppose.

parameters.cpp
ykempf: Still asking for more explanation here. :)

vlasiator.cpp
ykempf: And so this is handled now, correct?

fieldsolver/derivatives.cpp
```diff
@@ -825,5 +838,7 @@ void calculateScaledDeltasSimple(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geome
       computeTimer.stop(N_cells,"Spatial Cells");
    }
 
+
```
ykempf, line 841: Meh, stoopid githoob. I want to comment on 20 lines above, we can pare down from ALL_SPATIAL_DATA to less stuff being communicated.

vlasiator.cpp
lkotipal: Yes, in a different way

parameters.cpp
```diff
@@ -437,13 +440,15 @@ bool P::addParameters() {
    RP::add("AMR.refine_on_restart","If true, re-refine vlasov grid on restart", false);
    RP::add("AMR.force_refinement","If true, refine/unrefine the vlasov grid to match the config on restart", false);
    RP::add("AMR.should_filter","If true, filter vlasov grid with boxcar filter on restart",false);
-   RP::add("AMR.refine_threshold","Determines the minimum value of the refinement parameter to refine cells", 1.0);
-   RP::add("AMR.unrefine_threshold","Determines the maximum value of the refinement parameter to unrefine cells", 0.0);
-   RP::add("AMR.refine_multiplier","Refine every nth load balance", 10);
+   RP::add("AMR.use_alpha","Use alpha as a refinement index", false);
+   RP::add("AMR.alpha_refine_threshold","Determines the minimum value of alpha to refine cells", -1.0);
+   RP::add("AMR.alpha_coarsen_threshold","Determines the maximum value of alpha to unrefine cells", -1.0);
+   RP::add("AMR.use_J_per_B","Use J/B_perp as a refinement index", false);
```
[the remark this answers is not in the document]
lkotipal: Yes, I think it should be renamed along with alpha i.e. alpha1

fieldsolver/derivatives.cpp
lkotipal: This is pretty old. Are we good with `CELL_PARAMETERS` or?

ykempf: No, you require derivatives for jperb so those are needed too. And in CELL_PARAMETERS you don't need a whole lot but I don't know whether there is flags for just the things you need, rho, v and bvol probably, but pressure I'm not sure.

markusbattarbee: ALL_SPATIAL_DATA is:
```
  const uint64_t ALL_SPATIAL_DATA =
      CELL_PARAMETERS
      | CELL_DERIVATIVES | CELL_BVOL_DERIVATIVES
      | CELL_SYSBOUNDARYFLAG
      | POP_METADATA | RANDOMGEN;
      ```
      So I'd stick with this. (It's a bit confusingly named, but it indeed doesn't include distribution data.

File: parameters.cpp (hunk 9 above)
Comment (lkotipal) [reply parameters.cpp]: Better?

Comment (ykempf) [reply parameters.cpp]: Yes. I am baffled that I don't see the discussion on the diff view even though it's not resolved yet. But well...

File: projects/project.cpp [hunk 16]
```diff
@@ -600,14 +600,81 @@ namespace projects {
...
+         std::array<double,3> xyz {mpiGrid.get_center(id)};
+         SpatialCell* cell {mpiGrid[id]};
+         int refLevel {mpiGrid.get_refinement_level(id)};
+         Real r2 {pow(xyz[0], 2) + pow(xyz[1], 2) + pow(xyz[2], 2)};
+
+         if (!canRefine(mpiGrid[id])) {
+            // Skip refining, touching boundaries during runtime breaks everything
+            mpiGrid.dont_refine(id);
+            mpiGrid.dont_unrefine(id);
+         } else if (r2 < r_max2) {
```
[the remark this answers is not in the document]
lkotipal: The default right now is nonsensically large, it's `LARGE_REAL`

parameters.cpp
markusbattarbee: But if we put some relatively sane thresholds as default values (which I think we should do), we need a flag for activating / deactivating the parameter.

projects/project.cpp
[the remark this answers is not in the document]
lkotipal: That goes to inf on Godbolt and I _think_ that's how float overflow is guaranteed to work. Sounds nasty but works as expected here (any distance will compare smaller) and it won't propagate elsewhere.

projects/project.cpp
```diff
@@ -600,14 +600,81 @@ namespace projects {
...
+   }
+
    int Project::adaptRefinement( dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid ) const {
-      int myRank;
+      phiprof::Timer refinesTimer {"Set refines"};
+      int myRank;       
       MPI_Comm_rank(MPI_COMM_WORLD,&myRank);
       if (myRank == MASTER_RANK) {
-         cerr << "(Project.cpp) Base class 'adaptRefinement' in " << __FILE__ << ":" << __LINE__ << " called. Function is not implemented for project." << endl;
+         std::cout << "Maximum refinement level is " << mpiGrid.mapping.get_maximum_refinement_level() << std::endl;
```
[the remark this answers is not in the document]
ykempf: I would advocate to write – if anything – to logfile.txt because it's hard to tie events in slurm files with logfile post-hoc. Given we have the option of capping AMR since today's merge to dev, I would suggest indeed having a mention somewhere of max allowed and max theoretical refinement levels (since we're often quite good at modifying cfgs during the run without proper archiving of the evolution of the cfg along the run. But in logfile, not in stdout.

ykempf:
Among the many ideas: I think you could also add a feature in the SAVE STOP KILL DOLB series that triggers a refinement.

lkotipal:
> Among the many ideas: I think you could also add a feature in the SAVE STOP KILL DOLB series that triggers a refinement.

Implemented in latest commits. Right now restart refinement can be considered deprecated, it seems to be somewhat unstable and in the interests of maintainability I think it's better that every grid refinement is done in a single point in the code in the main loop.

parameters.cpp
lkotipal: Defaulted to 0.5 for now

grid.cpp
```diff
@@ -1403,20 +1404,25 @@ void mapRefinement(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
 
 bool adaptRefinement(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid, FsGrid<fsgrids::technical, FS_STENCIL_WIDTH> & technicalGrid, SysBoundary& sysBoundaries, Project& project, int useStatic) {
    phiprof::Timer amrTimer {"Re-refine spatial cells"};
-   calculateScaledDeltasSimple(mpiGrid);
-
    int refines {0};
    if (useStatic > -1) {
       project.forceRefinement(mpiGrid, useStatic);
    } else {
+      // Restarts don't have all the data we need, so refinement indices need to be read
```
markusbattarbee, line 1411: So I find this contradictory: The comment states that restarts do not have all the necessary data, but the if-check on line 1412 is not performed for restarts.

lkotipal: Ah, that's just a confusing comment. Restarts don't have some derivatives calculated, i.e. alphas are read from the restart file and not calculated here

grid.cpp
lkotipal: Yeah I think that's going to be the most relevant as well.

vlasiator.cpp
```diff
@@ -924,25 +929,36 @@ int main(int argn,char* args[]) {
...
+      MPI_Bcast( &doNow, 3 , MPI_INT , MASTER_RANK ,MPI_COMM_WORLD);
       writeRestartNow = doNow[0];
       doNow[0] = 0;
       if (doNow[1] == 1) {
          P::prepareForRebalance = true;
          doNow[1] = 0;
       }
+      if (doNow[2]) {
+         refineNow = true;
+         doNow[2] = false;
```
markusbattarbee, line 950: I think we should also set overrideRebalanceNow=true here so that the refinement happens actually right now instead of at the next LB.

lkotipal: doNow[2] also sets doNow[1] so it prepares for rebalance and goes through that first. iirc that requires one timestep for calculating weights.

parameters.cpp
```diff
@@ -436,16 +444,23 @@ bool P::addParameters() {
...
-   RP::add("AMR.refine_threshold","Determines the minimum value of the refinement parameter to refine cells", 1.0);
-   RP::add("AMR.unrefine_threshold","Determines the maximum value of the refinement parameter to unrefine cells", 0.0);
-   RP::add("AMR.refine_multiplier","Refine every nth load balance", 10);
+   RP::add("AMR.use_alpha1","Use the maximum of dimensionless gradients alpha_1 as a refinement index", false);
+   RP::add("AMR.alpha1_refine_threshold","Determines the minimum value of alpha_1 to refine cells", 0.5);
+   RP::add("AMR.alpha1_coarsen_threshold","Determines the maximum value of alpha_1 to unrefine cells", -1.0);
+   RP::add("AMR.use_alpha2","Use J/B_perp as a refinement index", false);
+   RP::add("AMR.alpha2_refine_threshold","Determines the minimum value of alpha_2 to refine cells", 0.5);
+   RP::add("AMR.alpha2_coarsen_threshold","Determines the maximum value of alpha_2 to unrefine cells", -1.0);
+   RP::add("AMR.refine_multiplier","Refine every nth load balance", 1); // Consider renaming
```
markusbattarbee, line 456: `AMR.refine_cadence` would clarify that it's about how often it takes place. Suggest making the default 5 or so.

projects/project.cpp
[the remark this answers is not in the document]
markusbattarbee: So we already have printouts about adapting refinement going to the logfile, let's delete this code block.

sysboundary/sysboundary.cpp
```diff
@@ -490,7 +490,10 @@ bool SysBoundary::classifyCells(dccrg::Dccrg<spatial_cell::SpatialCell, dccrg::C
             // Here for debugging since boundarytype should be fed from MPIGrid
             technicalGrid.get(x, y, z)->sysBoundaryFlag = sysboundarytype::N_SYSBOUNDARY_CONDITIONS;
             technicalGrid.get(x, y, z)->sysBoundaryLayer = 0;
-            technicalGrid.get(x, y, z)->maxFsDt = numeric_limits<Real>::max();
+            // Don't screw with maximum dt when refining
```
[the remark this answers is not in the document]
markusbattarbee: still, please correct.

grid.cpp
```diff
@@ -241,11 +241,11 @@ void initializeGrids(
             balanceLoad(mpiGrid, sysBoundaries);
          }
       } else if (P::refineOnRestart) {
+         // Get good load balancing for refinement
```
markusbattarbee, line 244: Add a "deprecated" comment here as well. I would also like to see a comment above on  line 235 about what Force refinement does, and what that i parameter to the call actually is.

ioread.cpp
```diff
@@ -1251,6 +1262,10 @@ bool exec_readGrid(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
    if(success) { success=readCellParamsVariable(file,fileCells,localCellStartOffset,localCells,"max_fields_dt",CellParams::MAXFDT,1,mpiGrid); }
    if(success) { success=readCellParamsVariable(file,fileCells,localCellStartOffset,localCells,"vg_drift",CellParams::BULKV_FORCING_X,3,mpiGrid); }
    if(success) { success=readCellParamsVariable(file,fileCells,localCellStartOffset,localCells,"vg_bulk_forcing_flag",CellParams::FORCING_CELL_NUM,1,mpiGrid); }
+   if (P::refineOnRestart) {
+      if(success) { success=readCellParamsVariable(file,fileCells,localCellStartOffset,localCells,"vg_amr_alpha",CellParams::AMR_ALPHA,1,mpiGrid); }
```
markusbattarbee, line 1266: I was thinking if we should rename these (and in iowrite.cpp) to amr_alpha1 and amr_alpha2, but that would break the current ongoing runs, so just add comments here about these being alpha1 and alpha2

parameters.cpp
```diff
@@ -162,13 +162,21 @@ bool P::adaptRefinement = false;
...
+bool P::useJPerB = false;
+Real P::jperbRefineThreshold = 0.5;
+Real P::jperbCoarsenThreshold = -1.0;
+Real P::alphaDRhoWeight = 1.0;
+Real P::alphaDUWeight = 1.0;
+Real P::alphaDPSqWeight = 1.0;
+Real P::alphaDBSqWeight = 1.0;
+Real P::alphaDBWeight = 1.0;
+
+uint P::refineMultiplier = 1;
```
markusbattarbee, line 177: How about renaming this to `P::refineCadence`?

parameters.cpp
```diff
@@ -692,13 +707,45 @@ void Parameters::getParameters() {
...
+      P::jperbCoarsenThreshold = P::jperbRefineThreshold / 2.0;
+   }
+   if (P::useJPerB && P::jperbRefineThreshold < 0) {
+      if (myRank == MASTER_RANK) {
+         cerr << "ERROR invalid alpha_2 refine threshold" << endl;
+      }
+      MPI_Abort(MPI_COMM_WORLD, 1);
+   }
+
    RP::get("AMR.refine_multiplier",P::refineMultiplier);
```
markusbattarbee, line 741: here as well renaming to cadence

parameters.h
```diff
@@ -191,13 +191,20 @@ struct Parameters {
...
    static bool shouldFilter;
-   static Real refineThreshold;
-   static Real unrefineThreshold;
+   static bool useAlpha;
+   static Real alphaRefineThreshold;
+   static Real alphaCoarsenThreshold;
+   static bool useJPerB;
+   static Real jperbRefineThreshold;
+   static Real jperbCoarsenThreshold;
    static uint refineMultiplier;
```
markusbattarbee, line 200: cadence

parameters.cpp
```diff
@@ -436,16 +444,23 @@ bool P::addParameters() {
...
    RP::add("AMR.should_refine","If false, do not refine Vlasov grid regardless of max spatial level",true);
    RP::add("AMR.adapt_refinement","If true, re-refine vlasov grid every refine_multiplier load balance", false);
-   RP::add("AMR.refine_on_restart","If true, re-refine vlasov grid on restart", false);
+   RP::add("AMR.refine_on_restart","If true, re-refine vlasov grid on restart. DEPRECATED, consider using the DOMR command", false);
    RP::add("AMR.force_refinement","If true, refine/unrefine the vlasov grid to match the config on restart", false);
    RP::add("AMR.should_filter","If true, filter vlasov grid with boxcar filter on restart",false);
-   RP::add("AMR.refine_threshold","Determines the minimum value of the refinement parameter to refine cells", 1.0);
-   RP::add("AMR.unrefine_threshold","Determines the maximum value of the refinement parameter to unrefine cells", 0.0);
-   RP::add("AMR.refine_multiplier","Refine every nth load balance", 10);
+   RP::add("AMR.use_alpha1","Use the maximum of dimensionless gradients alpha_1 as a refinement index", false);
```
markusbattarbee, line 450: since adaptive refinement is anyway behind `AMR.adapt_refinement`, I suggest making `use_alpha1` and `use_alpha2` default true.

markusbattarbee (changes requested):
Few more change suggestions and questions

sysboundary/sysboundary.cpp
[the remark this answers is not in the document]
lkotipal: Wait no, it's exactly as the comment says. The check fixes an issue where FS subcycles would reset on adaptive refinement.

vlasiator.cpp
```diff
@@ -985,16 +1001,36 @@ int main(int argn,char* args[]) {
       //TODO - add LB measure and do LB if it exceeds threshold
       if(((P::tstep % P::rebalanceInterval == 0 && P::tstep > P::tstep_min) || overrideRebalanceNow)) {
          logFile << "(LB): Start load balance, tstep = " << P::tstep << " t = " << P::t << endl << writeVerbose;
-         // Refinement includes LB
-         if (!dtIsChanged && P::adaptRefinement && P::tstep % (P::rebalanceInterval * P::refineMultiplier) == 0 && P::t > P::refineAfter) { 
+         if (refineNow || (!dtIsChanged && P::adaptRefinement && P::tstep % (P::rebalanceInterval * P::refineMultiplier) == 0 && P::t > P::refineAfter)) {
```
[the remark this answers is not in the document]
lkotipal: Managed to dig up the commit introducing that from the blame. Changing dt doesn't advance the timestep, so if this check wasn't here it would refine twice when timestep changes. Since dt change is checked after LB/refinement, refinement should occur before timestep changing just fine.

However I'm not entirely certain the check is necessary, since I'm pretty sure load balances aren't doubled either in that case (and we're inside the `if` for that).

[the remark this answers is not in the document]
markusbattarbee: Thank you! Now with this, digging into the main loop I again learned new things and ended up looking at git blames from 11 years ago. All good. I guess we could get rid of some head-scratching if there was a comment to that effect here, but that doesn't need to be your job.

markusbattarbee (approved):
I approve now!

PR #832, opened 2023-10-15
Add rescaleDensity to Flowthrough to stabilize testing.

alhom:
Suggest adding rescaleDensity to Flowthrough, to actually have the densities requested.
Notes:
- Still has the extra 4x factors to have it consistent with the old project. IMO these should go so sP.rho etc do what you would expect them to.
- Code duplication... How about a proper decomposition of `f(r,v)` to `f(r)f(v)`? Could be useful for parametrization, but not a priority...


markusbattarbee (approved):
I agree that the 4x multiplier is odd and should be removed, but that will cause testpackage diffs when it takes place. I recommend it be included in other, larger changes to testpackage, e.g. the GPU merge. 

The decomposition into spatial and v-space parametrization is also needed for efficient GPU initialization, and is WIP in one of my branches.

PR #833, opened 2023-10-16
Limit unrefinement

lkotipal:
Unrefine only if both refinement indices agree to unrefine. This gives a cleaner result and avoids the two indices "fighting" by unrefining and refining in turn, comparison in comments:

### DIFF
--- a/projects/Magnetosphere/Magnetosphere.cpp
+++ b/projects/Magnetosphere/Magnetosphere.cpp
@@ -724,23 +724,23 @@ namespace projects {
             mpiGrid.dont_unrefine(id);
          } else if (r2 < r_max2) {
             // We don't care about cells that are too far from the ionosphere
-            const Real beta {P::useJPerB ? std::log2(cell->parameters[CellParams::AMR_JPERB]) + logDx + P::JPerBModifier + refLevel : 0.0};
-            bool shouldRefine = cell->parameters[CellParams::AMR_ALPHA] > P::refineThreshold || beta > 0.5;
-            bool shouldUnrefine = cell->parameters[CellParams::AMR_ALPHA] < P::unrefineThreshold && beta < -0.5;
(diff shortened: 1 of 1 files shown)


lkotipal:
Old method, unrefine if neither index says to refine and either says to unrefine:
old_y_map_proton_vg_rho_0
![old_z_map_proton_vg_rho_0000102
000102](https://github.com/fmihpc/vlasiator/assets/12125199/88544a71-8698-46e5-8aea-b8c700033eb7)

projects/Magnetosphere/Magnetosphere.cpp
```diff
@@ -726,7 +726,7 @@ namespace projects {
             // We don't care about cells that are too far from the ionosphere
             const Real beta {P::useJPerB ? std::log2(cell->parameters[CellParams::AMR_JPERB]) + logDx + P::JPerBModifier + refLevel : 0.0};
             bool shouldRefine = cell->parameters[CellParams::AMR_ALPHA] > P::refineThreshold || beta > 0.5;
```
markusbattarbee, line 728: :oof: please do not use beta for checking alpha_2! Beta is a fundamental parameter of plasmas, there's a high risk of causing confusion with this notation. Admittedly, plasma beta cannot be negative,but still, in a plasma code let's resevere beta for plasma beta.

markusbattarbee (commented):
Note about parameter naming: But otherwise, I approve of this PR both for what it does and how the results look!

projects/Magnetosphere/Magnetosphere.cpp
lkotipal: Renamed `alphaTwo` to match the paper draft

markusbattarbee (approved):

lkotipal:
Breaks functionality when refining with a single parameter, merged into #831 with fix.

PR #839, opened 2023-10-31
Limit PR reports to 60kb in size.

ursg:
If they exceed 64kb, github refuses them. This mechanism just truncates them slightly before, which might not be ideal (since it chops of information at the end), but at least allows the PR to give the proper response code. Full testpackage output can still always be looked at as part of the uploaded artefact.

Note that this is a PR directly against the master branch, as the pr_report CI workflow always runs in the context of that branch, for unexplainable github security reasons.


markusbattarbee (commented):
Idea: could our testpackage STDOUT go to one file, STDERR to another, and in the uploaded info, first cat the STDOUT one, then the STDERR, and then truncate to 60kB? That way if it's truncated, we still get all the valuable stdout stuff.

PR #840, opened 2023-10-31
Try to squelch CI ucx warnings.

ursg:
(These were spamming the log files, thus exceeding the 64k check
 annotation limit)

Note that this is independent from the actual limiting of PR reports to 64k as part of PR #839


ursg:
Looks like this is working!

PR #841, opened 2023-10-31
Disable -ffast-math for building on carrington.

ursg:
Because fast math is bad for you and we shouldn't have had it there in the first place!

(and, in particular, not for CI test runs)


markusbattarbee:
For testpackage runs, `-ffast-math` is deactivated by the main Makefile at https://github.com/fmihpc/vlasiator/blob/d751cb0d0bf8fa54e24230d14fc9a57d32824473/Makefile#L10
Thus, let's close this, and instead make all `isnan` and `isinf`checks instead use `isfinite`, which according to https://en.cppreference.com/w/cpp/numeric/math/isfinite should not auto-return false with fast math active.

PR #842, opened 2023-11-02
Add CI stage to test testpackage cfg files

ursg:
Automatically run all project/ and testpackage/ cfg files through the check config script, so that they continue to stay up-to date.


markusbattarbee:
Superceded by #990

PR #843, opened 2023-11-02
Allow multiple boxes for refinement tests.

ykempf:
Here's my suggested addition of mutiple AMR boxes. Same stuff as in https://github.com/ursg/vlasiator/pull/66 ported to dev, whichever order we will end up merging and/or for others' pleasure.

I nuked two derived refinement functions which were copies of the project.cpp one.

And I do plan to allow for different shapes of the refinement blobs, that's coming up.


ykempf:
Yes so Flowthrough_amr changes because seems like I changed something with the box size. But I wanted to clean this up and get it up to date with dev before it rots too much, and as for different shapes as stated above, i seem to have nuked the repo where this was stashed so that'll be for whenever that's needed for debugging again.

PR #845, opened 2023-11-09
Switched all isnan and isinf calls to isfinites

markusbattarbee:


ursg:
Not working for me. I just wrote a small test program:
```c++
#include <iostream>
#include <limits>
#include <cmath>

int main() {

   double qnan = std::numeric_limits<double>::quiet_NaN();

   if(std::isnan(qnan)) {
      std::cout << "isnan(qnan) = true\n";
   } else {
      std::cout << "isnan(qnan) = FALSE\n";
   }

   if(std::isfinite(qnan)) {
      std::cout << "isfinite(qnan) = TRUE\n";
   } else {
      std::cout << "isfinite(qnan) = false\n";
   }

   return 0;
}
```
when compiled with ```g++ -O3 -o test test.cpp```, it correctly outputs:
```
isnan(qnan) = true
isfinite(qnan) = false
```
but if built with ```g++ -O3 -ffast-math -o test test.cpp```, I get:
```
isnan(qnan) = FALSE
isfinite(qnan) = TRUE
```

markusbattarbee:
Ok now this works as intended and with correct speed. And it shows that the `Ionosphere_small` test returns `NaN`s or `Inf`s in `vg_amr_du`and `vg_amr_jperb`!

markusbattarbee:
ok so PR #831 is going to fix those infs/nans, but what this now shows is that the testpackage script does not correctly identify a crashed run. @ursg How is that use if PIPESTATUS supposed to work? Trying now to enforce a failure mode if the output file is missing, but clearly this isn't working as intended.

markusbattarbee:
So now it seems to figure out a failed run.

ykempf:
Ouch, need new reference data or what?

PR #847, opened 2023-11-22
Fixed CI launcher arguments on Carrington, includes more performance metrics in output

markusbattarbee:


ykempf:
So this stops the yelling? But how's performance?

ykempf:
Ah it says in the report of course, close to 1, so fine indeed, though slower than more broken settings, correct?

markusbattarbee:
Yep, slightly slower indeed (a few percent, might fall into the system congestion category), but no yelling. And I believe you verified yesterday that these flags did not cause corrupt data on multi-node runs.
```
64.62        66.23         0.975691
```
Since this is all about getting our CI into better shape, instead of affecting performance on production runs, I think we can live with a small performance hit. But I'll add one more printout to the CI output, then happy with this. Just waiting for the servers to return.

ykempf:
Some are just above 1 too. :)
I can rerun just those flags in a few seconds once it's back, just to be sure.

ykempf:
Reconfirmed to not yell nor produce NaNs.

testpackage/small_test_carrington.sh
```diff
@@ -56,8 +56,8 @@ tasks_per_node=$(echo $units_per_node $t  | gawk '{print $1/$2}')
 export OMP_NUM_THREADS=$t
 
 #command for running stuff
-run_command="mpirun -mca pml ucx --mca btl ^vader,tcp,openib -x UCX_NET_DEVICES=mlx5_0:1 -x UCX_TLS=rc,sm -x UCX_IB_ADDR_TYPE=ib_global"
-small_run_command="mpirun -mca pml ucx --mca btl ^vader,tcp,openib -x UCX_NET_DEVICES=mlx5_0:1 -x UCX_TLS=rc,sm -x UCX_IB_ADDR_TYPE=ib_global -n 1 -N 1"
+run_command="mpirun --mca btl self -mca pml ^vader,tcp,openib,uct,yalla -x UCX_NET_DEVICES=mlx5_0:1 -x UCX_TLS=rc,sm -x UCX_IB_ADDR_TYPE=ib_global -np $tasks"
```
ykempf, line 59: So was this possibly not running in parallel? Though we don't use this scrip in CI anyway.

testpackage/small_test_carrington_github_ci.sh
```diff
@@ -86,7 +89,7 @@ flags=$(  $run_command $bin  --version |grep CXXFLAGS)
 solveropts=$(echo $flags|sed 's/[-+]//g' | gawk '{for(i = 1;i<=NF;i++) { if( $i=="DDP" || $i=="DFP" || index($i,"PF")|| index($i,"DVEC") || index($i,"SEMILAG") ) printf "__%s", $(i) }}')
 revision=$( $run_command $bin --version |gawk '{if(flag==1) {print $1;flag=0}if ($3=="log") flag=1;}' )
 
-$small_run_command $bin --version > VERSION.txt 2> $GITHUB_WORKSPACE/stderr.txt
+#$small_run_command $bin --version > VERSION.txt 2> $GITHUB_WORKSPACE/stderr.txt
```
ykempf, line 92: Not needed/done elsewhere?

testpackage/small_test_carrington_github_ci.sh
```diff
@@ -143,15 +146,18 @@ for run in ${run_tests[*]}; do
    test -e test_postproc.sh && ./test_postproc.sh
 
    ##Compare test case with right solutions
-   {
+   { {
    echo "--------------------------------------------------------------------------------------------"
    echo "${test_name[$run]}  -  Verifying ${revision}_$solveropts against $reference_revision"
    echo "--------------------------------------------------------------------------------------------"
+   } 2>&1 1>&3 3>&- | tee -a $GITHUB_WORKSPACE/stderr.txt;} 3>&1 1>&2 | tee -a $GITHUB_WORKSPACE/stdout.txt
```
ykempf, line 153: <3<3<3

ykempf (approved):

testpackage/small_test_carrington_github_ci.sh
markusbattarbee: Yeah not needed. Printing out version information into CI output could increase the size quite a bit, and well, it's inside CI so we should know which version is being run :)
The files in this directory will anyway be deleted on the next CI run.

testpackage/small_test_carrington.sh
markusbattarbee: Probably implicitly included, hadn't even notied that. I just edited the CI launch command and then copied it here.

testpackage/small_test_carrington_github_ci.sh
markusbattarbee: yeah fun with streams

PR #849, opened 2023-12-04
Add (manually triggered) CI workflow to generate new reference data.

ursg:


markusbattarbee:
@ursg should this have been merged into master in order to make the workflow available?

PR #856, opened 2023-12-04
Clang format trials

ykempf:
Just for laughs, I ran
```
clang-format -i --files {*,*/*,*/*/*,*/*/*/*}{.cpp,.h,.hpp}
```
so we see what comes out.

### DIFF
--- a/fieldsolver/fs_common.cpp
+++ b/fieldsolver/fs_common.cpp
@@ -52,7 +52,10 @@ Real divideIfNonZero(creal numerator, creal denominator) {
  */
 void reconstructionCoefficients(FsGrid<std::array<Real, fsgrids::bfield::N_BFIELD>, FS_STENCIL_WIDTH>& perBGrid,
                                 FsGrid<std::array<Real, fsgrids::dperb::N_DPERB>, FS_STENCIL_WIDTH>& dPerBGrid,
-                                std::array<Real, Rec::N_REC_COEFFICIENTS>& perturbedResult, cint i, cint j, cint k,
+                                std::array<Real, Rec::N_REC_COEFFICIENTS>& perturbedResult,
+                                cint i,
+                                cint j,
+                                cint k,
                                 creal& reconstructionOrder) {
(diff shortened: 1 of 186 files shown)


backgroundfield/backgroundfield.h
```diff
@@ -23,30 +23,21 @@
 #ifndef BACKGROUNDFIELD_H
 #define BACKGROUNDFIELD_H
 
-#include "fieldfunction.hpp"
-#include "../definitions.h"
 #include "../common.h"
+#include "../definitions.h"
+#include "fieldfunction.hpp"
```
ykempf, line 28: This means going alphabetic?

ykempf:
OK the number of conflicts seems to overwhelm the github interface so it'll have to go through the command line. Laalaalaa.

fieldsolver/fs_common.cpp
```diff
@@ -236,53 +277,57 @@ std::array<Real, 3> interpolatePerturbedB(
...
-               3 // Reconstruction order of the fields after Balsara 2009, 2 used for general B, but 3 used here to allow for cache reuse, see interpolatePerturbedJ below
+            reconstructionCoefficients(perBGrid,
+                                       dPerBGrid,
+                                       rc,
+                                       i,
+                                       j,
+                                       k,
+                                       3 // Reconstruction order of the fields after Balsara 2009, 2 used for general B,
+                                         // but 3 used here to allow for cache reuse, see interpolatePerturbedJ below
             );
```
ykempf, line 300: So I would prefer what's red in this spot, but TBH I'll get used to the style and/or run clang-format when needed as long as we agree on it and get going. :)

lkotipal: I agree, the green one looks pretty nasty.

ykempf:
I just stumbled an hour ago again across one spot where someone one day decied to have a few blocks with 2-space indentation and that annoys me every time again... And you'll be sweating when you get PRs from overseas with style you don't like. But OK, we can continue as is and comment and commit manually on missing curly braces and such fun.

PR #857, opened 2023-12-05
Clang format ldz electric field

ykempf:


ykempf:
So this works in principle, but we haven't settled on the style yet I guess.

ursg:
I'm still pretty unhappy with the way it indents openmp pragmas, and there is [currently no solution for it](https://bugs.llvm.org/show_bug.cgi?id=36020)

ykempf:
Which one do you mean specifically? That inside ifdef it doesn't indent?

fieldsolver/ldz_electric_field.cpp
```diff
@@ -1639,105 +1886,100 @@ void calculateElectricField(
...
+// Calculate upwinded electric field on inner cells
+#pragma omp parallel
    {
-      phiprof::Timer computeTimer {computeTimerID};
-      #pragma omp for collapse(2)
-      for (int k=0; k<gridDims[2]; k++) {
-         for (int j=0; j<gridDims[1]; j++) {
-            for (int i=0; i<gridDims[0]; i++) {
+      phiprof::Timer computeTimer{computeTimerID};
+#pragma omp for collapse(2)
```
ykempf, line 1934: No you said openmp, so you don't like it pushing all pragmas to the beginning of the line?

ursg: Exactly. I much prefer them to be aligned with the loop they apply to.

PR #858, opened 2023-12-05
Call MPI_Abort from all ranks in abort_mpi.

ursg:
This squeches a warning about "noreturn" function returning.

### DIFF
--- a/common.cpp
+++ b/common.cpp
@@ -96,7 +96,6 @@ void bailout(
       } else {
          std::cerr << __FILE__ << ":" << __LINE__ << ": " << str << std::endl;
       }
-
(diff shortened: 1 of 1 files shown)


ursg:
Ah, but now I'm wondering: if a different rank than the master reaches this error condition first, does it lead to the entire process set to be killed, thus never printing the error message?

markusbattarbee:
> Ah, but now I'm wondering: if a different rank than the master reaches this error condition first, does it lead to the entire process set to be killed, thus never printing the error message?

Perhaps.
So how about just making this function not be classified as `[[noreturn]]`? Any downside to that?

PR #860, opened 2023-12-07
Reintroduce maximum reflevel in AMR

lkotipal:
Fixes #831 accidentally reverting a change in #830.


lkotipal:
Should be yeah, I don't think any tests even do AMR

PR #863, opened 2023-12-11
Actually build on multiple cores on arriesgado.

ursg:
A slurm parameter for multiple cores was missing, so I suppose we were actually running on a single core before.

Let's see if this actually has a positive effect on riscv build times.


ursg:
Ok, hum, this is not considerably faster.

ursg:
Even though this is not faster, there's no reason *not* to merge this.

PR #864, opened 2023-12-12
Revert "Support for proper multi-unit read of fsgrids."

markusbattarbee:
Reverts fmihpc/vlasiator#581


markusbattarbee:
if https://github.com/fmihpc/vlsv/pull/53 works, we won't need this.

PR #865, opened 2023-12-12
Fs decomposition submodule

alhom:
Testing if submodule changes to runner fetch the new fsgrid properly.


tools/vlsvdiff.cpp
```diff
@@ -277,6 +280,52 @@ bool HandleFsGrid(const string& inputFileName,
 
 }
 
+bool getFsgridDecomposition(vlsvinterface::Reader& file, std::array<int,3>& decomposition){
```
markusbattarbee, line 283: Is it that you're keeping this also in this file because there's no parameters input for the override of decomposition?

.github/workflows/github-ci.yml
```diff
@@ -107,24 +107,25 @@ jobs:
...
       uses: actions/checkout@v3 
-    - name: Download libraries
-      uses: actions/download-artifact@v3
-      with:
-        name: libraries
-    - name: Unpack libraries
-      run: tar --zstd -xvf libraries.tar.zstd
-    - uses: ursg/gcc-problem-matcher@master
+      with: 
+        submodules: true
```
markusbattarbee, line 118: So is this now a temporary set-up, as you've only declared the fsgrid submodule, and all the other libraries come from the standard carrington_gcc_openmpi makefile? I guess the way ahead would be to set up the libraries tar file for those libraries which we don't control ourselves (e.g. vectorclass, zoltan, whatever) and set up the fmihpc libraries as submodules?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "submodules/fsgrid"]
+	path = submodules/fsgrid
+	url = git@github.com:fmihpc/fsgrid.git
```
markusbattarbee, line 3: and set a branch here?

markusbattarbee (commented):
More commentary

fieldtracing/fieldtracing.h
```diff
@@ -36,33 +36,33 @@ typedef float TReal;
 
 // Get the (integer valued) global fsgrid cell index (i,j,k) for the magnetic-field traced mapping point that node n is
 // associated with
-template<class T> std::array<int32_t, 3> getGlobalFsGridCellIndexForCoord(T& grid,const std::array<Real, 3>& x) {
```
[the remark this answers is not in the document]
alhom: Yeah, probably missed some after deciding to do it properly with typedefs.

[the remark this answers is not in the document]
alhom: Works as-is with 32 and 64 bit types, though.

fieldtracing/fieldtracing.cpp
```diff
@@ -528,7 +528,7 @@ namespace FieldTracing {
                   nodeStepCounter[n]++;
                   
                   // Check if the current coordinates (pre-step) are in our own domain.
-                  std::array<int, 3> fsgridCell = getLocalFsGridCellIndexForCoord(technicalGrid,{(TReal)x[0], (TReal)x[1], (TReal)x[2]});
+                  std::array<int64_t, 3> fsgridCell = getLocalFsGridCellIndexForCoord(technicalGrid,{(TReal)x[0], (TReal)x[1], (TReal)x[2]});
```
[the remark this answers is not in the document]
alhom: Right, so these things are fieldtracing helper functions with negative indices as null values, so need to be signed types.

parameters.cpp
```diff
@@ -327,6 +340,17 @@ bool P::addParameters() {
    RP::add("fieldsolver.minCFL",
            "The minimum CFL limit for field propagation. Used to set timestep if dynamic_timestep is true.", 0.4);
 
+   RP::add(
```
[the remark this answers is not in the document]
alhom: This way it is consistent with other XYZ inputs.

tools/vlsvdiff.cpp
alhom: It is there because the ioread interfaces ParallelReader, which is not trivially compatible with vlsvinterface::Reader, and besides changing that introduces a whole lot of new dependencies for vlsvdiff.

fieldtracing/fieldtracing.h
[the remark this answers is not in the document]
markusbattarbee: Just want to make sure we don't flood our build log with type mismatch warnings.

.github/workflows/github-ci.yml
alhom: Yup. We could also point submodules to commits on their respective repos (looks like everything is available via git*), or fork those under fmihpc. Besides, forking vectorclass would be handy to 'pre-install' the plugin header in a commit of our own.

.github/workflows/github-ci.yml
```diff
@@ -107,24 +107,25 @@ jobs:
 
   build_production:
     # Build Vlasiator witth production flags
-    runs-on: ubuntu-latest
-    container: ursg/vlasiator_ci:20230220_1
+    runs-on: carrington
+      #container: ursg/vlasiator_ci:20230220_1
```
ursg, line 111: If we move the whole build process onto carrington now, we could just drop the container creation (and the whole build_libraries stage) altogether, I suppose?

The nice thing about running in the container image on azure VMs so far was just that they are free and readily available, even if our own runners are stalled for whatever reason.

.github/workflows/github-ci.yml
ursg: Hmm, that is indeed an interesting idea. It'll also be less of a moving target for us to track then.
(On the other hand, if there are ever any important updates for some new architecture, and we only update *then*, we might have a whole bunch of things to fix on our hands)

.github/workflows/github-ci.yml
```diff
@@ -107,24 +107,25 @@ jobs:
...
+          #      with:
+          #        name: libraries
+          #    - name: Unpack libraries
+          #      run: tar --zstd -xvf libraries.tar.zstd
+          #    - uses: ursg/gcc-problem-matcher@master
     - name: Compile vlasiator (Release build)
       run: |
-        VLASIATOR_ARCH=github_actions make clean
-        VLASIATOR_ARCH=github_actions make -j 3
+        srun -M carrington --job-name tp_compile --interactive --nodes=1 -n 1 -c 16 --mem=40G -p short -t 0:10:0 bash -c 'module purge; module load GCC/11.2.0; module load OpenMPI/4.1.1-GCC-11.2.0 ; module load PMIx/4.1.0-GCCcore-11.2.0; module load PAPI/6.0.0.1-GCCcore-11.2.0; export VLASIATOR_ARCH=carrington_gcc_openmpi; make clean; make -j 16'
```
ursg, line 128: The job naim should probably not be "tp_compile" here, if it is building a release build.

MAKE/Makefile.arriesgado
```diff
@@ -48,7 +48,7 @@ LIB_ZOLTAN = -L${LIBRARY_PREFIX}/lib -lzoltan
 INC_VLSV = -I$(LIBRARY_PREFIX)/vlsv
 LIB_VLSV = -L$(LIBRARY_PREFIX)/vlsv -lvlsv
 
-INC_DCCRG = -I${LIBRARY_PREFIX}/dccrg -I${LIBRARY_PREFIX}/fsgrid
+INC_DCCRG = -I${LIBRARY_PREFIX}/dccrg -I./submodules/fsgrid
```
ursg, line 51: Why is that fsgrid path in the INC_DCCRG variable here anyway?
Ah well, I guess that one's on me...

sysboundary/ionosphere.cpp
```diff
@@ -1149,7 +1149,7 @@ namespace SBC {
             }
 
             // Local cell
-            std::array<int,3> lfsc = getLocalFsGridCellIndexForCoord(technicalGrid,nodes[n].xMapped);
+            std::array<int64_t,3> lfsc = getLocalFsGridCellIndexForCoord(technicalGrid,nodes[n].xMapped);
```
ursg, line 1152: I guess this should also be ```FsGridTools::FsIndex_t``` ?

ursg (commented):
I like the submodules! Totally in favour of using it for all our header libraries, at least.

As I took this opportunity to also look through the fs decomposition changes, I wondered: how do we get this information into the visit plugin then?
But then I realized: we are *already* supplying the visit plugin with an entirely separate set of decomposition information, as part of the MESH property of each fsgrid, each of which is containing a separate copy of what is essentially the CellID mechanism (see the implementation [here](https://github.com/fmihpc/vlasiator/blob/master/iowrite.cpp#L943-L976)). Wow, aren't we wasteful with totally redundant information there?

sysboundary/ionosphere.cpp
alhom: Yup, these are coming in...

alhom:
Nope, didn't get the fsgrid commit like so...

.github/workflows/github-ci.yml
markusbattarbee: What about submodules that require building, like vlsv?

I like the idea of verifying our library and build process on a cloud runner.

markusbattarbee:
> Nope, didn't get the fsgrid commit like so...

Is the indentation in `.gitmodules`broken perhaps?

.github/workflows/github-ci.yml
alhom: > What about submodules that require building, like vlsv?
We're doing that anyway, just needs to point the build process there. _And_ then the library build process can also be entirely contained in the main repo!

MAKE/Makefile.arriesgado
markusbattarbee: daft indeed, but might as well fix here.

.github/workflows/github-ci.yml
markusbattarbee: > . _And_ then the library build process can also be entirely contained in the main repo!

A bit of a double-edged sword there, perhaps.

alhom: How so? As I see it, this would mean taking Yann's build_libraries.sh, including it in the repo and pointing it to submodules.

alhom:
Okay, anything remaining with this here PR? All makefiles now have the submodule path to fsgrid, and vlsvdiff should write dummy fsgrid decomposition data... but that particular tool doesn't give me now any `--diff` outputs, even on dev (trying with restart_write outputs..).

.github/workflows/github-ci.yml
markusbattarbee: Hm. Well, perhaps ok indeed. We then need to create a carrington makefile for vlsv and phiprof.

ioread.cpp
```diff
@@ -812,25 +858,48 @@ template<unsigned long int N> bool readFsGridVariable(
...
-   if(size == numWritingRanks) {
-      // Easy case: same number of tasks => slurp it in.
+   std::array<FsGridTools::Task_t,3> fileDecomposition={0,0,0};
+   // No override given (zero array)
+   if(P::overrideReadFsGridDecomposition == fileDecomposition){
+      // Try and read the decomposition from file
+      if (readFsgridDecomposition(file, fileDecomposition) == false) {
+         exitOnError(false, "(RESTART) Failed to get Fsgrid decomposition", MPI_COMM_WORLD);
+      }
+   } else {
```
markusbattarbee, line 877: So... now that you found the GlobalIDs in the .vlsv files, do we even need this manual override for reading any more?

ioread.cpp
```diff
@@ -1389,3 +1409,102 @@ bool readFileCells(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
...
+   uint64_t arraySize;
+   uint64_t vectorSize;
+   vlsv::datatype::type dataType;
+   uint64_t byteSize;
+
+   int myRank;   
+   MPI_Comm_rank(MPI_COMM_WORLD,&myRank);
+
+
+   phiprof::Timer computeDomainDecomposition {"computeDomainDecomposition"};
```
markusbattarbee, line 1425: this timer is now named incorrectly

ioread.cpp
```diff
@@ -1389,3 +1409,102 @@ bool readFileCells(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
...
+         x_corners.insert(inds[0]);
+         y_corners.insert(inds[1]);
+         z_corners.insert(inds[2]);
+         ++ids_ptr;
+         begin_rank += rank_size;
+      }
+
+      decomposition[0] = x_corners.size();
+      decomposition[1] = y_corners.size();
+      decomposition[2] = z_corners.size();
```
markusbattarbee, line 1489: Ha! Nifty.

ioread.cpp
```diff
@@ -1389,3 +1409,102 @@ bool readFileCells(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
...
+
+
+bool readFsgridDecomposition(vlsv::ParallelReader& file, std::array<FsGridTools::Task_t,3>& decomposition){
+   list<pair<string,string> > attribs;
+   uint64_t arraySize;
+   uint64_t vectorSize;
+   vlsv::datatype::type dataType;
+   uint64_t byteSize;
+
+   int myRank;
```
markusbattarbee, line 1421: I wonder if it'd be better to perform this with MASTER_RANK only, and then communicate the data? Instead of every task reading numWritingRanks single values from the file at non-contiguous positions. Or will the `file.read`do reading and MPI distribution under the hood?

ioread.cpp
alhom: Not really, but it has potential debugging value if something breaks down the line.

ioread.cpp
alhom: It is piped to the collective MPI IO, but indeed maybe better to be safe and do it from master only.

.github/workflows/github-ci.yml
```diff
@@ -213,7 +220,9 @@ jobs:
 
     steps:
     - name: Checkout source
-      uses: actions/checkout@v3 
+      uses: actions/checkout@v3
+      with:
```
markusbattarbee, line 224: technically, this step does not need the submodules.

.github/workflows/github-ci.yml
```diff
@@ -123,8 +121,11 @@ jobs:
     - uses: ursg/gcc-problem-matcher@master
     - name: Compile vlasiator (Release build)
       run: |
-        VLASIATOR_ARCH=github_actions make clean
-        VLASIATOR_ARCH=github_actions make -j 3
+         VLASIATOR_ARCH=github_actions make clean
+         VLASIATOR_ARCH=github_actions make -j 3
+   #  - name: Compile vlasiator (Release build)
```
markusbattarbee, line 126: If this works, we can prune these comments

.github/workflows/github-ci.yml
```diff
@@ -250,23 +259,26 @@ jobs:
...
     # Build IonosphereSolverTests miniApp (currently broken)
-    runs-on: ubuntu-latest
-    container: ursg/vlasiator_ci:20230220_1
+    runs-on: carrington
+      #container: ursg/vlasiator_ci:20230220_1
     needs: build_libraries
     steps:
     - name: Checkout source
       uses: actions/checkout@v3 
-    - name: Download libraries
```
markusbattarbee, line 259: here as well the libraries should be downloaded ( and I guess we could return it to the cloud)

PR #866, opened 2023-12-13
New boundary padding which works with dipole type 4 and the new time-varying boundaries

markusbattarbee:
This supercedes #791.
Hasn't yet been tested.


markusbattarbee:
I'm actually now (once again) wondering about this parameter: `Magnetosphere.noDipoleInSW`. 

If I recall correct, we wanted it so that in 5D runs, the inflow would be a pure IMF, and the flushing solar wind wouldn't end up having dipole remnants. How with 3D Dipoles or even the vector dipole which scales away by the inflow value, do we want to use this parameter? I'd guess not...

common.h
```diff
@@ -257,7 +257,7 @@ namespace bvolderivatives {
  * back and forth.
  */
 namespace fsgrids {
-   enum bfield {
+   enum bfield : int {
```
ykempf, line 260: For my own edification, what does this `int?` do? Aren't enums mapping to ints anyway?

projects/Magnetosphere/Magnetosphere.cpp
```diff
@@ -336,15 +336,30 @@ namespace projects {
...
+               setPerturbedField(bgFieldDipole, BgBGrid, fsgrids::bgbfield::BGBXVDCORR);
+               bgVectorDipole.initialize(8e15 *this->dipoleScalingFactor, 0.0, 0.0, 0.0, this->dipoleTiltPhi*M_PI/180., this->dipoleTiltTheta*M_PI/180., this->dipoleXFull, this->dipoleXZero, this->dipoleInflowB[0], this->dipoleInflowB[1], this->dipoleInflowB[2]);
+               setPerturbedField(bgVectorDipole, BgBGrid, fsgrids::bgbfield::BGBXVDCORR, true);
                if (P::isRestart == false) {
-                  bgFieldDipole.initialize(-8e15 *this->dipoleScalingFactor, 0.0, 0.0, 0.0, 0.0 );
-                  setPerturbedField(bgFieldDipole, perBGrid);
-                  bgVectorDipole.initialize(8e15 *this->dipoleScalingFactor, 0.0, 0.0, 0.0, this->dipoleTiltPhi*M_PI/180., this->dipoleTiltTheta*M_PI/180., this->dipoleXFull, this->dipoleXZero, this->dipoleInflowB[0], this->dipoleInflowB[1], this->dipoleInflowB[2]);
-                  setPerturbedField(bgVectorDipole, perBGrid, true);
+                  // If we are starting a new simulation, we also copy this data into perB.
+                  const auto localSize = BgBGrid.getLocalSize().data();
```
ykempf, line 350: Add a schedule?

vlasiator.cpp
```diff
@@ -485,6 +485,13 @@ int main(int argn,char* args[]) {
       *project
    );
    
+   // There are projects that have non-uniform and non-zero perturbed B, e.g. Magnetosphere with dipole type 4.
+   // For inflow cells (e.g. maxwellian), we cannot take a FSgrid perturbed B value from the templateCell,
+   // because we need a copy of the value from initialization in both perBGrid and perBDt2Grid and it isn't
+   // touched as we are in boundary cells for components that aren't solved. We do a straight full copy instead
+   // of looping and detecting boundary types here.
+   perBDt2Grid = perBGrid;
```
ykempf, line 493: So with time-varying inflow, are we keeping this as is or should we have perBDt2Grid obtain the potentially different values from the time-variation?

ykempf (commented):
:christmas_tree: :gift:

common.h
markusbattarbee: They *should* map to the smallest integer type which can contain the whole enum. However, with this templating business, I was getting issues with the correct template type not being found, so had to declare this explicitly to ensure compilation. A bit weird, but no harm.

lkotipal: Probably, but it turns out that's impementation defined.

vlasiator.cpp
markusbattarbee: Hmm. You're correct, perBDt2 needs to also get the correct values at the inflow. Hmm.

common.h
ykempf: OK so template+compiler taming, fine by me.

vlasiator.cpp
ykempf: Sorry... TBH this now sounds like a massive can of worms – we don't want to replace the dt/2 B entries in the RK algorithm with these interpolated values, do we? And if we do it'll be a massive PITC. And that wasn't done by the OP so I'd leave this as is, maybe with a new issue for our archive and a request for @henry2004y's opinion on this? :)

markusbattarbee: Well this is one of those things that the *correct* way would be to set up boundaries with the absolutely correct time indeed, to retain restart bit-identicality. However, the value of dt/2 should in most use scenarios be quite small with respect to the timescales of the time-varying inflow.

markusbattarbee: On the other hand, at restarts we anyway only read one value of perb and one for E, not the _dt2 values. So...

ykempf: So I think I am wrong on this one, the half-timestep values are computed in the solver so we're good here, and shall live with non-half dt values for moments.

projects/Magnetosphere/Magnetosphere.cpp
markusbattarbee: default is same as for `schedule(static,1)`, and this just copies over a few FSgrid values - not worth it.

vlasiator.cpp
markusbattarbee: Indeed this will now update the state of the BC pre-translation, and use that value for both for the perB and the perBDt2 fields. But it's slow evolution so should be fine, and works the same for restarts and non-restart runs.

markusbattarbee:
Tested and appears to work as intended.

backgroundfield/backgroundfield.h
```diff
@@ -38,15 +39,98 @@ void setBackgroundFieldToZero(
...
 
-void setPerturbedField(
-   const FieldFunction& bgFunction,
-   FsGrid< std::array<Real, fsgrids::bfield::N_BFIELD>, FS_STENCIL_WIDTH> & perBGrid,
-   bool append=false
-);
+/**
+   Templated function for setting the perturbed B field to zero.
+   Function is templated so it may, if necessary, be called to set the
+   vector dipolel correction terms inside the background field FSgrid
```
ykempf, line 45: dipole
and could take the doxygen formatting

sysboundary/sysboundarycondition.cpp
```diff
@@ -899,6 +899,7 @@ namespace SBC {
                if( technicalGrid.get(i+ii,j+jj,k+kk) // skip invalid cells returning NULL
                    && (technicalGrid.get(i+ii,j+jj,k+kk)->SOLVE & mask) == mask // Did that guy solve this component?
                    && technicalGrid.get(i+ii,j+jj,k+kk)->sysBoundaryFlag != sysboundarytype::DO_NOT_COMPUTE // Do not copy from there
+                   &&  technicalGrid.get(i+ii,j+jj,k+kk)->sysBoundaryFlag != sysboundarytype::OUTER_BOUNDARY_PADDING // Do not copy from there either
```
ykempf, line 902: space space

sysboundary/sysboundarycondition.cpp
```diff
@@ -912,6 +913,7 @@ namespace SBC {
...
```
ykempf, line 916: too too

testpackage/tests/Magnetosphere_3D_small/Magnetosphere_3D_small.cfg
```diff
@@ -114,7 +114,7 @@ file_x+ = sw1.dat
 constBgBX = 0.0
 constBgBY = 0.0
 constBgBZ = -5.0e-9
-noDipoleInSW = 1.0
+noDipoleInSW = 0 # Use 0 with dipole type 4
```
ykempf, line 117: Make this foolproofer and check in the code?

ykempf (approved):
mergatur!

PR #867, opened 2023-12-18
Store FSgrid globalIDs only when writing FSgrid output

markusbattarbee:
This PR makes it so that regular FSgrid metadata is written always, but the more disk consuming list of "GlobalIDs" is only stored if FSgrid output is active.


ursg (approved):
Looking gold, testpackage is green too.

PR #868, opened 2024-01-05
Use multiple handlers of device mesh wrapper to avoid relocatable code device builds

sfantao:
This PR remove the relocatable code build and link options for hipcc builds and only does host code linking using clang++.

To avoid further code disruption, the idea pursued is to let each compilation unit have their own instance of the device pointer to the device mesh wrapper in constant memory and use static object Ctors to facilitate the registration of all these instances so that each can be initialized.

The downside is that there will be extra pointers in device memory (66 at the moment - 528 bytes) the upside is that parallel builds are a small fraction of the time they used to be. There should't be any performance degradation.

Portability of the approach needs to be verified to NVIDIA toolchain.

### DIFF
--- a/MAKE/Makefile.lumi_hipcc
+++ b/MAKE/Makefile.lumi_hipcc
@@ -1,5 +1,5 @@
 CMP = hipcc
-LNK = hipcc
+LNK = clang++
 
 # Modules loaded (after clean shell, no module purging, one-by-one not oneline)
 # module load LUMI/22.08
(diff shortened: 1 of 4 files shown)


markusbattarbee:
Great! With the CUDA toolchain I can confirm this compiles and runs. Only needed to turn off the `-dc` flag.

I also did this diff:
```
diff --git a/velocity_mesh_parameters.h b/velocity_mesh_parameters.h
index ab4da9fa..82f02771 100644
--- a/velocity_mesh_parameters.h
+++ b/velocity_mesh_parameters.h
@@ -120,7 +120,7 @@ namespace vmesh {
    // address, we use the Ctor of a static object to register all intances so that the allocated memory pointer
    // could be copied to all of them.
    __device__ __constant__ MeshWrapper* meshWrapperDevInstance;
-   ARCH_HOSTDEV static MeshWrapper* gpu_getMeshWrapper() { return meshWrapperDevInstance; };
+   ARCH_DEV static MeshWrapper* gpu_getMeshWrapper() { return meshWrapperDevInstance; };
    // Static object and corresponding instance whose Ctor is used register all the instances of device meshWrapperDev
    // symbols.
    struct meshWrapperDevRegistor {
```
to get rid of some compiler warnings. And of course we'll want to squelch some of those printouts in production.

Were you able to pass a full testpackage test with this on AMD/HIP? I ask because on Mahti CUDA all other tests complete successfully but the 3D AMR tests fail. (I'll continue to check if it's some unrelated bug I've fixed in a under-development feature branch)

markusbattarbee:
Was able to get those tests to pass, so I trust this, merging it. Thank you!

PR #869, opened 2024-01-07
Submodule treatment for dccrg

alhom:
Continuing the submoduling of header libraries after #865.

Added DCCRG as a submodule, makefiles updated.


markusbattarbee:
Worth merging dev into this I guess and making the yaml file match how the fsgrid PR ended up

markusbattarbee (approved):
Looks good!

ykempf:
Well, you asked for my input... I would have appreciated if the Makefiles had a variable for the path to submodules and not the hardcoded `./submodules/` on several lines. But if we don't want to go for that, I can do the click...

.gitmodules
```diff
@@ -3,3 +3,6 @@
 	url = git@github.com:fmihpc/fsgrid.git
 
 
+[submodule "submodules/dccrg"]
```
ykempf, line 6: So how does this work? I don't see where it specifies that it should land on the `vlasiator-version` branch.

submodules/dccrg
alhom, line 1: @ykempf Here - these point to commits primarily.

ykempf: Well hidden and in particular not highlighted in green.

PR #870, opened 2024-01-09
Patch to read in float fields.

ykempf:
Can also go into #627 if preferred.

### DIFF
--- a/particles/readfields.h
+++ b/particles/readfields.h
@@ -119,20 +119,37 @@ std::vector<double> readFieldData(Reader& r, std::string& name, unsigned int num
       exit(1);
    }
 
-   if(dataType != vlsv::datatype::type::FLOAT || byteSize != 8 || vectorSize != numcomponents) {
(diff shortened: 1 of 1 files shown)


ykempf:
This came up when trying to extract the fluxfunction out of runs which were written as float.

particles/readfields.h
```diff
@@ -119,20 +119,37 @@ std::vector<double> readFieldData(Reader& r, std::string& name, unsigned int num
       exit(1);
    }
 
-   if(dataType != vlsv::datatype::type::FLOAT || byteSize != 8 || vectorSize != numcomponents) {
```
markusbattarbee, line 122: Shouldn't there still be a check for:
```
if(dataType != vlsv::datatype::type::FLOAT || vectorSize != numcomponents) {

ursg: We *could* hypothetically plot the technical grid? The particle pusher might, for example, extract boundary information from there?

PR #874, opened 2024-01-16
Swap left-right to right-left in curvature calculation (sign error fix)

alhom:


markusbattarbee (approved):

ykempf (approved):
:crossed_swords: or so :eyes: looking good!

PR #879, opened 2024-01-23
Add config option to copysphere for setting static B at inner boundary

JonasSuni:
Edits to the files copysphere.cpp and copysphere.h to support a new flag "zeroPerB" under the copysphere option category in config files. zeroPerB is 0 by default, and in this case copysphere behaves as before. If set to 1, zeroPerB will set the magnetic field at the inner boundary to always be the original background B.


JonasSuni:
I'll run some tests to actually test the functionality soon

sysboundary/copysphere.cpp
```diff
@@ -59,6 +59,7 @@ namespace SBC {
       Readparameters::add("copysphere.geometry", "Select the geometry of the copysphere, 0: inf-norm (diamond), 1: 1-norm (square), 2: 2-norm (circle, DEFAULT), 3: 2-norm cylinder aligned with y-axis, use with polar plane/line dipole.", 2);
       Readparameters::add("copysphere.precedence", "Precedence value of the copysphere system boundary condition (integer), the higher the stronger.", 2);
       Readparameters::add("copysphere.reapplyUponRestart", "If 0 (default), keep going with the state existing in the restart file. If 1, calls again applyInitialState. Can be used to change boundary condition behaviour during a run.", 0);
+      Readparameters::add("copysphere.zeroPerB","Keep magnetic field static at the inner boundary",0);
```
ykempf, line 62: Please add the explicit mention that this has to be set to 1 to activate the feature.

ykempf (approved):

PR #880, opened 2024-01-23
Flowthrough AMR

lkotipal:
Remove `canRefine` and `adaptRefinement` from Flowthrough, defaulting to using the same AMR as Magnetosphere.


ykempf (approved):
Yes, I believe this is equivalent to a fix I used in some tests myself.

PR #881, opened 2024-01-23
Face neighbor caching branch

lkotipal:
See https://github.com/fmihpc/dccrg/pull/33


.gitmodules
```diff
@@ -5,4 +5,5 @@
 
 [submodule "submodules/dccrg"]
 	path = submodules/dccrg
-	url = https://github.com/fmihpc/dccrg.git
```
markusbattarbee, line 8: So I think we shouldn't need to edit this file in order to get a specific branch from a specific repo to test. We didn't need it for the FSgrid decomposition fix - the submodule commit update should handle it, even if it's a remote repo as long as it's forked from this one.

markusbattarbee: so please remove the below lines :)

markusbattarbee: Unless I'm wholly mistaken, the reference (submodules -> dccrg) is to a commit, so when the DCCRG PR gets merged, it still points to the same commit. I'm not sure how the mechanism also searches for commits from forks, but it seems to work. I believe Markku verified that having the "branch" line here online 9 does not do anything, just misleads. And I think we want our dev branch to always point to fmihpc/submodule, not someone's personal fork.

lkotipal: Yeah, but that drops the merge commit. If there's any commits on the master branch (or vlasiator-version in this case) after branching off I'm not sure if those would get dropped.

alhom: No need to touch `.gitmodules`, though. You can update the vlasiator submodule to point to the merge commit after that is available, as well.

The branch tag should help with default `pull`/`clone` behaviour, but it's not relevant here.

lkotipal:
Submodules having a normal one, opened a new PR #933

PR #882, opened 2024-01-24
Adjust blocks at sysboundaries

markusbattarbee:
Whenever a sysboundary template cell (or for ionosphere, an actual cell) is updated, it goes through block adjustment to add the necessary buffer zones. The comment states that unnecessary blocks would also be pruned, but that is not the case. This PR remedies that. Might help with artefacts along side walls in low-res-vspace simulations.

### DIFF
--- a/sysboundary/donotcompute.cpp
+++ b/sysboundary/donotcompute.cpp
@@ -79,8 +79,9 @@ namespace SBC {
          
          //let's get rid of blocks not fulfilling the criteria here to save
          //memory.
-         for (uint popID=0; popID<getObjectWrapper().particleSpecies.size(); ++popID)
-            cell->adjustSingleCellVelocityBlocks(popID);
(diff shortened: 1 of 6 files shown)


ursg:
Looks good structurally. I guess the nonzero testpackage diffs again stem from the fact that our reference data is kinda outdated?

sysboundary/donotcompute.cpp
```diff
@@ -79,8 +79,9 @@ namespace SBC {
          
          //let's get rid of blocks not fulfilling the criteria here to save
          //memory.
-         for (uint popID=0; popID<getObjectWrapper().particleSpecies.size(); ++popID)
-            cell->adjustSingleCellVelocityBlocks(popID);
+         for (uint popID=0; popID<getObjectWrapper().particleSpecies.size(); ++popID) {
+            cell->adjustSingleCellVelocityBlocks(popID,true);
```
ykempf, line 83: Why do we even need to call it in DO_NOT_COMPUTE? But it won't hurt.

PR #883, opened 2024-01-24
Mem and cell reports

ykempf:
Here's a new function to report spatial cell counts per refinement level as well as velocity blocks per population and refinement level into the logfile at the same time as memory reports.

Includes the addition of t and timestep in the MEM lines of the logfile for better grepping and plotting.


markusbattarbee:
Excellent!
Now I agree with the established tradition that the walltime estimates etc should come not-too-often, i.e. every 10th diagnostic output. However, perhaps cell counts and memory usage should be output at the diagnostic interval? That's what I would suggest.

ykempf:
Olepa hyvä.

markusbattarbee (approved):

PR #885, opened 2024-01-26
add submodule use to ref data generator

markusbattarbee:


ursg:
I guess we need to merge it to see whether it works?

markusbattarbee:
Yeah, as well as deleting old submodules from all CI runners before running this

ursg:
Just checked the documentation, I don't believe separate cleaning of "old" submodules is necessary here. So in that case, this can go right in.

PR #890, opened 2024-01-30
Added Eigen as a submodule

alhom:
Still needs Makefiles update.


ursg:
Fun, this now fails compiling on appleM!, because apparently, C++17 deprecated and removed the std::binder1st calls that this Eigen branch uses.

alhom:
Eigen:
> Notice: that 3.4.x will be the last major release series of Eigen that will support c++03. The master branch will drop c++03 support after this release.

Well, let's see if 3.4.0 works still.

markusbattarbee:
Should we use `-isystem` for Eigen instead, as it's not *our* library?

ursg:
Note that the build_library.sh script before was just using whatever latest bleeding edge version of Eigen, and that seemed to have worked fine recently.

alhom:
> Should we use `-isystem` for Eigen instead, as it's not _our_ library?

Yup. Maybe it's now less-spammy anyway with warnings, but still good.

> Note that the build_library.sh script before was just using whatever latest bleeding edge version of Eigen, and that seemed to have worked fine recently.

And the initial commit set Eigen to track the 3.2 branch so that's a bit weird. I think I took that from some of our build setup, and it indeed is quite old.

ursg (approved):

markusbattarbee:
Interestingly now the testpackage shows a huge amount of numerical low-level diffs even in tests where we don't usually see them. So has a newer version of eigen changed some underlying algorithm which now provides different results?!?

alhom:
And noting here that my commit messages do mean 3.4.0 Eigen...

markusbattarbee (approved):

markusbattarbee:
Commenting for posterity - the numerical level non-zero diffs in e.g. acceleration tests in this PR are indeed caused by a version difference in Eigen, and we'll now switch to the newer 3.4.0 commit.

PR #895, opened 2024-01-30
Move submodule include paths

lkotipal:
Move submodule include paths to the main Makefile, before including architecture-specific Makefiles. This reduces code repetition, while allowing the include paths to be overridden if you actually want to do that.


markusbattarbee:
Instead of simply hard-coding these directories - how about setting them to those values assuming the INC_XXX value is not set? That way an user could still override if they want. (And we want this to also do the same for vectorclass, I think)

lkotipal:
That's already how it works in this PR. Submodule includes are before `include MAKE/Makefile.${ARCH}`, so if someone points to their own fsgrid path in an architecture-specific makefile it overrides that.

markusbattarbee:
Ah, good. How about a comment to that end? Also, I now see plenty of conflicts.

lkotipal:
Conflicts fixed btw, they were from the addition of vectorclass which got the same treatment

alhom:
Hum. Should this, at the same time, set all other library paths similarly, according to the default build target folders of build_libraries.sh?

markusbattarbee:
> Hum. Should this, at the same time, set all other library paths similarly, according to the default build target folders of build_libraries.sh?

I'd say no. Build_libraries.sh is a nice tool and can be used by our CI, but I think we still want to continue to have centrally  managed libraries on the big systems.

alhom:
Fair enough! Though, IMO it would be a logical continuation to have the libraries as well as their build process tied to the production branches and logged there as commits. Eventually. Maybe.

PR #899, opened 2024-01-31
Use newer docker image in CI runs.

ursg:
The old one was almost a year old now.
This potentially now brings in newer compiler versions and more helpful warnings.


ursg:
Looks like this works just as well as the old image, and doesn't even produce any extra warnings. Ah well. Still makes sense to stay up-to-date with the times.

PR #900, opened 2024-02-01
Ensure abort_mpi does not return

lkotipal:
Prevent execution of `abort_mpi` reaching the end of the function and returning void.


ursg:
Note that this abort_mpi() might get called on a single rank only, and not necessarily all of them at the same time. So in that situation the MPI_ABORT *might* kill the whole thing, or instead just hang on the barrier indefinitely.

Anyway, the simulation would be toast in that situation one way or the other, so we can't do much more that count on correct behaviour by the MPI implementation, so I guess it's the best we can do.

lkotipal:
Are we not ensuring all ranks call `abort_mpi`? In that case the function itself is malformed, as it only aborts if `MASTER_RANK` calls it...

ursg:
> Are we not ensuring all ranks call `abort_mpi`? In that case the function itself is malformed, as it only aborts if `MASTER_RANK` calls it...

At least [some calls of it](https://github.com/fmihpc/vlasiator/blob/master/sysboundary/inflow.cpp#L127) are certainly such that they could trigger only on a single rank somewhere.

markusbattarbee:
This now spams the error message from all ranks, instead of just master_rank like before. Intended? Good or bad?

lkotipal:
> This now spams the error message from all ranks, instead of just master_rank like before. Intended? Good or bad?

Intended, since critical sections aren't really supported by MPI

PR #901, opened 2024-02-02
Github actions workflow to automatically sync branches to HLRS gitlab

ursg:
Right now, this would keep the dev, master and cudasiator branches up to date there, so that development is visible to EuroHPC and can potentially trigger CI runs on their hardware, too.

I guess this can only really be tested by merging into dev, because why would it otherwise trigger?


markusbattarbee:
The modern GPU development branch is not cudasiator, it's vlasiator_gpu

PR #902, opened 2024-02-02
AMR Testpackage

lkotipal:
Basic test for adaptive mesh refinement in Flowthrough. This should cover refining, coarsening, outer boundaries, induces and overrides.


markusbattarbee:
In principle, this looks good. However, I think the block count should be larger (or acceleration should be turned off). For a very coarse v-space even small numerical diffs can grow significantly, which can impact the alphas and thus mesh refinement.

ykempf:
So after some discussion with Markus: is it becoming a lot slower if you set the box initial temperature to 1e6 instead of 1e5? What are reported block counts (you can post a path to the logfile on zulip if you prefer)?

markusbattarbee (approved):

testpackage/test_definitions_small.sh
```diff
@@ -142,10 +142,18 @@ variable_names[${index}]="proton/vg_rho proton/vg_v proton/vg_v proton/vg_v fg_b
 variable_components[${index}]="0 0 1 2 0 1 2 0 1 2"
 ((index+=1))
 
+# 14 Large (dynamic) AMR translation flowthrough test
+test_name[${index}]="Flowthrough_damr"
+comparison_vlsv[${index}]="bulk.0000001.vlsv"
+comparison_phiprof[${index}]="phiprof_0.txt"
+variable_names[${index}]="proton/vg_rho proton/vg_v proton/vg_v proton/vg_v fg_b fg_b fg_b fg_e fg_e fg_e"
```
markusbattarbee, line 149: How about at least alpha1 and alpha2 as extra output variables? Others?

testpackage/tests/Flowthrough_damr/Flowthrough_damr.cfg
```diff
@@ -0,0 +1,113 @@
...
+
+[proton_Flowthrough]
+T = 1.0e5
+rho  = 1.0e6
+VX0 = 1e5
+VY0 = 0
+VZ0 = 0
+
+[loadBalance]
+algorithm = RCB
```
markusbattarbee, line 109: Should the LB cadence be increased, in addition to having a LB cadence of 2?

testpackage/tests/Flowthrough_damr/Flowthrough_damr.cfg
```diff
@@ -0,0 +1,113 @@
...
+x_length = 16
+y_length = 8
+z_length = 8
+x_min = -8e7
+x_max = 8e7
+y_min = -4e7
+y_max = 4e7
+z_min = -4e7
+z_max = 4e7
+t_max = 182.0
```
markusbattarbee, line 31: The pulse propagates about 22% of the simulation domain here. I would recommend propagating it about 80% and adding output files at 20,40,60, and 80% extents.

testpackage/tests/Flowthrough_damr/Flowthrough_damr.cfg
```diff
@@ -0,0 +1,113 @@
...
+[proton_vspace]
+vx_min = -1.92e6
+vx_max = +1.92e6
+vy_min = -1.92e6
+vy_max = +1.92e6
+vz_min = -1.92e6
+vz_max = +1.92e6
+vx_length = 16
+vy_length = 16
+vz_length = 16
```
markusbattarbee, line 43: So this is a low block content. How long does this take to run on Carrington?

markusbattarbee (commented):
Some questions and suggestions.

testpackage/tests/Flowthrough_damr/Flowthrough_damr.cfg
lkotipal: Well on Mahti it's 66s, so about 150% of Flowthrough_amr

markusbattarbee: Hmm. And if gridbuilder.t_max is increased to 722, it'll be at least 4x or more like 5x that. :-/

testpackage/tests/Flowthrough_damr/Flowthrough_damr.cfg
lkotipal: Adjusted per recommendation.

testpackage/tests/Flowthrough_damr/Flowthrough_damr.cfg
```diff
@@ -0,0 +1,113 @@
+ParticlePopulations = proton
+
+project = Flowthrough
+propagate_field = 1
+propagate_vlasov_acceleration = 1
```
markusbattarbee, line 5: I suggest turning acceleration off - at least trialling if it still works?

testpackage/tests/Flowthrough_damr/Flowthrough_damr.cfg
lkotipal: Yep, around that. Letting the pulse propagate the entire box might not be that important for this test, my aim was just to have a test that both refines and coarsens cells dynamically, which it does in the shorter time.

testpackage/tests/Flowthrough_damr/Flowthrough_damr.cfg
lkotipal: So that doesn't work quite right; instead of a L2 refinement following the pulse we get a L1 region at 100s and no refinement after, so that loses out on testing induced refinement at least.

testpackage/test_definitions_small.sh
```diff
@@ -136,16 +136,24 @@ variable_components[${index}]="0 0 1 2"
 
 # 13 Large AMR translation flowthrough test
 test_name[${index}]="Flowthrough_amr"
-comparison_vlsv[${index}]="bulk.0000001.vlsv"
+comparison_vlsv[${index}]="bulk.0000001.vlsv bulk.0000002.vlsv bulk.0000003.vlsv bulk.0000004.vlsv"
```
markusbattarbee, line 139: We don't want the extra bulk files listed here, they should go on line 147.

testpackage/tests/Flowthrough_damr/Flowthrough_damr.cfg
```diff
@@ -0,0 +1,116 @@
...
+propagate_vlasov_translation = 1
+dynamic_timestep = 0
```
ykempf, line 7: Is that intended?

markusbattarbee: Turning off dyndt? yes.

ykempf: OK. Is it needed? We're possibly catching more potential bugs if it's dynamic, no? But fine by me if it's static.

PR #904, opened 2024-02-05
Error handling on failed writes

lkotipal:
- Failed bulk writes are fatal. If the disk/quota is full, failed writes essentially cause data loss and unreliable output. Due to writing cadence we can't expect someone to be always on hand to clear disk space.
- Try to clean up failed restart writes by removing the malformed file. Since restart writes are less common and much larger than bulk files, it's much easier to recover from a failed restart write. By removing the failed restart, the simulation can possibly run normally for at least some time, hopefully enough for someone to clear disk space and manually trigger a restart write.

### DIFF
--- a/vlasiator.cpp
+++ b/vlasiator.cpp
@@ -1029,6 +1029,10 @@ int main(int argn,char* args[]) {
                   version,
                   config,
                   outputReducer,"restart",(uint)P::t,P::restartStripeFactor) == false ) {
+            // If restart write fails, remove the malformed file and hope someone clears space soon
+            MPI_Barrier(MPI_COMM_WORLD);
+            std::remove(P::lastRestart.c_str());
+            P::lastRestart = "";
             logFile << "(IO): ERROR Failed to write restart!" << endl << writeVerbose;
             cerr << "FAILED TO WRITE RESTART" << endl;
          }
--- a/vlasiator.cpp
+++ b/vlasiator.cpp
@@ -590,7 +590,7 @@ int main(int argn,char* args[]) {
             writeGhosts
          ) == false
       ) {
-         cerr << "FAILED TO WRITE GRID AT " << __FILE__ << " " << __LINE__ << endl;
+         abort_mpi("FAILED TO WRITE GRID", 1);
       }
 
       phiprof::stop("Initialization");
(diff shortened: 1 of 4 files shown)


markusbattarbee (approved):

ykempf:
Next time, call it "retrigger CI" or something, or just "CI", instead of "dummy". :)

vlasiator.cpp
```diff
@@ -590,7 +590,7 @@ int main(int argn,char* args[]) {
             writeGhosts
          ) == false
       ) {
-         cerr << "FAILED TO WRITE GRID AT " << __FILE__ << " " << __LINE__ << endl;
+         abort_mpi("FAILED TO WRITE GRID", 1);
```
ykempf, line 593: This and later: I would greatly prefer to have the file name and line number passed down to abort_mpi and then error type 0.

vlasiator.cpp
```diff
@@ -1029,6 +1029,10 @@ int main(int argn,char* args[]) {
                   version,
                   config,
                   outputReducer,"restart",(uint)P::t,P::restartStripeFactor) == false ) {
+            // If restart write fails, remove the malformed file and hope someone clears space soon
+            MPI_Barrier(MPI_COMM_WORLD);
+            std::remove(P::lastRestart.c_str());
```
markusbattarbee, line 1034: only try this if the filename is not empty

markusbattarbee (approved):

ykempf:
So from our f2f2z discussion:
- just get the file name out from writeRestart instead of plonking it into P:: as we don't seem to be needing it there atm, and reduces the risk of somehow nuking also the previous file
- the MPI_Barrier is risky atm if not all tasks return a false.

lkotipal:
Addressed points from the discussion, but untested for now

ykempf:
Right so now I agree with the logic of the checks around the file removal. Let's see about testing as just discussed.

lkotipal:
So for the test status on this -- the success path is checked by CI. Failure path is a bit tougher, since Vlasiator is too smart to try to write to paths without access. I did try hardcoding to `/dev/full` in iowrite, which does fail as expected but doesn't really check all possible failure modes or the file deletion feature.

markusbattarbee:
Have you verified that the string containing the restart file name is correct, even in case of a successful write?

PR #918, opened 2024-02-13
fsgrid RAII

lkotipal:
- Split everything but MPI initialization and finalization away from `main` into a function called `simulate`. Since the lifetime of ~every object is now within MPI lifetime, every class can (and should!) handle freeing of MPI objects in their destructor automatically.
- Take advantage of https://github.com/fmihpc/fsgrid/pull/30 to handle fsgrid finalization via RAII.


ursg (approved):
Yep, I approve.

One thing I have been wondering about for a while, which this is also a step in the right direction for: Would it be possible to actually encapsulate all simulation state and code into one C++ object (basically by moving all state into a singleton like the object wrapper, and making it not-a-singleton anymore), so that Vlasiator could concievably be instantiated from some external language bindings. Like if you want to do a parameter study, you could use some external python script that creates a bunch of Vlasiator objects, runs them one after the other, and then does something smart with the results.

But anyway, that is quite a big step away still.
I'm in favour of splitting simulate() out there, since main() is too long anyway.

ykempf:
@kstppd can you comment on whether this new swap should be used in filtering too?

kstppd:
> @kstppd can you comment on whether this new swap should be used in filtering too?

Yup gets the job done!

alhom:
```
🔶 Flowthrough_amr: Nonzero diffs: `proton/vg_rho` has absolute error 7.15e+04, `fg_e` has relative error 0.446. Max timestamp difference is -0.859. Speedup: 1.01056
```

alhom:
This is now running on fresh CI, but the `flowthrough_amr` diff persists.

markusbattarbee (commented):
almost there

vlasiator.cpp
```diff
@@ -1372,27 +1325,73 @@ int main(int argn,char* args[]) {
```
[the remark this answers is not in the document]
lkotipal: Would be a bit awkward since we already have a "simulationTimer"

markusbattarbee (approved):

fieldsolver/gridGlue.cpp
```diff
@@ -94,31 +94,27 @@ void filterMoments(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
...
 
-               // Get pointers to our cells
-               std::array<Real, fsgrids::moments::N_MOMENTS> *cell;  
-               std::array<Real, fsgrids::moments::N_MOMENTS> *swap;
-            
                // Set Cell to zero before passing filter
-               swap = swapGrid.get(i,j,k);
-               for (int e = 0; e < fsgrids::moments::N_MOMENTS; ++e) {
-                  swap->at(e)=0.0;
+               for (auto& moment : *swap) {
```
markusbattarbee: I know you like to use these new auto element loops, but I really dislike it here. In my view, this obfuscates what is going on. Plus we have a similar loop over moments on line 115 which cannot be converted, so legibility would benefit from them having the same syntax.

lkotipal: I completely disagree about it being worse, sure that could be `real&` but the range-based loop clarifies what's actually going on rather than obfuscating it. I'll begrudgingly accept that it's bad that the two loops look different now though and revert that one, since we don't have C++23 and `zip` yet.

It also like I also managed to faff up the scope on `cell` which is thankfully not too important.

fieldsolver/gridGlue.cpp
```diff
@@ -113,8 +115,8 @@ void filterMoments(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
                for (int c=-kernelOffset; c<=kernelOffset; c++){
                   for (int b=-kernelOffset; b<=kernelOffset; b++){
                      for (int a=-kernelOffset; a<=kernelOffset; a++){
-                        auto* cell {momentsGrid.get(i+a,j+b,k+c)};
                         for (int e = 0; e < fsgrids::moments::N_MOMENTS; ++e) {
+                           const auto* cell {momentsGrid.get(i+a,j+b,k+c)};
```
markusbattarbee: I agree this can be const, but why put it inside the e loop?

lkotipal: Yep, moved that back up in scope

fieldsolver/gridGlue.cpp
ursg: My opinion:

Semantically, the contents of a fsgrid container really aren't "something to loop over". Even though they are stored as arrays of floating point numbers, they really are just structs in a different dress (mostly for historical reasons).
So while "go through all these numbers and set them to zero" is something that a range-based loop does correctly, it does not correspond to the semantics of what's going on here.

If you want to optimize it, you can use a memset call here. :)

lkotipal: I did actually consider `*swap = std::array<Real, N_MOMENTS> {0}`. Not sure if that would end up optimizing to `memset`, but it probably doesn't matter too much either way 😄.

PR #920, opened 2024-02-15
Refinement param renames

lkotipal:
- Use alpha_2 instead of J/B internally and in data reducers.
- Rename alpha parameters to alpha_1 and alpha_2 to match the literature.

### DIFF
--- a/common.h
+++ b/common.h
@@ -207,8 +207,8 @@ namespace CellParams {
       AMR_DPSQ,
       AMR_DBSQ,
       AMR_DB,
-      AMR_ALPHA,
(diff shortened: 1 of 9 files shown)


markusbattarbee:
these as well for completeness?
https://github.com/fmihpc/vlasiator/blob/8c79d0541ccc93e06d10a54ab422d426619aae63/parameters.cpp#L164
https://github.com/fmihpc/vlasiator/blob/8c79d0541ccc93e06d10a54ab422d426619aae63/parameters.h#L198

markusbattarbee (approved):

PR #921, opened 2024-02-20
Use Github "Retry Step" Action to wait for vlasiator binary on lustre.

ursg:
Instead of speculatively waiting for 10 seconds and then potentially failing the whole action, this uses the "retry step" action to run "ls vlasiator", and if that fails, wait 10 seconds to try again, a total of three times.
Maybe that helps with the lustre synchronization issues on vakka.


ursg:
Looks like [https://github.com/fmihpc/vlasiator/actions/runs/7974724582/job/21771295213?pr=921] this worked exactly as intended, failing to see the vlasiator on the first attempt, but getting it on the second.

markusbattarbee (approved):
Awesome!

PR #924, opened 2024-02-24
Vlasiator gpu bring up to date with dev

markusbattarbee:
Noteworth that the addition of no unsafe math optimizations to testpackage math flags causes diffs compared to previous version, but is of course preferable.


markusbattarbee:
Noted some more minor diffs due to the derivatives sign fix. Even after check-reverting that there were some microscopic changes in e.g. close-to-zero velocity components, but those were then well in the acceptable range.

PR #926, opened 2024-02-26
Add support for DCCRG/FSgrid split feature (requires support from DCCRG and FSgrid)

hokkanen:
A more detailed description of the feature is given in: 

https://github.com/fmihpc/fsgrid/pull/26

and 

https://github.com/fmihpc/dccrg/pull/30


iowrite.cpp
```diff
@@ -381,7 +381,7 @@ bool writeDataReducer(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
       }
    }
 
-   if( success ) {
+   if( dataReducer.getName(dataReducerIndex).find("fg_", 0) != 0 ) {
```
markusbattarbee, line 384: So this check is true when the datareducer name does *not* start with `fg_`. This means that it now tries to go into this section for both DCCRG ionospheric reducers.

So I suggest that:
 * here we check if it *does* start with `fg_`, and then do the the code on current line 429 onwards
 * if that fails, check if it *does* start with `ig_`, and in that case perform the code on line 463 onwards
 * if both those fail, fallback to DCCRG data, and perform the lines 387-425

hokkanen: Yes, this is the issue. I implemented this and do some testing now.

hokkanen:
Let's not merge yet, I noticed now some potential new issue when the dccrg/fsgrid split is activated. I'm looking into it.

markusbattarbee:
@ursg had an idea about how this could work nicely together with DLB https://pm.bsc.es/dlb if we have e.g. one FSgrid task per node, with all other tasks on the node dedicated to vlasov. The tasks could then share cores to one another during the waiting periods. How could one set the communicators here to respect such a request of distribution?

grid.cpp
```diff
@@ -503,9 +503,9 @@ void balanceLoad(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid, S
       //counter which is updated in acceleration, otherwise we just
       //use the number of blocks.
 //      if (P::propagateVlasovAcceleration) 
-      mpiGrid.set_cell_weight(cells[i], mpiGrid[cells[i]]->parameters[CellParams::LBWEIGHTCOUNTER]);
+//      mpiGrid.set_cell_weight(cells[i], mpiGrid[cells[i]]->parameters[CellParams::LBWEIGHTCOUNTER]);
 //      else
-//         mpiGrid.set_cell_weight(cells[i], mpiGrid[cells[i]]->get_number_of_all_velocity_blocks());
+      mpiGrid.set_cell_weight(cells[i], (Real) (1 + mpiGrid[cells[i]]->get_number_of_all_velocity_blocks()));
```
ursg, line 508: Has this slipped into the PR by accident, or is this actually intended to skip LBWEIGHTCOUNTER alltogether when load balancing?

ursg: (LBWEIGHTCOUNTER is set to exactly the num of velocity blocks just before. That +1 might have a small effect on distorting our load balancing, but I'm not to worried about it.)

markusbattarbee: (I'm fixing this along with some other interesting merge conflicts)

hokkanen: I don't think this is accident, I remember that the original setup was causing issues with the split cases. Unfortunately, the discussion appears to have disappeared from the rocket chat after 6 moths.

markusbattarbee: Well, we don't want to merge that as-it-is into Vlasiator dev, as that would wreck havoc without load balance. I suggest we do `(Real)1 + LBWEIGHTCOUNTER` for now, and add a comment about FS-split functionality here. How about that?

hokkanen: Sure. It should work without splitting, but there may be then some issues with some splits that may not function correctly.

hokkanen:
Something I thought about too. Would adding some tests with the split setup make sense? To make sure future developments do not break the split option.

PR #927, opened 2024-02-26
Reworked testpackage in a major fashion.

markusbattarbee:
This PR changes the testpackage:
* vlsvdiff is called only once per file instead of once for absolute and once for relative diffs
* Each test can diff several files (e.g. acceleration tests initial state, after half a gyration, after one gyration, after 10 gyrations
* Cleaned up the old transtest_2 which was pointless
* Added tests which run acc and fieldsolver after restarting
* Added AMR translation single-velocity-cell tests
* etc etc


markusbattarbee:
[dev_tp_vs_notp_log.txt](https://github.com/fmihpc/vlasiator/files/14406643/dev_tp_vs_notp_log.txt)

Sample log from a manual run: dev with SPF production flags compared against DPF testpackage flags

markusbattarbee:
Gonna still do some changes, e.g. probably run the AMR signal tests for a shorter time and edit which timesteps are being compared against reference. Also want to already change both the regular and the CI runners to print out file time difference if it isn't zero.

markusbattarbee:
Hmm, now the final CI report is truncated...

testpackage/small_test_definitions.sh
```diff
@@ -17,139 +17,230 @@ run_dir="run"
...
+
+# choose tests to run (default: all tests)
+run_tests=( )
+for (( c=1; c<$index; c++ ))
+do
+    run_tests+=($c)
+done
+
+# Alternatively, set tests manually, e.g.
+# run_tests=( 1 6 9 )
```
ursg, line 246: not too end-user friendly that these need to be added by numerical index, if these indices are now dynamically generated. So any addition of new tests will now shift the numbers, and requires manual re-numbering.

This could be done by selecting them by their ```test_name```, maybe?
Anyway, it's fine in the context of this PR. Just as soon as we *do* set up smaller tests somewhere (other architectures), this should be reworked.

ursg (approved):
Looks good.

I like the script cleanups, and the newly-introduced tests make sense.
(Didn't very thoroughly read through their configs, but I assume "you know what you are doing there").

testpackage/small_test_definitions.sh
markusbattarbee: In the past, they were also selected by index, and the index was added manually (and I remember sometimes having issues with tests mixing indexes). So even though that part of it hasn't been drastically improved, it's not actually worse either.

I added currently up-to-date numbers to the test descriptions for now.

testpackage/tests/Flowthrough_x_inflow_y_outflow_acc/Flowthrough_x_inflow_y_outflow_acc.cfg
```diff
@@ -38,9 +38,9 @@ y_min = -1.3e8
 y_max = 1.3e8
 z_min = -6.5e6
 z_max = 6.5e6
-
-t_max = 650
+t_max = 600
 dt = 2.0
```
ykempf, line 42: You probably don't want to do that but some of the explicit dt are obsolete since there is dynamic timestepping.

testpackage/tests/Ionosphere_small/Ionosphere_small.cfg
```diff
@@ -148,6 +148,7 @@ constBgBX = 0.0
 constBgBY = 0.0
 constBgBZ = -5.0e-9
 noDipoleInSW = 0 # Use 0 with dipole type 4
+dipoleScalingFactor = 0.1 # smaller dipole for faster testing
```
ykempf, line 151: Is that helping with the radius too then? We/I've procrastinated that other PR on this for a while.

testpackage/tests/Magnetosphere_3D_small/Magnetosphere_3D_small.cfg
```diff
@@ -115,6 +115,7 @@ constBgBX = 0.0
 constBgBY = 0.0
 constBgBZ = -5.0e-9
 noDipoleInSW = 0 # Use 0 with dipole type 4
+dipoleScalingFactor = 0.1 # smaller dipole for faster testing
```
ykempf, line 118: Don't tell Minna. :P

testpackage/tests/restart_read/restart_read.cfg
```diff
@@ -2,17 +2,18 @@ project = Flowthrough
 propagate_field = 0
 propagate_vlasov_acceleration = 0
 propagate_vlasov_translation = 1
-dynamic_timestep = 1
+dynamic_timestep = 0
```
ykempf, line 5: Why?

ykempf (commented):
OK looks fine, we'll see soon enough whether it also works as intended in the case of non-missing reference data. The biggest question I have is why we had/have a restart test with only one solver component (we must have had a good reason?!) and now you added one "misleadingly" called *acc which has all trans+acc+fs on.

ursg (approved):
Yup, lookie good to me. Much cleaner than before.

testpackage/tests/restart_read/restart_read.cfg
markusbattarbee: The new _acc restart tests handle dynamic timestepping.

ykempf: Do we really need both? Can't we just run one restart test? Also, but this goes beyond this PR, note for further development, we should test restart including e.g. ionosphere.

testpackage/tests/Magnetosphere_3D_small/Magnetosphere_3D_small.cfg
markusbattarbee: Now runs to 6 seconds instead of 3 seconds
```
Magnetosphere_3D_small
95.35        40.63         2.34679
```
up for discussion of course.

testpackage/tests/Ionosphere_small/Ionosphere_small.cfg
markusbattarbee: I didn't change radius, but thought that to get any valid ionosphere output we need a longer run. Now runs to 30 seconds instead of 3 seconds.
```
Ionosphere_small 
125.2        271.9         0.460463

testpackage/tests/Magnetosphere_3D_small/Magnetosphere_3D_small.cfg
ykempf: I think it's OK. This is not a realistic m'sphere setup anyway. If/when we develop a larger-scale testpackage I suppose we should stick closer to production-like parameters.

testpackage/tests/restart_read/restart_read.cfg
markusbattarbee: I agree that an ionosphere restart would probably be a good idea. 
Restart tests cost like half a second to run, and it's a lot easier to debug differences in the restart state without having ACC+FS active. At the same time, I think we need a test which verifies that they run in a sane manner after a restart.
I guess, if we wanted, we could do just one write and restart from it with different settings, but that might get confusing.

testpackage/tests/Ionosphere_small/Ionosphere_small.cfg
ykempf: OK then this guy's downmap radius (6.5 RE) is still within the inner radius of 70000 km. Weakening the dipole should allow to pull in the radius enough, or conversely push out the downmap radius and still map. Can be handled as part of the separate ionosphere radius tp PR.

testpackage/tests/restart_read/restart_read.cfg
ykempf: Production-relevant restarting could be tested off the final state of the Ionosphere* test for example. But then maybe again not in this PR? :)

testpackage/tests/Ionosphere_small/Ionosphere_small.cfg
markusbattarbee: In https://github.com/fmihpc/vlasiator/pull/797 I wrote:
> I've made a test run at `/wrk-vakka/users/markusb/KTH/Ionosphere_intermediate` with 30 seconds of simulation, FS subcycling, the inner boundary at 25e6 metres (just under 4 RE), and a dipole scaling factor of 0.001. This ran in 756 seconds on Carrington so is still relatively expensive for CI. I'm also a bit surprised that FS subcycles were consistently at the maximum allowed value. Up for discussion and adjustment.

An issue with pulling the dipole magnitude down a lot (like to 0.1%) is that then the shock does not really form as the obstacle is too weak. And despite that, pushing the inner boundary inwards still took a lot of time to simulate. 

This is the state at 15s:
image

But yea, getting ionospheric outputs which can actually be evaluated is too much for this PR.

testpackage/tests/restart_read/restart_read.cfg
markusbattarbee: Again, it's a question of how it would be good to have tests which allow answering not only the question "does this run?" but also  "*why* does this not run?". If there's a bug, it's a lot easier to debug from a 2D simulation IMO. But I agree that we could add an Ionosphere_restart test, I just think this has value as well.

testpackage/tests/Magnetosphere_3D_small/Magnetosphere_3D_small.cfg
ursg: Yep, I'm ok with it too. The main reason we run the testpackage is to get all solvers to play with some semi-realistic numbers, and to check whether we broke anything.
Proper physical validation of the code should be a bit of an orthogonal concept, which we might also want to formalize in some way at some point.

But not in this PR. :)

testpackage/tests/restart_read/restart_read.cfg
ykempf: Yep, in the interest of wrapping this PR up I'll resolve this discussion too. It is production-critical that we can restart with all the bells and whistles, and previous experience has shown that we can forget some ionosphere solver variables, or connectivity data, or whatever. But again, not here, resolving this now. Pinpointing individual solvers in this test is useful too.

testpackage/tests/Flowthrough_x_inflow_y_outflow_acc/Flowthrough_x_inflow_y_outflow_acc.cfg
ykempf: "next time", resolving

markusbattarbee: done.

testpackage/small_test_definitions.sh
ykempf: One day, this will be eg a python script and we can use dictionaries or so?

ykempf:
Flowthrough_amr (l. 373 of the testpackage running output) reports a
WARNING! TIMESTAMPS DO NOT MATCH! dt=-7.04858e-12
which will raise eyebrows. Are we living with the diff levels that need "experienced" reading (like a bunch of relatively big but absolutely insignificant variable diffs reported), or do we brush time differences under a carpet somewhere around ns level?
If we keep it like this, let's merge.
If you do one last push, I'd prefer these lines not report `dt` as my brain sees this as the time step, but I'll tell my brain to learn this new trick. ;-)

ykempf:
- OK let's keep it in as a reminder to sensibilise Flowthrough_amr at some point. :)
- `delta-t `, `t2-t1` of `time difference`, as you like. $$\Delta$$ will break things I suppose. :sweat_smile:

PR #930, opened 2024-03-04
Makefile and github actions steps to build on apple M1

ursg:
In absence of an actual test system, I will use the CI runners themselves to try to get this to work, which means I will probably spam a bunch of commits here.
(Maybe smash them all together into one once it runs.)
Please complain if things get too spammy. :)


ursg:
Ah well, [the absence of mpic++ on that runner](https://github.com/fmihpc/vlasiator/actions/runs/8138017342/job/22237812841?pr=930#step:3:9) is already not too encouraging.

ursg:
Alright, that first issue was worked around by [installing MPI on the fly as part of the job step](https://github.com/mpi4py/setup-mpi).
But next joy: [the macOS default LLVM compiler does not support openmp](https://github.com/fmihpc/vlasiator/actions/runs/8138059369/job/22237943159#step:4:10).

ursg:
Well, we have green check marks across the board, so let's merge it and see how quickly it breaks in the future!

PR #931, opened 2024-03-06
Fix CI runs on RISCV failing, probably due to old libraries still lingering in the work directory.

ursg:
First attempt: let's clean the CI workspace before actually building anything.

### DIFF
--- a/build_libraries.sh
+++ b/build_libraries.sh
@@ -22,7 +22,13 @@ cd library-build
 # Build phiprof
 git clone https://github.com/fmihpc/phiprof/ 
 cd phiprof/src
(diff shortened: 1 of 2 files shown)


build_libraries.sh
```diff
@@ -22,7 +22,13 @@ cd library-build
 # Build phiprof
 git clone https://github.com/fmihpc/phiprof/ 
 cd phiprof/src
-make -j 4 CCC=mpic++
+
+if [[ $PLATFORM != "-arriesgado" ]]; then
```
markusbattarbee, line 26: errant `-` here

PR #934, opened 2024-03-08
Use mpirun instead of srun for tool runs on carrington for CI

ursg:
Diagnosing something entirely unrelated, I found that testpackage runs spend plenty of time on the carrington nodes doing nothing at 100% cpu idle.
Apparently, srun wants to wait for some timeout to place the jobs locally. Maybe this is related to the recent slurm restructuring?

Anyway, mpirun seems to do the job.


ursg:
Yep, 9 min faster. Let's take it.

PR #939, opened 2024-04-04
Add IPShock option for performing coordinate transformation or not

henry2004y:
The current IPShock project always perform a coordinate transformation into the deHoffman-Teller frame for the variables in the physical cells, but not the boundary cells. This PR adds an option to let users choose whether the coordinate transformation is performed or not.

Side notes:

1. There is a hidden assumption that the shock always locates at x = 0.

### DIFF
--- a/projects/IPShock/IPShock.cfg
+++ b/projects/IPShock/IPShock.cfg
@@ -134,6 +134,7 @@ BX0d =   3.93058e-09
 BY0d =       0.00000
 BZ0d =    1.1553350e-08
 Width = 1.0e6
+deHoffmanTeller = 0
 
 AMR_L1width = 3.0e6
 AMR_L2width = 1.5e6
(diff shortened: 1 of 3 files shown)


ykempf:
Because we (I at least) usually set our parameters in the dHT frame with an external tool so the in-code transformation probably just did nothing.

ykempf:
I would suggest that changing the frame would not drop the interpolation, it is useful to be able to have a smoother transition than a direct step.

henry2004y:
Shock width support has been added for non-dHT frame.

markusbattarbee:
This PR has the fundamental flaw, that boundary condition .dat files are cannot really be edited inside Vlasiator, and they have to match the conditions set up by the project - thus, we cannot change the conditions inside Vlasiator itself.

The only reason there are these calculations inside the project is to get fancy interpolation between upstream and downstream conditions retaining physical correctness throughout the transition.

PR #941, opened 2024-04-09
Cross-compile on BSC RiscV cluster for much faster CI builds

ursg:
Instead of building on the (slow) riscV-nodes itself, it uses the "synth-hca" server in the cluster (which is nominally used to build the FPGA payloads for the hardware development platforms, but is mostly just idle), to cross-compile for the nodes.

The result has been verified to run just fine on the arriesgado cluster.

This accelerates riscv library + code builds from 26 minutes to about 2.5 minutes.


ursg:
And indeed:
riscv

PR #946, opened 2024-04-22
Update iowrite.cpp

ArnaudLalague:
Write parameters into the VLSL file


iowrite.cpp
```diff
@@ -1075,6 +1075,12 @@ bool writeIonosphereGridMetadata(vlsv::Writer& vlsvWriter) {
   } else {
     vlsvWriter.writeArray("MESH", xmlAttributes, 0, 1, ionosphereGridElementsAndCorners.data());
   }
+  
+  // Write different parameters of the class Ionosphere into the VLSL file
```
ykempf, line 1079: And that should read VLSV, not VLSL. :)

ykempf, line 1079: Oh and this set of writing calls could be put behind an if check that we actually are running an ionosphere, let me dig up an example for you.

ykempf:
See #949

PR #948, opened 2024-04-23
Amr translation cleanup addon

alhom:
A bit of restructuring the pencil creation - dimension-looping out of grid.cpp, old pencil removal to be done only once.

Not sure if we actually need those named criticals - while good to have, those may be a bit risky..


ykempf:
Uh you want them critical regions to have *different* names! Or you might have one waiting for the other, or sth similar.

alhom:
> Uh you want them critical regions to have _different_ names! Or you might have one waiting for the other, or sth similar.

Those regions are handling the same vector, so we want those push_backs still block each others, hence the same name. Keeping track of that is ofc much easier when you don't have named criticals and everything is blocking everything.

markusbattarbee (approved):

ursg (approved):
I like!

PR #951, opened 2024-04-23
Update carrington CI jobscript to use srun instead of mpirun.

ursg:
After discussion with IT, this seems to be the preferred way to launch jobs on carrington right now to get actual parallel openmp performance.

Let's see if this makes CI testpackage runs faster.


ursg:
But alas, it looks like it isn't faster at all.
Even though sideloading to the slurm job compute node shows me that it is now 100% of the CPU power (compared to 25% before).
I suppose this is now the same problem that was fixed by #934 switching to mpirun in the first place?

Anyway, let's place this one on hold until I have talked to IT about it.

markusbattarbee:
Already incorporated in #1074

PR #953, opened 2024-04-23
Dccrg RAII

lkotipal:
Working on #918, reduce scope of mpiGrid so it can clean its communicator up.


lkotipal:
> Flowthrough_damr: Nonzero diffs: `proton/vg_rho` has absolute error 1.94e+04, `fg_e` has relative error 0.0547. Max timestamp difference is 0.. Speedup: 1.00253

Ah... But no such error in `Flowthrough_amr`, can this be a fluke?

PR #955, opened 2024-04-26
Move spatial cell to cpu-specific file

markusbattarbee:
This introduces a spatial_cell wrapper which should aid with making sense of the future GPU branch PR.


ursg:
This is only boilerplate change so far, right? Files were renamed and includes adapted and all types still have the same name, as far as I can see.

What's the next step then? Will mpiGrid continue to carry "spatialCell" as its contents, meaning the CPU-side version? Or will that have some typename change (and thus major refactoring everywhere)?

markusbattarbee:
This is indeed boilerplate change, so that when the GPU pull request comes in, it'll be easier to see and evaluate the changes. This should help with the fact that Github doesn't deal well with showing changes when there's simultaneously
 * renaming a file (spatial_cell.hpp -> spatial_cell_cpu.hpp)
 * changes to contents (the above file)
 * a new file with a same name as the old one (wrapper spatial_cell.hpp - thus making it explicitly spatial_cell_wrapper.hpp)

ursg (approved):
Alright then, looks good to me!

alhom (approved):

PR #956, opened 2024-04-26
Gpu port first version into dev

markusbattarbee:
This is a draft PR. May well still end up being closed, but showcases the current GPU branch changes compared to dev.


datareduction/datareductionoperator.cpp
```diff
@@ -672,50 +670,48 @@ namespace DRO {
...
          // these updates need to be atomic:
-         # pragma omp critical
+# pragma omp critical
          {
-            PTensor[0] += thread_nvyvz_sum;
-            PTensor[1] += thread_nvzvx_sum;
-            PTensor[2] += thread_nvxvy_sum;
+            PTensor[0] += sum[2];
+            PTensor[1] += sum[1];
+            PTensor[2] += sum[0];
```
ykempf, line 714: :eyes: :interrobang:

ykempf: Ah no ok they switched order in the loop too.

grid.cpp
```diff
@@ -553,14 +537,15 @@ void balanceLoad(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid, S
 
          int prepareReceives {phiprof::initializeTimer("Preparing receives")};
          int receives = 0;
+         #pragma omp parallel for
```
ykempf, line 540: Should that take some schedule too, just to be safer?

grid.cpp
```diff
@@ -633,17 +618,48 @@ void balanceLoad(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid, S
...
+         SC->dev_upload_population(popID);
+      }
    }
+   // Call GPU routines for per-thread memory allocation for Vlasov solvers
+   // deallocates first if necessary
+   //GPUTODO: Also count how many pencils exist
+   gpu_vlasov_allocate(gpuMaxBlockCount);
+   gpu_acc_allocate(gpuMaxBlockCount);
+   gpuMallocTimer.stop();
+#endif
```
ykempf, line 662: Mark `USE_GPU` here at the end too just to make it easier to look for it?

iowrite.cpp
```diff
@@ -173,20 +174,21 @@ bool writeVelocityDistributionData(const uint popID,Writer& vlsvWriter,
...
    attribs.clear();
    attribs["mesh"] = getObjectWrapper().particleSpecies[popID].name;
    attribs["type"] = vlsv::mesh::STRING_UCD_AMR;
 
-   // stringstream is necessary here to correctly convert refLevelMaxAllowed into a string 
+   // stringstream is necessary here to correctly convert refLevelMaxAllowed (hardcoded to zero now) into a string 
    stringstream ss;
-   ss << static_cast<unsigned int>(getObjectWrapper().velocityMeshes[meshID].refLevelMaxAllowed);
+   //ss << static_cast<unsigned int>(vmesh::getMeshWrapper()->velocityMeshes->at(meshID).refLevelMaxAllowed);
+   ss << static_cast<unsigned int>(0);
```
ykempf, line 191: That zero will be removed in due time, correct?

markusbattarbee: This zero is what is used to store in the VLSV file that this file does not contain velocity mesh refinement. I think it's probably safer to keep storing the value in the file, like before. I can clean up the comment lines, though.

PR #957, opened 2024-05-02
Add LFM (Merkin et al 2010) ionosphere solver test.

ursg:
This adds a small test from [the Merkin et al (2010) paper](https://dx.doi.org/10.1029/2010ja015461), where eq 13 specifies a ionosphere FAC test case with an analytic solution.

Also gives the miniApp's makefile a much more flexible multipole test up to arbitrary orders.

This is, of course, then being used to produce plots for the ionosphere paper.

### DIFF
--- a/.github/workflows/github-ci.yml
+++ b/.github/workflows/github-ci.yml
@@ -313,8 +313,9 @@ jobs:
 
   build_ionosphereTests:
     # Build IonosphereSolverTests miniApp
-    runs-on: carrington
-    needs: [build_libraries]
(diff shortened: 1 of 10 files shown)


ursg:
This now automatically runs the ionosphere tests as a CI stage (on Github cloud runners) and [produces an "ionosphereTestResults" artefact](https://github.com/fmihpc/vlasiator/actions/runs/9275969419) that contains plot output pictures. Now if only those could be somehow displayed as a check result, like the testpackage results...

markusbattarbee (approved):

PR #958, opened 2024-05-03
Gpu update - new block adjustment algorithm and plenty of other improvements

markusbattarbee:


markusbattarbee:
Also brings in Hashinator as a submodule (though a commit which is in fact a merge of two of my development branches)

PR #960, opened 2024-05-06
Updates to the Kelvin-Helmholtz instability project

vetarvus:
Fixes and updates to the existing 'KHB' Vlasiator project. 

The updated version has a pressure balance across the simulation domain and options for a multichromatic initial perturbation with random phases.

Also added an example config file to samples/, which can be used as a quick test/demonstration.


ursg (approved):
Looks all good. Also passes the testpackage all-green, so thumbs up from me!

samples/KHB/KHB.cfg
```diff
@@ -0,0 +1,153 @@
...
+file_x- = mxm.dat
+file_x+ = mxp.dat
+
+# Variables (data reduction operators) to be written to the output VLSV files and to the diagnostic.txt file
+# fg_* variables on the finest-resolution field solver grid (/!\ can be several GB per component in large runs)
+# vg_* variables on the adaptive mesh storing the plasma distributions and moments
+# populations_* variables are stored separately for each specified ion population as <population>/vg_*
+[variables]
+output = populations_vg_rho
+output = populations_vg_v
```
ykempf, line 110: I suggest that you add a few more parameters relevant for plotting/analysing KH as an example here.

markusbattarbee:
I was now testing this, and the memory footprint is massive. The cause is, I think, that this project doesn't inherit the `TriAxisSearch` class, thus defaulting to the base project v-space initialization which in fact initializes all blocks (and because `shrink_to_fit` is bypassed, the footprint remains huge).

The solution is to make this inherit `TriAxisSearch`, but that requires that it implements a `getV0()` function which impacts initialization in general.

markusbattarbee:
Implemented in https://github.com/vetarvus/vlasiator/pull/1

ursg (approved):

PR #961, opened 2024-05-07
Concurrency check and trap to scancel testpackage_run if caught cancel signal

alhom:
Two modifications:
1) Concurrency group definition: One github-ci workflow per PR / branch/tag reference, see https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre/72408109#72408109

2) Parse testpackage sbatch output and create a trap to cancel the job if Ctrl-C signal encountered. This is a bit hacky with a srun --dependency call to bypass the sbatch -W functionality.


ursg:
Yay, looks like this works as intended

Issue #959, opened 2024-05-06
generate_version.sh sources /etc/profile possibly? changes the environment


astatide:
Hola hola!

Not sure if it's an issue on more than just MareNostrum 5, but there might be a line that could catch users unawares in the generate_version.sh: https://github.com/fmihpc/vlasiator/blob/583b8cede3d8a04f9d3dfc61a7574e9b8e7deb37/generate_version.sh#L22

this could cause issues on machines where a default set of modules are loaded via the profile, particularly for compilers.  It might be worth mentioning this in the isntall documentation (unless I missed it) as a user may well already have their modules set and depending on the cluster/computer, this might either generate errors or override the loaded compilers!

(for instance, as the default on MN5 is Intel, using gcc is fine because it just refuses to load the intel modules; however, if I'm using the intel compilers but not the 'default' selection, it'll end up running at least that script with the wrong version.

What isn't clear to me is whether these potential module changes persist beyond the running of generate_version.sh, or whether that occurs in an isolated environment and once that make task is run it goes back to the parent environment.  So, sorry if it's a false issue : )

ursg:
Ah, interesting find. We have never run into problems of this kind ourselves, but I can see how it could confuse the build process. (In most cases, we just ignore errors caused by generate_version.sh, which is maybe not the best style).

Anyway, looking at the git blame output for this file, it seems it has been there "forever", even before I joined the Vlasiator project, and the reason for its inclusion has been lost to the ages. So I suggest we just drop it, see if CI builds still work fine, and continue.

ykempf:
Agreed, this is a (pre)historic remnant, do post a PR removing this line.

astatide:
Will do! : D

ursg:
Fixed by #963

PR #963, opened 2024-05-10
Remove . /etc/profile from generate_version.sh

astatide:
https://github.com/fmihpc/vlasiator/issues/959

As discussed in the above linked issue, this PR removes the sourcing of /etc/profile, as it's likely no longer needed and can potentially cause some minor issues on clusters that have a default module set (such as MN5)


ursg:
Looks good, CI is green. I shall now push this button here.

PR #964, opened 2024-05-10
Add Makefile for MareNostrum 5's general partition

astatide:
Hola hola!

This PR adds a Makefile for compiling Vlasiator on MareNostrum 5's general partition (that is, the partition without NVIDIA accelerators) : )


astatide:
ooops; messed up my git commit history.  Gimme a minute :sweat_smile:

astatide:
Okay, fixed the messed up history that accidentally included the other PR :sweat_smile: ah, git.

astatide:
I should add that this has _not_ been tested for numerical correctness of results, but only that it does indeed compile so long as you've properly compiled all the necessary libs.

I also have a version that uses the Intel compilers instead of GCC, which are technically the default on MN5.

ykempf:
Feel free to also commit the Intel compiler version. Historically on most machines we've used (mostly Cray however) GNU compilers were just a bit faster, but we've been able to run with Intel (or recently AOCC) all right as well, we have nothing against non-GNU compilers.

astatide:
Ah, well, to be honest, it's only a subtle change from the Intel compiler to the GCC one; I just change a compiler flag and the compiler variables.  Perhaps it would be more straightforward if I simply added a few commented lines of code to swap between them, rather than a whole separate file?

ykempf:
Sure, whatever works best for you as the main user of this Makefile at the moment. :) If/when we start to have "too many" Makefiles in this folder we'll come up with a way of sorting them better...

astatide:
Went ahead and added some comments to make it 'straightforward' to compile with Intel instead of GCC : )  also fixed a variable, too.

ursg (approved):

PR #966, opened 2024-05-14
Parallel timer fixes

lkotipal:
Pre-initialize timer cell-semilag-acc, and remove other timers initialize in parallel regions


ykempf (approved):
What is there not to like? Let's just wait for the CI rerun I just triggered.

PR #974, opened 2024-05-28
Ci carrington fs ops to srun

ykempf:


markusbattarbee (commented):
Sadly this did not solve the issue of the missing ionosphere and mag3d reports.

Also, I had a look that it is likely not caused by the time difference check, as although that's a relatively new thing, it actually crops up in several runs.

markusbattarbee:
Hmm. I think perhaps this character curb was a bit extreme, lessening legibility. We could instead indeed drop some output files (some tests have many bulk files listed), and remove e.g. `Comparing file /wrk-vakka/users/uganse/github-runner/work/vlasiator/vlasiator/testpackage/run_2024.05.28_17.59.43/acctest_1_maxw_500k_100k_20kms_10deg/fullf.0000020.vlsv against reference
` from the output, as `vlsvdiff`prints the files anyway.

markusbattarbee:
> and remove e.g. `Comparing file /wrk-vakka/users/uganse/github-runner/work/vlasiator/vlasiator/testpackage/run_2024.05.28_17.59.43/acctest_1_maxw_500k_100k_20kms_10deg/fullf.0000020.vlsv against reference ` from the output, as `vlsvdiff`prints the files anyway.

Realized this won't do, as the full file names are printed only when diffing VDFs, not bulk variables.
Perhaps we can then drop that latter printout from vlsvdiff itself?

ykempf:
This test's `testpackage_run_output.txt` is now 51K. Isn't that good enough for now?
What might be easier to not repeat all the time is the ID `d39c71de08d35a5f82dfde1756170f090fe25818___DACC_SEMILAG_PQM__DTRANS_SEMILAG_PPM__DDP__DDPF__DVEC4D_AGNER`?

ykempf:
OK I wasn't looking at the correct file for that size, but anyway, with these latest changes we should be pretty OK.

ykempf:
I'm trying to drop the "We detected OpenMPI so we set the cvars value to disable ompio, MCA io: ^ompio" line in an elegant way.

ykempf:
Good, so this version's `testpackage_check_description.txt` is now 50K, giving us some space to add a test or three before it annoys us again. @ursg has the next step up his sleeve: have several such reports come in.

markusbattarbee:
Yep, looking good, thank you!
The perfectionist would like lines 230, 253, 276 to somehow align all outputs, but that would be pointless work and would actually likely inflate the file size in the end (because e.g. a test with `proton/vg_ptensor_nonthermal_diagonal_0` would need all other lines also to be really long)

markusbattarbee (approved):

ykempf:
Now I'm done. How u like it?

testpackage/small_test_carrington_github_ci.sh
```diff
@@ -175,9 +186,9 @@ for run in ${run_tests[*]}; do
    speedup=$( echo $refPerf $newPerf |gawk '{if($2 == $2 + 0 && $1 == $1 + 0 ) print $1/$2; else print "NA"}')
    { {
    echo  "$refPerf        $newPerf         $speedup"
-   echo "------------------------------------------------------------"
-   echo "  variable     |     absolute diff     |     relative diff | "
-   echo "------------------------------------------------------------"
+   echo "-------------------------------------------"
+   echo " variable | absolute diff | relative diff |"
```
markusbattarbee, line 190: Oh, that tabulation fix is absolutely great! 
However, I think we also want it on this line so that the absolute and relative diff column headers "align".

Leaving it up to you to decide on how many dashes to include - one option would be to just cover "variable" in the top line, and all the way on the bottom one?

.github/workflows/github-ci.yml
```diff
@@ -292,16 +292,28 @@ jobs:
...
+        cd $GITHUB_WORKSPACE ;
+        tar -czf testpackage-output.tar.gz testpackage_check_description.txt testpackage_output_variables.txt ;
+        MORO
+        )
+        srun -M carrington --dependency afterany:${VLASIATOR_CIRUNNER_JOBSPEC_TP[0]} bash -c "$RUN_STRING"
         if [ -f $GITHUB_WORKSPACE/testpackage_failed ]; then
           # Fail this step if any test failed.
           exit 1
         fi
+    - name: Make sure the output binary is visible in lustre
```
markusbattarbee, line 310: Suggestion: make the name correct. :) This is a tarball of the log file, not a binary file per se.
Perhaps the job could also print out the size of the log file, so we now when it's approaching 64kb?

ykempf: that's copy-pasted from above! :P

markusbattarbee (approved):

PR #976, opened 2024-05-28
Write out vg_f_saved if any VDFs are coming as well.

ykempf:


markusbattarbee:
So this approach will write f_saved to all bulk files, even e.g. FSgrid-tagged bulk files without VDFs, as long as *some* bulk files in that run will include VDFs?

ykempf:
Yes, we always write all vg ones in all file types. I'm not rewriting *that* today.

PR #982, opened 2024-06-04
Nudge once at ionosphere update, only add the drift.

ykempf:
Restart incorrect for now.

Putting this up for discussion.


ykempf:
So I'll need to add storing the old drift velocity into restarts and reading from there. And I think I should fix that suspected missing neighbour update I pointed out on Zulip the other day.

PR #985, opened 2024-06-07
Refinement exclusion halfspaces

ykempf:
Adds a new set of parameters to disable refinement left or right of values in each directoin.

The parameters and variables are not termed very poetically, feel free to suggest more betterish jargon which won't make you forget what they mean the second you see them.

Also, I'm not yet 100% certain my boolean logic for unrefinement is good, though I *think* it works. Fresh :eyes: welcome.

Tested with a Flowthrough_amr test, would need better benchmarks at some point but not this week any more.

I will *not* expand this to become e.g. boxes of a given max allowed refinement, that would become a nightmare in our cfg setup.


lkotipal:
Should definitely be useful. I'd apply this to static refinement as well to get rid of tail refinement at box boundaries (and also set min x to a couple cells less than the simulation minimum x)

markusbattarbee (approved):

PR #986, opened 2024-06-12
Harris update, after #967

alhom:
Dangling update to make the Harris config 5.3-supported as done during the Plasma-PEPSC workshop. Was built on the slightly updated master branch used at the workshop, so this contains also PR#967.


alhom:
Rebased on dev after PR 976 to clean up the mess.

PR #988, opened 2024-06-13
Test running github CI on carrington using srun --interactive

ursg:
That way, we should have immediate stdout output, and action run kills should result in the slurm job being automatically reapt, too.


ursg:
Not pursuring this further, as #989 probably superseeded it.

PR #990, opened 2024-06-13
Add a "check cfg files" CI step.

ursg:
This is to ensure that we don't have stale cfg files lying around somewhere in our project directories.


.github/workflows/github-ci.yml
```diff
@@ -406,3 +409,34 @@ jobs:
...
+      uses: actions/download-artifact@v4
+      with:
+        name: vlasiator-release
+    - name: Make release binary executable
+      run: |
+        chmod +x $GITHUB_WORKSPACE/vlasiator
+    - name: Check cfg files in projects/
+      run: |
+        retval=0
+        for cfg in `find projects -name \*.cfg`; do
```
ykempf, line 432: Include samples and testpackage while at it?

ursg: Done

PR #998, opened 2024-06-18
Document refinement region limits behaviour

ykempf:


markusbattarbee:
Dunno if "induced neighbours" is actually the correct terminology - it's induced refinement within a neighbourhood. So perhaps just "induced refinement"?

markusbattarbee (approved):
This makes British English is sad :-(

ykempf:
> This makes British English is sad :-(

I know, but historically ("it's faster to type", Honkonen, private communication, 2011) we've tried to stick to US spelling for code at least. And I wouldn't like both US and UK spelling applied to variables in various regions of the code or it'll bite us nastily if someone uses `updatedNeighbourlist` after having run a loop doing, say, `updatedNeighborList.push_back(id)`.

PR #1000, opened 2024-06-19
Update DCCRG neighbor definitions, decrease neighborhood size by 1.

ursg:
After fixing a small number of bugs that were left, I suppose this is now ready for testing:

This PR updates the DCCRG submodel to point at my branch, in which the definition of what neighbor is has been changed. As a result, we no longer need any extra speculative neighborhood sizes to make sure we capture all relevant neighbors at AMR interfaces, and can run with a neighborhood_size that is one unit smaller.

This results in smaller induced refinement regions, less communication overhead and thus, hopefully, an overall speedup.

After running a number of testpackage tests by hand, I'm now keen to see if this runs the whole package, or whether there are still issues to iron out.

Note: This PR also updates all makefiles to build with c++20. Alternatively, the erase_if implementation in dccrg needs to be replaced with a backwards-compatible erase(remove) idiom.


vlasovsolver/cpu_trans_map.cpp
```diff
@@ -699,12 +703,22 @@ void update_remote_mapping_contribution(
             //2) is remote cell, 3) if the source cell in center was
             //translated
             ccell->neighbor_block_data[0] = pcell->get_data(popID);
+
+            for(unsigned int cell = 0; cell<VELOCITY_BLOCK_LENGTH * pcell->get_number_of_velocity_blocks(popID); ++cell) {
```
markusbattarbee, line 707: Put NaN checks inside debug ifdefs?

vlasovsolver/cpu_trans_map.cpp
```diff
@@ -734,14 +748,13 @@ void update_remote_mapping_contribution(
       break;
    }
    
-#pragma omp parallel
```
markusbattarbee, line 737: was this removed by accident?

vlasovsolver/cpu_trans_map.cpp
```diff
@@ -674,23 +657,44 @@ void update_remote_mapping_contribution(
       CellID p_ngbr = INVALID_CELLID;
       CellID m_ngbr = INVALID_CELLID;
 
-      for (const auto& [neighbor, dir] : mpiGrid.get_face_neighbors_of(local_cells[c])) {
+      //if(local_cells[c] == 4 || local_cells[c] == 5) {
```
markusbattarbee, line 660: this comment looks like it should be removed

spatial_cell_cpu.cpp
```diff
@@ -642,6 +652,13 @@ namespace spatial_cell {
             if ( P::amrMaxSpatialRefLevel == 0 || receiving || ranks.find(receiver_rank) != ranks.end()) {
                
                for ( int i = 0; i < MAX_NEIGHBORS_PER_DIM; ++i) {
+                  if(!receiving) {
+                     for(unsigned int j=0; j < this->neighbor_number_of_blocks[i]*VELOCITY_BLOCK_LENGTH; j++) {
```
markusbattarbee, line 656: This should probably be deleted or go behind a debug ifdef

markusbattarbee (changes requested):

ursg:
Hmm, these testpackage outputs are a bit puzzling. Flowthrough_amr diffs make sense with smaller induced refinement, but those magnetospheres shouldn't probably be that different.
And what's with those Waitall errors? It looks like output still got written? Are they happening at the end of runs?

ykempf:
> Hmm, these testpackage outputs are a bit puzzling. Flowthrough_amr diffs make sense with smaller induced refinement, but those magnetospheres shouldn't probably be that different. And what's with those Waitall errors? It looks like output still got written? Are they happening at the end of runs?

See my comments on Zulip...

grid.cpp
```diff
@@ -994,24 +976,6 @@ void initializeStencils(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpi
    // In spatial AMR using DCCRG, the neighbors are considered relative to a given cell's size.
    // To get two coarse neighbors from a fine cell at interfaces, the stencil size needs to be increased by one.
    int addStencilDepth = 0;
```
markusbattarbee, line 978: addStencilDepth should be purged throughout

.gitmodules
```diff
@@ -4,7 +4,7 @@
 
 [submodule "submodules/dccrg"]
 	path = submodules/dccrg
-	url = https://github.com/fmihpc/dccrg.git
+	url = https://github.com/ursg/dccrg.git
```
alhom, line 7: Shouldn't the relevant commit be visible via the PR in fmihpc - no need to point to ursg/dccrg?

vlasovsolver/cpu_trans_map.cpp
ursg: done.

vlasovsolver/cpu_trans_map.cpp
ursg: Indeed. Removed.

grid.cpp
ursg: Done.

.gitmodules
ursg: Looks like that works indeed. Nice!

ursg:
But actually, our fancy erase_if mechanism from DCCRG turns out not to even compile in the build_testpackage CI step!

markusbattarbee:
Included in #1060

PR #1003, opened 2024-07-02
AMR Filtering Update

lkotipal:
Draft for an improved filtering scheme, to be brought in line with fsgrid filtering in order to avoid "banding" from refined cells dropping filtering passes on fsgrid.
TODO:
- Testing :)
- Implementation of triangle filter to minimize communication
- `P::filterPasses` is a temporary parameter for testing, ideally this should just be based on fsgrid filter passes


lkotipal:
Back in the oven since it relies on a dccrg update that has been invalidated

PR #609, opened 2022-02-14
Amr translation optimizations

markusbattarbee:
Here are some slightly controversial optimizations to AMR translation. Each one is a config setting, default to off. They can be activated with:
```
[vlasovsolver]
LocalTranslate = 1

[AMR]
transSplitPencilsOnlyForFace = 1
```

1) When constructing AMR pencils, only split pencils when the first face neighbours are on a higher refinement level, instead of extending up to the full VLASOV_STENCIL_WIDTH. Beyond the first (target) cell, source values are averaged over all four spatial cells at that distance.

   It should be noted that the AMR blockdata loading was reformatted in order to achieve this, meaning explicit prefetching of block data isn't meaningful anymore (but this hadn't been active in the code anyway and in recent (and re-done) tests hadn't given any discernable speed benefit).

2) Perform translation with local data only, removing interim communications between Cartesian directions (the remote neighbour contribution parts). This is achieved by, for each Cartesian translation direction, translating all local translated cells, and also all their face neighbours in the relevant direction. Only local cells have their values updated. This option induces some diffs at MPItask boundaries but should help with balancing the computational load, as if a task has a different amount of translations to perform in X, Y, and Z, it won't matter as long as the total is relatively well balanced. This has not yet been implemented for the non-AMR solver. DCCRG neighbourhoods have not been extended, so if close to refinement boundaries the last source cell has not been updated to local memory, the last good cell in that direction is duplicated.

   This is the most controversial change, but this should be useful especially in large simulations and is a proof-of-concept for what is needed in order to manage timeclasses and efficient GPU usage in the future.

3) (as part of 2, uses those same cell lists) When translating boundary cells, only translate in direction flowing into simulation domain, not along the boundary (as those values would not have been updated anyway).

Phiprof speedup for Magnetosphere_3D_small testpackage (using 16 nodes on Vorna):
```
dev
 | 147 | 3   | Propagate       | 1  | 87.06     | 96        | 87.28     | 19 | 86.77     | 57 | 39    | 1.145e+08 | 2.603e+06   | Cells/s         | 
 | 148 | 4   |  Spatial-space  | 1  | 55.61     | 63.88     | 55.76     | 55 | 55.37     | 62 | 39    | 1.793e+08 | 4.075e+06   | Cells/s         | 
Less pencil splitting
[...]

PR #1006, opened 2024-07-09
Ghost translation

markusbattarbee:
Replaces #609 
Incorporates #1000

This PR introduces ghost translation (previously known as local translation). Allows for running all Vlasov translation with coalesced MPI just before the actual call, reducing the overhead related to 3D imbalances. This comes at the cost of having to actually compute translation for a boundary of ghost cells. This also requries a grid neighborhood of VLASOV_STENCIL_WIDTH+1 which does incur some overhead.

Turn on ghost translation with `vlasovsolver.GhostTranslate` set to true. This current form of the PR forces this on so we get CI with it active, but it should probably be turned off as default.

The correctness of this method was validated with local tests but this PR should re-confirm it with CI. If so preferred, a smaller ghost neighbourhood can be translated instead of all required, resulting in some diffs at process boundaries, but less of an overhead. This might be useful for example for early flushing of a simulation. The size of the stencil to include in ghost translation can be set with`vlasovsolver.GhostTranslateExtent`. The parameter has a default of `0` which is automatically replaced with `VLASOV_STENCIL_WIDTH+1`. Values greater than that are capped. Thus, possible values are between `1` and `VLASOV_STENCIL_WIDTH+1`.


markusbattarbee:
WIP (e.g. the default handling breaks the ionosphere tests, I see) and evaluating performance bottlenecks.

grid.h
```diff
@@ -54,7 +54,13 @@ void initializeGrids(
 
     \param[in,out] mpiGrid The DCCRG grid with spatial cells
 */
-void balanceLoad(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid, SysBoundary& sysBoundaries);
+void balanceLoad(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid, SysBoundary& sysBoundaries,
+   bool doTranslationLists = true);
+
+/* helper for calculating AMR flags and cell lists and building pencils
+ */
+void prepareAMRLists(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid);
```
ykempf, line 62: Can go if not called outside grid.cpp?

parameters.h
```diff
@@ -60,6 +60,8 @@ struct Parameters {
                                         timestep if useCFLlimit is true. */
    static Real vlasovSolverMinCFL;   /*!< The minimum CFL limit for propagation of distribution function. Used to set
                                         timestep if useCFLlimit is true. */
+   static bool vlasovSolverGhostTranslate;   /*!< Flag for activating all-local ghost translation. */
+   static uint vlasovSolverGhostTranslateExtent;   /*!< Defule stencil size in all-local ghost translation. */
```
ykempf, line 64: define?

grid.h
markusbattarbee: Is actually called in grid.cpp before the function is defined, so is needed.

lkotipal: It can be forward declared in `grid.cpp` to reduce scope, but I'm not entirely sure either way is strictly better.

ursg:
#1060 merges this with current dev, to hopefully get rid of the merge conflicts.

PR #1008, opened 2024-07-19
Only proc boundary cells in one trans loop

ykempf:
In my hunt for the memory issue... I found this, and it ran without hideously crashing so can't be *that* wrong a change. Let's see with the CI diffs...


ykempf:
No significant speedup, as I feared. Well, up to y'all, this is anyway changing soon #TM with the incoming translation rehaul I assume.

PR #1012, opened 2024-08-14
Caching of DCCRG <-> FSgrid coupling

veetihaaja:
Changed it so that the coupling between the two grids is not recalculated every timestep, but only during load balancing, aka when it actually changes. The coupling is calculated and saved in vlasiator.cpp instead of the functions that use it, and then fed in as arguments.


markusbattarbee:
Excellent improvement!

Since it's something that only ever gets used from gridGlue, I would however recommend making those sets top-level variables in `gridGlue.cpp` instead of bulking up the main vlasiator.cpp and adding even more function arguments. Of course happy to hear counter-arguments!

Also, FYI, `getLocalCells();` only returns a cached value, so calling it again does nothing. There's a function used for re-calculating the local cells cache at https://github.com/fmihpc/vlasiator/blob/62ac639d96eba005b66ba589bd69c2093663db92/vlasiator.cpp#L283 which is already called inside `balanceLoad()`.

grid.cpp
```diff
@@ -376,9 +380,12 @@ void initializeGrids(
          calculateCellMoments(mpiGrid[cells[i]], true, true);
       }
    }
-   
+
+   // recompute coupling
+   computeCoupling(mpiGrid, cells, momentsGrid, onDccrgMapRemoteProcessGlobal, onFsgridMapRemoteProcessGlobal, onFsgridMapCellsGlobal);
```
ykempf, line 385: This call is not needed, nothing changed the load balance,/task layout/coupling between the previous call and this one.

fieldsolver/gridGlue.hpp
```diff
@@ -97,3 +111,69 @@ int getNumberOfCellsOnMaxRefLvl(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geomet
...
+     //compute to which processes this cell maps
+     std::vector<CellID> fsCells = mapDccrgIdToFsGridGlobalID(mpiGrid, dccrgCells[i]);
+
+     //loop over fsgrid cells which this dccrg cell maps to
+     for (auto const &fsCellID : fsCells) {
+       int process = momentsGrid.getTaskForGlobalID(fsCellID).first; //process on fsgrid
+       onDccrgMapRemoteProcess[process].insert(dccrgCells[i]); //add to map
+     }    
+  }
+}
```
ykempf, line 179: Please make sure all indentations is made with spaces, and 3 of them per level. :)

vlasiator.cpp
```diff
@@ -1110,6 +1113,9 @@ int main(int argn,char* args[]) {
          // Make sure the ionosphere communicator is up-to-date, in case inner boundary cells
          // moved.
          SBC::ionosphereGrid.updateIonosphereCommunicator(mpiGrid, technicalGrid);
+
+         // recompute coupling of grids after load balance
+         computeCoupling(mpiGrid, cells, momentsGrid, onDccrgMapRemoteProcessGlobal, onFsgridMapRemoteProcessGlobal, onFsgridMapCellsGlobal);
```
ykempf, line 1118: So how about we put the call inside `balanceLoad()` in `grid.cpp` instead?

vlasiator.cpp
```diff
@@ -647,6 +647,9 @@ int main(int argn,char* args[]) {
       );
    }
 
+   // computing coupling of grids (recalculated when load balancing)
+   computeCoupling(mpiGrid, cells, momentsGrid, onDccrgMapRemoteProcessGlobal, onFsgridMapRemoteProcessGlobal, onFsgridMapCellsGlobal);
+
```
ykempf, line 652: Also here, this call could just be put inside balanceLoad(). (see comment further down)

fieldsolver/gridGlue.hpp
```diff
@@ -97,3 +111,69 @@ int getNumberOfCellsOnMaxRefLvl(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geomet
...
+  onFsgridMapRemoteProcess  maps dccrg processes  (key) => set of dccrg cellIDs owned by dccrg-process that map to current rank fsgrid cells 
+  onFsgridMapCells          maps remote dccrg CellIDs to local fsgrid cells
+*/
+
+// this function is declared here as it is a template function
+
+template <typename T, int stencil> void computeCoupling(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
+							const std::vector<CellID>& cells,
+							FsGrid< T, stencil>& momentsGrid,
+              std::map<int, std::set<CellID> >& onDccrgMapRemoteProcess,
```
markusbattarbee, line 128: Could you just use the `onDccrgMapRemoteProcessGlobal` etc in this function without providing them as arguments?

vlasiator.cpp
```diff
@@ -647,6 +647,9 @@ int main(int argn,char* args[]) {
       );
    }
 
+   // computing coupling of grids (recalculated when load balancing)
```
markusbattarbee, line 650: This should not be needed, as we call it in `initializeGrids()` inside `grid.cpp`

vlasiator.cpp
```diff
@@ -1110,6 +1113,9 @@ int main(int argn,char* args[]) {
          // Make sure the ionosphere communicator is up-to-date, in case inner boundary cells
          // moved.
          SBC::ionosphereGrid.updateIonosphereCommunicator(mpiGrid, technicalGrid);
+
```
markusbattarbee, line 1116: Let's also instead of here, call it inside `balanceLoad()` inside `grid.cpp`, at line 607 or so (at the same place as re-calculating local cells cache)

markusbattarbee: ( I do appreciate that it's in the same place as the updateIonosphereCommunicator, but still, that would be my recommendation)

markusbattarbee (changes requested):
Could you please edit the source files you touched with the correct Vlasiator style in use? That's e.g. 3-space indentation. Now it's a bit of a mess.

grid.cpp
```diff
@@ -1544,6 +1551,10 @@ bool adaptRefinement(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGri
 
    mapRefinement(mpiGrid, technicalGrid);
```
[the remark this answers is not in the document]
ykempf: Actually no, `mapRefinement()` just looks at refinement levels and coordinates. (Also, it's called a couple more times elsewhere anyway.)

grid.cpp
```diff
@@ -353,6 +356,7 @@ void initializeGrids(
    volGrid.updateGhostCells();
    fsGridGhostTimer.stop();
    phiprof::Timer getFieldsTimer {"getFieldsFromFsGrid"};
+   computeCoupling(mpiGrid, cells, volGrid);
```
markusbattarbee: This isn't needed - coupling is still valid from the balanceLoad call on line 338.

ursg:
Still compiles after #1019 has been merged too, good.

PR #1014, opened 2024-08-22
Revert "Adding omp simd-pragmas to moment filtering"

markusbattarbee:
Reverts fmihpc/vlasiator#1002 to see if the simd-pragmas cause diffs in the dAMR testpackage test


markusbattarbee:
This confirmed, adding those pragmas caused diffs to damr. Closing.

PR #1015, opened 2024-08-22
turn off -mavx and -march=znver2 in testpackage compilation on Carrin…

markusbattarbee:
This might solve the issue of filtering omp simd pragmas causing diffs in testpackage_damr.


markusbattarbee:
Although running testpackage without vectorization flags might be a nice way to more "safely" evaluate how code changes affect the simulation output, it takes the testpackage information even further away from production flags. And, if we are doing optimizations, we are also interested in how large a magnitude change they have, even if it makes CI evaluation a bit more challenging.

The real reason though is that runtime increases too much with this. (order of 10 minutes for a TP run)

If we at some point decide to implement several different CI pipelines with varying levels of accuracy and runtime, we should re-visit this topic.

PR #1017, opened 2024-08-22
Split cell transfer into a number of passes in load balance.

ykempf:
Allows to avoid memory peaks and is adaptive.
Hard-coded to allow up to 10% of a tasks's blocks being sent in one pass.


ykempf:
Ready to merge if the test on Mahti reports sensible numbers of transfers.
@markusbattarbee:
```
grep num /scratch/project_2005018/202407_FIG_derived_memory_checks/FIG_CopyAndEXB_memreports_determine_chunks_from_blocks_shrink_1.1_cleanedup/slurm-3674502.out
```
will give the list of the number of chunks it took during that slot once it's run.

ykempf:
I added a parenthesis to this guy as suggested by some CI test, but it doesn't change the functionality so I guess it can go in soon?

markusbattarbee (approved):

PR #1019, opened 2024-08-22
Update vlasiator to use updated fsgrid constructor

cscjlan:
fsgrid doesn't use grid coupling to anything, so it was removed completely.

This should **not** be merged before the [PR of fsgrid](https://github.com/fmihpc/fsgrid/pull/31) is merged, since the hash of the submodule may change in the merge.


cscjlan:
Seems that `tools/vlsvdiff.cpp` doesn't include a header it needs (`#include <iomanip>`), which it got "accidentally" from `fsgrid.hpp`

markusbattarbee:
We have a merge day upcoming - is this ready to be moved out of draft status? (remembering that the fsgrid commit needs updating after FSgrid PR 31)

cscjlan:
Yes, this can be merged!

PR #1022, opened 2024-08-30
New amr criteria

lkotipal:
New criteria for refinement:

- Pressure anisotropy, refining if anisotropy is below `AMR.anisotropy_threshold`
- Vorticity scaled to Alfven velocity, usage similar to alpha parameters


parameters.cpp
```diff
@@ -167,6 +167,7 @@ Real P::alpha1CoarsenThreshold = -1.0;
 bool P::useAlpha2 = true;
 Real P::alpha2RefineThreshold = 0.5;
 Real P::alpha2CoarsenThreshold = -1.0;
+Real P::anisotropyThreshold = -1;
```
markusbattarbee, line 170: This could be named more clearly to be associated with AMR refinement criteria

vlasovsolver/cpu_moments.cpp
```diff
@@ -147,9 +144,12 @@ void calculateCellMoments(spatial_cell::SpatialCell* cell,
       pop.P[2] = mass*array[2];
```
markusbattarbee, line 144: For multipop shenanigans later on; I would recommend storing the full pressure tensor for all populations and refining if any population has a pressure anisotropy which would trigger the threshold.

projects/project.cpp
```diff
@@ -598,6 +598,13 @@ namespace projects {
             bool shouldRefine {(r2 < r_max2) && ((P::useAlpha1 ? cell->parameters[CellParams::AMR_ALPHA1] > P::alpha1RefineThreshold : false) || (P::useAlpha2 ? cell->parameters[CellParams::AMR_ALPHA2] > P::alpha2RefineThreshold : false))};
             bool shouldUnrefine {(r2 > r_max2) || ((P::useAlpha1 ? cell->parameters[CellParams::AMR_ALPHA1] < P::alpha1CoarsenThreshold : true) && (P::useAlpha2 ? cell->parameters[CellParams::AMR_ALPHA2] < P::alpha2CoarsenThreshold : true))};
 
+            // Pressure anisotropy forces refinement
```
markusbattarbee, line 601: Please elaborate (low pressure anisotropy; when below threshold, assume some kind of beam which will require higher resolution)

lkotipal:
Now that I'm off vacation @JonasSuni please give your PR review in the PR, not the commits 😄

JonasSuni:
> Now that I'm off vacation @JonasSuni please give your PR review in the PR, not the commits 😄

My comments are in the PR to your branch :)

vlasovsolver/cpu_moments.cpp
lkotipal: If we want it later on, I think we should implement it later on

projects/project.cpp
lkotipal: This is documented in the `parameters.cpp` description as it should. As for the scientific motivation, I guess that could be added there as well but I wouldn't be the one to ask about that :grin:

JonasSuni:
Hmm, I guess the CI was run again, but there's still diffs. Are they still an issue?

lkotipal:
Well the grid is the same, and this PR doesn't change anything except refinement behavior (with output after every pass iirc). So it's probably summation order, cosmic rays or whatever.

vlasovsolver/cpu_moments.cpp
markusbattarbee: To get our goals on the road, this can be postponed for a bit, but we cannot have code breaking multipop behaviour just because we aren't running multipop actively at this time. Please prepare a new PR fixing this issue.

One set of thresholds read as parameters is enough, but they should be compared against the rotated pressure tensor of each population as well as the total rotated pressure tensor.

spatial_cell_cpu.cpp
```diff
@@ -743,7 +743,7 @@ namespace spatial_cell {
          // Refinement parameters
          if ((SpatialCell::mpi_transfer_type & Transfer::REFINEMENT_PARAMETERS)){
             displacements.push_back(reinterpret_cast<uint8_t*>(this->parameters.data() + CellParams::AMR_ALPHA1) - reinterpret_cast<uint8_t*>(this));
-            block_lengths.push_back(sizeof(Real) * (CellParams::AMR_ALPHA2 - CellParams::AMR_ALPHA1 + 1)); // This is just 2, but let's be explicit
+            block_lengths.push_back(sizeof(Real) * (CellParams::AMR_VORTICITY - CellParams::AMR_ALPHA1 + 1)); // This is just 4, but let's be explicit.
```
ursg, line 746: This is a bit inelegant. If anybody adds a new refinement criterion, they need to remember changing this index, too (or otherwise spend two weeks scratching their head as to why it suddenly misbehaves on MPI boundaries).
I don't have any good suggestion how to make this prettier though, these being enum members and all.

warnings.txt
```diff
@@ -0,0 +1,34 @@
...
+make: *** [Makefile:256: KHB.o] Interrupt
+make: *** [Makefile:256: Larmor.o] Interrupt
+make: *** [Makefile:256: Magnetosphere.o] Interrupt
+make: *** [Makefile:256: MultiPeak.o] Interrupt
+make: *** [Makefile:256: Riemann1.o] Interrupt
+make: *** [Makefile:256: Shock.o] Interrupt
+make: *** [Makefile:256: Template.o] Interrupt
+make: *** [Makefile:256: test_fp.o] Interrupt
+make: *** [Makefile:256: testHall.o] Interrupt
+make: *** [Makefile:256: test_trans.o] Interrupt
```
ursg: Did this file slip into the PR by accident?

vlasovsolver/cpu_moments.cpp
lkotipal: > To get our goals on the road, this can be postponed for a bit, but we cannot have code breaking multipop behaviour just because we aren't running multipop actively at this time

Unless I'm missing something it's not breaking multipop behaviour? It's just saving the off-diagonal bulk pressure to `CellParams`, where we used to store the diagonal bulk pressure already. I can add the pop-wise elements to `Population.P` but chose not to for this PR because:
- It wasn't needed for the feature coded
- I genuinely have no idea how entangled that bit is. e.g. do we loop through _all_ (not three) elements of pressure somewhere?

markusbattarbee: Break not in the sense of crashing the code, but in the sense that it will not refine when just a single population fulfills the criterion, while bulk pressure anisotropy not. And that may then lead to loss of beams for a less-common population.

So at the point of this comment, add pop.P[3...5] additions, and manage them as well when evaluating pressure anisotropy. And you'll want to add per-population pressure anisotropy data reducers as well.

fieldsolver/gridGlue.cpp
```diff
@@ -383,6 +394,15 @@ void getFieldsFromFsGrid(
       mpiGrid[cellAggregate.first]->derivativesBVOL[bvolderivatives::dPERBZVOLdx] = cellAggregate.second.sums[FieldsToCommunicate::dPERBZVOLdx] / cellAggregate.second.cells;
       mpiGrid[cellAggregate.first]->derivativesBVOL[bvolderivatives::dPERBZVOLdy] = cellAggregate.second.sums[FieldsToCommunicate::dPERBZVOLdy] / cellAggregate.second.cells;
       mpiGrid[cellAggregate.first]->derivativesBVOL[bvolderivatives::dPERBZVOLdz] = cellAggregate.second.sums[FieldsToCommunicate::dPERBZVOLdz] / cellAggregate.second.cells;
+      mpiGrid[cellAggregate.first]->derivativesV[vderivatives::dVxdx] = cellAggregate.second.sums[FieldsToCommunicate::dVxdx] / cellAggregate.second.cells;
```
ykempf, line 398: So summary of our f2f discussion: we're doing here communications every step that are not needed that often, like BGBVOL, or never, like dVxdx, dVydy, dVzdz. Let's add a comment somewhere near here and optimize later. In the latest production run this function took 53 s out of 16000 s so it's minor.

datareduction/datareducer.cpp
```diff
@@ -2962,7 +2962,21 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
       }
       if(P::systemWriteAllDROs || lowercase == "vg_amr_alpha2") {
          outputReducer->addOperator(new DRO::DataReductionOperatorCellParams("vg_amr_alpha2",CellParams::AMR_ALPHA2,1));
-         outputReducer->addMetadata(outputReducer->size()-1,"","","$\\alpha_2","");
+         outputReducer->addMetadata(outputReducer->size()-1,"","","$\\alpha_2$","");
```
ykempf, line 2965: Can you add DROs for the velocity derivatives? We have a flag for all derivatives, and allow them individually too iirc.

projects/project.cpp
```diff
@@ -563,6 +563,81 @@ namespace projects {
...
+         // Induced refinement still possible just beyond that limit.
+         (xyz[0] < P::refinementMinX) || (xyz[0] > P::refinementMaxX)
+         || (xyz[1] < P::refinementMinY) || (xyz[1] > P::refinementMaxY)
+         || (xyz[2] < P::refinementMinZ) || (xyz[2] > P::refinementMaxZ)) {
+         shouldUnrefine = true;
+      }
+
+      return shouldUnrefine;
+   }
+
```
ykempf, line 640: 7 lines below: should that warn about all 4 criteria?

projects/project.cpp
```diff
@@ -595,56 +667,20 @@ namespace projects {
...
-            if(!shouldUnrefine
-               // If this cell is planned to remain at the current refinement level, but is outside the allowed refinement region,
-               // attempt to unrefine it instead. (If it is already at the lowest refinement level, DCCRG should not go belly-up.)
-               // Induced refinement still possible just beyond that limit.
-              && ((xyz[0] < P::refinementMinX) || (xyz[0] > P::refinementMaxX)
-               || (xyz[1] < P::refinementMinY) || (xyz[1] > P::refinementMaxY)
-               || (xyz[2] < P::refinementMinZ) || (xyz[2] > P::refinementMaxZ))) {
-               shouldUnrefine = true;
-            }
+            // TODO starting to look truly cursed. Might need refactoring
```
ykempf: Is that not better now that there's the two functions?

spatial_cell_cpu.cpp
ykempf: Yep, these are among the "few" ones where one really needs to remember.

ursg (approved):

datareduction/datareducer.cpp
ykempf: OK that'll come in in an upcoming PR.

PR #1024, opened 2024-09-03
Fix check vlasiator cfg script

ykempf:
Make cfg boundary = options case insensitive for the first letter
Maxwellian maxwellian Outflow outflow Ionosphere ionosphere Copysphere copysphere
Goes along with fixes in check_vlasiator_cfg.sh

Missing Copysphere in check_vlasiator_cfg.sh.
Also missing identification of incorrect boundary types loaded.
Also make them case insensitive.

Let's see if we had lurking buggy cfgs now in CI :).

### DIFF
--- a/sysboundary/sysboundary.cpp
+++ b/sysboundary/sysboundary.cpp
@@ -164,7 +164,7 @@ void SysBoundary::initSysBoundaries(Project& project, creal& t) {
    }
 
    for (it = sysBoundaryCondList.begin(); it != sysBoundaryCondList.end(); it++) {
-      if (*it == "Outflow") {
+      if (*it == "Outflow" || *it == "outflow") {
          this->addSysBoundary(new SBC::Outflow, project, t);
 
          anyDynamic = anyDynamic | this->getSysBoundary(sysboundarytype::OUTFLOW)->isDynamic();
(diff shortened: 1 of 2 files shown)


ursg:
Now we have *one more point* in the source tree where we'd have to add a new boundary name, if we ever happen to add a new one.
From a software engineering perspective, it would be nicer to actually have the list of valid boundary names in the vlasiator help output, and have this script parse it.

But in actual practical matters: Whoever will implement a new boundary in the future will surely write a test for it, and that will surely come with a cfg file, and this script would loudly complain in CI if it hasn't yet been adjusted yet, and then that person will just fix it. So maybe it's just fine like this.

ursg (approved):

sysboundary/sysboundary.cpp
```diff
@@ -164,7 +164,7 @@ void SysBoundary::initSysBoundaries(Project& project, creal& t) {
    }
 
    for (it = sysBoundaryCondList.begin(); it != sysBoundaryCondList.end(); it++) {
-      if (*it == "Outflow") {
+      if (*it == "Outflow" || *it == "outflow") {
```
lkotipal, line 167: 1. Make this into a `switch`
2. Just lowercase `*it`. `boost::algorithm::to_lower_copy` if we want to use that or the one-liner from [cppreference](https://en.cppreference.com/w/cpp/string/byte/tolower).

markusbattarbee: Let's not add any new boost requirements please. But we do already have a to-lowercase action in datareduction:
https://github.com/fmihpc/vlasiator/blob/f995f8eff292ee4294634e6c92545555b05bc75e/datareduction/datareducer.cpp#L48

lkotipal: Yeah I figured more boost deps might be a negative. Note some sources say the implementation in datareduction is UB/unsafe so see cppreference. We could just make this into a function in `common.h`.

markusbattarbee:
Now I wonder - if this makes Vlasiator accept a case-insensitive (or rather first letter case-insensitive) boundary type, are the parameters describing that specific boundary type also case-insensitive?

Issue #940, opened 2024-04-08
fluxfunction compilation broken


ykempf:
Building fluxfunction fails on Carrington. Are we missing a file in the submodule? Did I wrongly update the submodules? How did that get past the make tools CI thing if it's not my PEBKAC?
```
$ make fluxfunction                                                                                                                                                                                                                                                                                                   
mpicxx -O3 -fopenmp -funroll-loops -std=c++17 -mavx -march=znver2  -Wall -Wextra -Wno-unused -DPAPI_MEM -DUSE_JEMALLOC -DJEMALLOC_NO_DEMANGLE -MMD -D OMPI_SKIP_MPICXX -DPROFILE -DNDEBUG -DACC_SEMILAG_PQM -DTRANS_SEMILAG_PPM -I /proj/group/spacephysics/libraries/gcc/11.2.0/OpenMPI/4.1.1/carrington/phiprof/include -isystem /proj/group/spacephysics/librarie
s/gcc/11.2.0/OpenMPI/4.1.1/carrington/jemalloc/include -DDP -DSPF -DVEC8F_AGNER  -c tools/fluxfunction.cpp -I/proj/group/spacephysics/libraries/gcc/11.2.0/OpenMPI/4.1.1/carrington/vlsv -I ./submodules/vectorclass/ -I ./submodules/vectorclass-addon/vector3d/ -I./submodules/fsgrid/ -I/home/ykempf/proj/vlasiator  -Itools -o fluxfunction.o                   
In file included from /home/ykempf/proj/vlasiator/particles/boundaries.h:24,                                                                                                                                                                                                                                                                                        
                 from /home/ykempf/proj/vlasiator/particles/field.h:26,                                                                                                                                                                                                                                                                                             
                 from tools/fluxfunction.cpp:32:                                                                                                                                                                                                                                                                                                                    
/home/ykempf/proj/vlasiator/particles/particles.h:30:7: error: ‘Vec3d’ does not name a type; did you mean ‘Vec8d’?

cscjlan:
It got past the CI, because the CI doesn't build [fluxfunction](https://github.com/fmihpc/vlasiator/blob/c81422d8ce8b40306c200254affcd0446d9a7b16/.github/workflows/github-ci.yml#L267)

lassejsc:
Seems to be fixed + the fluxfunction is tested in CI these days

PR #1029, opened 2024-09-25
Remove vec from fluxfunction and fix compilation

ykempf:
This includes #1026 at the moment, apologies.

Should address #940.

Adds fluxfunction to the tools built in github CI even though we don't use it (yet). That binary is not uploaded, either.

Fixes compilation of fluxfunction and removes the vectorclass from there, and the couple files it uses from particles/. Otherwise in particles/ only converting Vec3d to Vec3Dd so it compiles.

Neither fluxfunction nor particle pusher are tested so far!


particles/field.h
```diff
@@ -60,79 +58,157 @@ struct Field
       }
    }
 
-   Vec3d getCell(int x, int y, int z) {
+   std::vector<double> getCell(int x, int y, int z) {
```
ursg: I would suggest that this be ```std::array<double, 3>```. Because why would you use a variable-sized vector here?

ursg (changes requested):
Well, hum. Functional thumbs up, because a working fluxfunction is obviously better than a broken one.

But instead of partially keeping with vectorclass and partially unrolling vector operations into handcrafted scalar operations, maybe the better way to go here would be to just switch to Eigen vectors, like in other parts of the code (like the ionosphere)?

Also, ```std::vector<double>``` as a vector type instead of ```std::array<double,3>``` really doesn't look right to me.

markusbattarbee:
+1 for using Eigen vectors instead.

ursg:
[This](https://github.com/ykempf/vlasiator/pull/16) Pull request against this PR introduces Eigen vectors instead.

particles/boundaries.h
```diff
@@ -140,7 +140,7 @@ struct ReflectBoundary : public Boundary
 
    // Vector to multiply with in order to flip velocity
    // vectors for our dimension
-   Vec3d flip_v;
+   Vec3Dd flip_v;
```
ykempf: But so we still want to keep these as `vectorclass`, the others as `Eigen` vectors?

ursg (approved):

.github/workflows/github-ci.yml
```diff
@@ -466,6 +469,13 @@ jobs:
         path: testpackage-output.tar.gz
       # Note: Testpackage output is further processed in the pr_report.yml workflow
       # (to produce Checks against pull requests)
+    - name: Run fluxfunction test
+      run: |
+        chmod +x $GITHUB_WORKSPACE/fluxfunction
+        srun --job-name fluxtest -M carrington -N 1 -c 1 --mem=1G bash -c "module purge; module load GCC/13.2.0; module load OpenMPI/4.1.6-GCC-13.2.0 ; module load PMIx/4.2.6-GCCcore-13.2.0; module load PAPI/7.1.0-GCCcore-13.2.0; $GITHUB_WORKSPACE/fluxfunction testpackage/run_*/Magnetosphere_small/bulk.0000001.vlsv equatorial.bin"
+        srun --job-name fluxtest -M carrington -N 1 -c 1 --mem=1G bash -c "module purge; module load GCC/13.2.0; module load OpenMPI/4.1.6-GCC-13.2.0 ; module load PMIx/4.2.6-GCCcore-13.2.0; module load PAPI/7.1.0-GCCcore-13.2.0; $GITHUB_WORKSPACE/fluxfunction testpackage/run_*/Magnetosphere_polar_small/bulk.0000001.vlsv polar.bin"
+        diff -q equatorial.bin /proj/group/spacephysics/vlasiator_testpackage/CI_reference/equatorial.bin
```
ykempf, line 586: OK so this now reported nothing as expected, but if it were saying there's a diff it wouldn't be very prominent. How do we raise that?

particles/boundaries.h
ykempf: Uh wait that's old.

ykempf: But still a valid question from me.

.github/workflows/github-ci.yml
```diff
@@ -430,6 +431,8 @@ jobs:
         chmod +x $GITHUB_WORKSPACE/vlasiator
         chmod +x $GITHUB_WORKSPACE/vlsv*_DP
         cd testpackage
+        # Delete any old run directories
+        rm -rf run_20*
```
markusbattarbee, line 544: This will break in 75 years time! ;)

.github/workflows/github-ci.yml
markusbattarbee: pipe the output into a variable, and if the variable is acceptable, touch a file? then can have a yaml action checking for that file and failing on file not found?
@ursg might have a more elegant solution.

particles/scenario.cpp
```diff
@@ -22,6 +22,12 @@
 #include <random>
 #include <iostream>
 #include "scenario.h"
+#include "field.h"
+
+#define cross_product(av,bv) (av).cross(bv)
+#define dot_product(av,bv) (av).dot(bv)
+#define vector_length(v) (v).norm()
+#define normalize_vector(v) (v).normalized()
```
markusbattarbee, line 30: These are already defined in relativistic_math.h

particles/relativistic_math.h
```diff
@@ -20,9 +20,14 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+#include <Eigen/Dense>
+#define Vec3d Eigen::Vector3d
+#define cross_product(av,bv) (av).cross(bv)
+#define dot_product(av,bv) (av).dot(bv)
+#define vector_length(v) (v).norm()
+#define normalize_vector(v) (v).normalized()
+
```
markusbattarbee, line 29: place in fields.h instead?

particles/particle_post_pusher.cpp
```diff
@@ -37,12 +37,17 @@
 #include "scenario.h"
 #include "boundaries.h"
 
+#define cross_product(av,bv) (av).cross(bv)
+#define dot_product(av,bv) (av).dot(bv)
+#define vector_length(v) (v).norm()
+#define normalize_vector(v) (v).normalized()
+
```
markusbattarbee, line 44: already defined in relativistic_math.h

particles/histogram.h
```diff
@@ -185,7 +187,7 @@ class LinearHistogram2D : public Histogram2D
 
       virtual void addValue(Vec2d value, double weight=1.) {
          value -= low;
-         value /= high - low;
+         value = value.cwiseQuotient(high - low);
```
markusbattarbee, line 190: should this also be a macro like the other vector operations?

particles/scenario.cpp
ursg: I simply used these defines as a drop-in replacement for vectorclass. I originally took them from sysboundaries/ionosphere.cpp

So... maybe a global Eigen-as-vectorclass-header would make sense?

PR #1030, opened 2024-09-26
Jemalloc purge in amr

ykempf:
This significantly pushed down HWM in my 200-node FIE-based test.


ykempf:
Oops. See https://github.com/markusbattarbee/vlasiator-public/pull/18 instead, then go back to #1027.

PR #1034, opened 2024-10-08
Skip simd 2

ykempf:


markusbattarbee:
This is not fixing the jitter we're seeing after all, so let's close it. SIMD operations should indeed be nice.

PR #1036, opened 2024-10-09
Fix refine/coarsen logging

lkotipal:
Use correct datatype for logging coarsened cells. Oops...


markusbattarbee:
So the counts still don't quite add up.

According to this line:
https://github.com/fmihpc/dccrg/blob/d7a7746bcbcf53909acfd85d19dd29de8b0f123a/dccrg.hpp#L2495C6-L2495C7
each process tags only one of all siblings to unrefine. However, I guess if two siblings are on different processes, that would count as two unrefines, and if siblings exist on three different processes, then three.

The easiest workaround might be to actually store all siblings in this set, so that the reducet unrefine count would include all siblings, but change it so that when executing unrefines, siblings beyond the first are ignored. Thoughts?

ykempf:
I suppose one part of the question is: do we actually need the exact values? But I fear part of the answer is: as soon as we decide that "no", next week we'll start debugging a case that will require the actually correct values. So... I suppose we do want them?

lkotipal:
> So the counts still don't quite add up.
> 
> According to this line: https://github.com/fmihpc/dccrg/blob/d7a7746bcbcf53909acfd85d19dd29de8b0f123a/dccrg.hpp#L2495C6-L2495C7 each process tags only one of all siblings to unrefine. However, I guess if two siblings are on different processes, that would count as two unrefines, and if siblings exist on three different processes, then three.
> 
> The easiest workaround might be to actually store all siblings in this set, so that the reducet unrefine count would include all siblings, but change it so that when executing unrefines, siblings beyond the first are ignored. Thoughts?

I could just refactor the loop on https://github.com/fmihpc/dccrg/blob/d7a7746bcbcf53909acfd85d19dd29de8b0f123a/dccrg.hpp#L3503 earlier on, like the end of `override_unrefines`?

markusbattarbee (approved):

PR #1041, opened 2024-10-11
Makefile (and small library build script fix) for Meluxina.

ursg:
Based on Lumi's 24.3 makefile, it managed to compile and run a Magnetosphere_small test on a single node quite painlessly.


ursg:
I guess nobody else can verify this, so you'll just have to take my word for it.

PR #1042, opened 2024-10-17
Remove velocity block params.

ursg:
Fully redundant information about velocity space coordinates and DV, duplicated billions of times? Away with it!


ursg:
Whoops, looks like I broke restart reading there.

ursg:
Testpackage looking good, and at least there doesn't appear to be any significant slowdown. Next up: Let's check how this affects memory usage at larger scales.

ursg:
With some new ideas how to employ blockparameters in future optimizations, this idea can be discarded.

PR #1045, opened 2024-10-24
Chop off cells in x

ykempf:
This introduces a parameter giving a number of +x base-grid cells that get shoved into DO_NOT_COMPUTE, effectively reducing the size of the computed domain by a certain slab of cells.

We have operational need for this and it seems to work for the very specific purpose at hand (no refinement nearby, only this wall of the domain, no particular resetting of cells or so). Anything more general is going to require a lot more work, and we don't want to ever have to reuse this. But I suppose we want to keep a trace of it as it'll likely enter production in the next days.

If nothing else, a couple :eyes: for cross-checking before said production would be nice.

### DIFF
--- a/parameters.cpp
+++ b/parameters.cpp
@@ -61,6 +61,8 @@ uint P::xcells_ini = numeric_limits<uint>::max();
 uint P::ycells_ini = numeric_limits<uint>::max();
 uint P::zcells_ini = numeric_limits<uint>::max();
 
+uint P::chopOffNCellsInX = 0;
+
 Real P::t = 0;
 Real P::t_min = 0;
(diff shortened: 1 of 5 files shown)


ursg:
This also includes the DT2 ghost update fix right now. I suggest these two are disentangled?

ursg (approved):
Functionally, this looks ok. It's not elegant or anything, so I'm not even sure we'd want to merge this into dev in this way.
(If anything, it should be turned into a "bounding box" feature, similarly to what the elVentana branch contains. But maybe we don't bother beyond the current use case?)

ykempf:
Now with only this specific commit.

PR #1047, opened 2024-10-31
Use curl instead of wget in build_libraries.sh

ursg:
Concretely, this is importint for systems that can only access the internet via a SOCKS proxy, which wget doesn't support.


ursg:
This looks good now. Feel free to push the button.

PR #1048, opened 2024-10-31
Less L2 boundary VDFs

ykempf:
...and less communication thereof.

Compiles.

Let the CI do its magic.

PR #1049, opened 2024-11-01
SysBoundary communicate VDFs only in smaller neighborhood

ykempf:


ykempf:
This is a subset of #1048 and will need more features/changes in any case.

ykempf:
So this is still not doing fluffy copy, which is used in copysphere, correctly. Will be fixed soon. I came back to this branch/PR as I am not taking in everything that was pushed to #1048. Mainly, I let it create (local) VDFs and compute their moments when needed, instead of adding spaghetti to save on that. What's expensive ar second-neighbour VDFs and their communication, not the actual local L2 VDFs.

datareduction/datareducer.cpp
```diff
@@ -314,6 +314,10 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
             const std::string& pop = species.name;
             outputReducer->addOperator(new DRO::DataReductionOperatorPopulations<Real>(pop + "/vg_rho", i, offsetof(spatial_cell::Population, RHO), 1));
             outputReducer->addMetadata(outputReducer->size()-1,"1/m^3","$\\mathrm{m}^{-3}$","$n_\\mathrm{"+pop+"}$","1.0");
+            outputReducer->addOperator(new DRO::DataReductionOperatorPopulations<Real>(pop + "/vg_rho_r", i, offsetof(spatial_cell::Population, RHO_R), 1));
```
ykempf: Ditch those again

sysboundary/ionosphere.cpp
```diff
@@ -3474,6 +3474,20 @@ namespace SBC {
 
       calculateCellMoments(&templateCell,true,false,true);
 
+      for (uint popID=0; popID<getObjectWrapper().particleSpecies.size(); ++popID) {
```
ykempf: Are these needed?

sysboundary/ionosphere.h
```diff
@@ -395,6 +394,10 @@ namespace SBC {
          const uint popID,
          const bool calculate_V_moments
       ) override;
+      virtual void setupL2OutflowAtRestart(
```
markusbattarbee: Instead of having these warnings in all the "wrong" boundaries, I think it should exist in the base class as a warning and be just missing from all the others.

sysboundary/outflow.cpp
```diff
@@ -420,6 +425,96 @@ namespace SBC {
...
+            for (uint popID=0; popID<getObjectWrapper().particleSpecies.size(); ++popID) {
+               const OutflowSpeciesParameters& sP = this->speciesParams[popID];
+               for(uint i=0; i<6; i++) {
+                  if(isThisCellOnAFace[i] && facesToProcess[i] && !sP.facesToSkipVlasov[i]) {
+                     switch(sP.faceVlasovScheme[i]) {
+                        case vlasovscheme::NONE:
+                           break;
+                        case vlasovscheme::COPY:
+                           if(mpiGrid[cellID]->sysBoundaryLayer == 1) {
+                              // at this stage we don't actually need the VDFs yet I think. And they were crashing LUMI apparently?!
```
markusbattarbee: Outdated comment, right?

sysboundary/outflow.cpp
```diff
@@ -420,6 +425,96 @@ namespace SBC {
...
+      const vector<CellID>& cells = getLocalCells();
+      #pragma omp parallel
+      {
+         #pragma omp for schedule(guided,1)
+         for(uint i=0; i<cells.size(); i++) {
+            const CellID cellID = cells[i];
+            if(mpiGrid[cellID]->sysBoundaryFlag != this->getIndex()) {
+               continue;
+            }
+
```
markusbattarbee, line 451: Put the check for `if(mpiGrid[cellID]->sysBoundaryLayer == 1) {` here to skip earlier for cells we don't want to evaluate?

sysboundary/outflow.cpp
```diff
@@ -420,6 +425,96 @@ namespace SBC {
...
+         #pragma omp barrier // maybe useless
+
+         // then 2nd pass and copy from the closest L1 outflow neighbor
+         #pragma omp for schedule(guided,1)
+         for(uint i=0; i<cells.size(); i++) {
+            const CellID cellID = cells[i];
+            if(mpiGrid[cellID]->sysBoundaryFlag != this->getIndex()) {
+               continue;
+            }
+
```
markusbattarbee, line 492: Put the check for `if(mpiGrid[cellID]->sysBoundaryLayer == 2) {` here to skip earlier for cells we don't want to evaluate?

sysboundary/sysboundarycondition.cpp
```diff
@@ -384,24 +408,31 @@ namespace SBC {
...
+      if(copyMomentsOnly) {
+         to->get_population(popID).RHO = from->get_population(popID).RHO;
          if (copy_V_moments) {
             to->get_population(popID).RHO_V = from->get_population(popID).RHO_V;
          } else {
             to->get_population(popID).RHO_R = from->get_population(popID).RHO_R;
          }
-         
          for (uint i=0; i<3; i++) {
+            to->get_population(popID).V[i] = from->get_population(popID).V[i];
```
markusbattarbee, line 419: Uhh should this be here? and the same for `P[i]` on line 427. Might change behaviour?

spatial_cells/spatial_cell_cpu.hpp
```diff
@@ -95,7 +95,10 @@ namespace spatial_cell {
             max_dt[i] = 0;
          }
          for (uint i=0; i<3; ++i) {
-            V[i] = V_R[i] = V_V[i] = P[i] = P_R[i] = P_V[i] = 0;
+            V[i] = V_R[i] = V_V[i] = 0;
+         }
+         for (uint i=0; i<6; i++) {
```
markusbattarbee, line 100: Can you please add these fixes to spatial_cell_gpu.hpp as well?

markusbattarbee (changes requested):
A Few minor suggestions still. But on a basic level this has been thoroughly tested in FIF so is fundamentallly sounds (though indeed total removal of L2 VDFs and their initialization as well would be a good idea at some point IMO)

sysboundary/outflow.cpp
ykempf: yep

sysboundary/outflow.cpp
ykempf: Bit sneaky, but yes, I comment on this so whoever changes next reverts it.

sysboundary/outflow.cpp
ykempf: Same story.

sysboundary/sysboundarycondition.cpp
ykempf: I think in my finicky getting-the-very-beginning-diff-free I needed those too. But they get recomputed at the right spot when needed so I think we're good?

ykempf: `calculateInterpolatedVelocityMoments()` is called right after boundaries in the main loop.

spatial_cells/spatial_cell_cpu.hpp
ykempf: willco

sysboundary/sysboundarycondition.cpp
markusbattarbee: Ah, right, I remembered that the call to interpolated moments needs to sometimes write values `.P` and sometimes `.P_DT2`, but that's only for the total cellparams. Population-specific values do not keep a separate `_dt2` version. So this should be safe indeed.

markusbattarbee (approved):

sysboundary/sysboundarycondition.h
```diff
@@ -161,6 +161,14 @@ namespace SBC {
             const bool calculate_V_moments
         )=0;
 
+        virtual void setupL2OutflowAtRestart(
+            dccrg::Dccrg<SpatialCell, dccrg::Cartesian_Geometry>& mpiGrid
+        ) {
+            std::cerr << "ERROR: base class SysBoundaryCondition::setupL2OutflowAtRestart called!" << std::endl;
+        }
```
ursg, line 168: I still continue to be in favour of actually *not* defining the base class implementations of virtual functions that we don't actually want called anyway. That way, it becomes a compile-time problem.

Does that work in this case, or is the code still too spaghetti to allow for it?

ursg (approved):
Looks good to me, with just a minor style remark

sysboundary/sysboundarycondition.h
ykempf: We call loops over boundaries, as "always" – if not like this we need an implementation in every other derived class too, which I think we just removed recently?...

PR #1050, opened 2024-11-04
No L2 NOT_SYSBOUNDARY EXB nudge

ykempf:
Olkaapa hyvät


sysboundary/ionosphere.cpp
```diff
@@ -3128,19 +3127,7 @@ namespace SBC {
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
             case ForceL2EXB:
```
ursg: This name is a bit misleading now. It is called "ForceL2EXB", but it is *no longer* actually forcing the L2 E×B.

ursg (approved):
Looks sane. The only thing this change does, is *no longer* forcing L2 cells, and doesn't otherwise touch any functionality.

markusbattarbee:
Looks good, pending CI

PR #1052, opened 2024-11-04
Multipop pressure anistropy

lkotipal:
Continuation of #1022, support storing off-diagonal pressure in populations and use _minimum_ of all pops and total for refinement.


lkotipal:
Committing 3c44999 and reverting it on #1022 apparently "merged" this one. Oops...

PR #1053, opened 2024-11-05
Node hours report

ArnaudLalague:
Add core-hours and node-hours to the logfile for every step
Fixed with Yann

### DIFF
--- a/vlasiator.cpp
+++ b/vlasiator.cpp
@@ -907,9 +907,9 @@ int main(int argn,char* args[]) {
          double remainingTime=min(timePerStep*(P::tstep_max-P::tstep),timePerSecond*(P::t_max-P::t));
          time_t finalWallTime=time(NULL)+(time_t)remainingTime; //assume time_t is in seconds, as it is almost always
          struct tm *finalWallTimeInfo=localtime(&finalWallTime);
-         logFile << "total node-hours " << nNodes*(after - startTime)/3600 << " h" << endl;
+         logFile << "total node-hours " << nNodes*(currentTime - startTime)/3600 << " h" << endl;
          #if _OPENMP
-            logFile << "total core-hours " << omp_get_max_threads()*mpiProcs*(after - startTime)/3600 << " h" << endl;
+            logFile << "total core-hours " << omp_get_max_threads()*mpiProcs*(currentTime - startTime)/3600 << " h" << endl;
          #endif
          logFile << "(TIME) current walltime/step " << timePerStep<< " s" <<endl;
          logFile << "(TIME) current walltime/simusecond " << timePerSecond<<" s" <<endl;
--- a/vlasiator.cpp
+++ b/vlasiator.cpp
@@ -417,15 +417,42 @@ int main(int argn,char* args[]) {
          exit(1);
       }
    }
+
    int mpiProcs;
    MPI_Comm_size(MPI_COMM_WORLD,&mpiProcs);
+
+   char nodename[MPI_MAX_PROCESSOR_NAME]; 
+   int namelength, nodehash;
+   int nProcs, nodeRank, interRank;
+   int nNodes;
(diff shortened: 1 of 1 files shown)


vlasiator.cpp
```diff
@@ -417,17 +417,42 @@ int main(int argn,char* args[]) {
...
+
+   int mpiProcs;
+   MPI_Comm_size(MPI_COMM_WORLD,&mpiProcs);
+
+   char nodename[MPI_MAX_PROCESSOR_NAME]; 
+   int namelength, nodehash;
+   int nProcs, nodeRank, interRank;
+   int nNodes;
+   const double GiB = pow(2,30);
+   const double TiB = pow(2,40);
```
ursg: These are never actually used below, as far as I can see?

vlasiator.cpp
```diff
@@ -417,17 +417,42 @@ int main(int argn,char* args[]) {
...
+   
+   //intra-node communicator
+   MPI_Comm_split(MPI_COMM_WORLD, nodehash, myRank, &nodeComm);
+   MPI_Comm_rank(nodeComm,&nodeRank);
+   //create communicator for inter-node communication
+   MPI_Comm_split(MPI_COMM_WORLD, nodeRank, myRank, &interComm);
+   MPI_Comm_rank(interComm, &interRank);
+   MPI_Comm_size(interComm, &nNodes);
+
+   MPI_Comm_free(&interComm);
```
ursg, line 445: The nodeComm is never freed.

vlasiator.cpp
```diff
@@ -884,13 +909,18 @@ int main(int argn,char* args[]) {
          double remainingTime=min(timePerStep*(P::tstep_max-P::tstep),timePerSecond*(P::t_max-P::t));
          time_t finalWallTime=time(NULL)+(time_t)remainingTime; //assume time_t is in seconds, as it is almost always
          struct tm *finalWallTimeInfo=localtime(&finalWallTime);
+         logFile << "(TIME) total node-hours " << nNodes*(currentTime - startTime)/3600 << " h" << endl;
+         #if _OPENMP
+            logFile << "(TIME) total core-hours " << omp_get_max_threads()*mpiProcs*(currentTime - startTime)/3600 << " h" << endl;
```
ykempf: Let's change the wording to "current" or "cumulated" here, "total" at the end (our own copy-paste problem here :)).

ursg (approved):
Nice addition to the output!

vlasiator.cpp
```diff
@@ -1360,6 +1391,11 @@ int main(int argn,char* args[]) {
       double timePerSecond=double(after  - startTime) / (P::t-P::t_min+DT_EPSILON);
       logFile << "(MAIN): All timesteps calculated." << endl;
       logFile << "\t (TIME) total run time " << after - startTime << " s, total simulated time " << P::t -P::t_min<< " s" << endl;
+      logFile << "\t (TIME) total node-hours " << nNodes*(after - startTime)/3600 << " h" << endl;
+      #if _OPENMP
+         logFile << "\t (TIME) total core-hours " << omp_get_max_threads()*mpiProcs*(after - startTime)/3600 << " h" << endl;
```
ykempf: And I would suggest to reorder and have all of these (above and here) say sth like "(TIME) total x.y node-hours" instead of "node-hours x.y h"

ykempf (changes requested):

vlasiator.cpp
ykempf: Good catch. Instead in between we were freeing interComm twice. :grimacing:

vlasiator.cpp
ykempf: Right these came from where we copied the rest of the code and can go indeed.

ykempf (approved):

markusbattarbee (approved):

PR #1054, opened 2024-11-06
Build testpackage data with new gcc version on carrington.

ursg:
Otherwise, the "generate testpackage data" Github Action failed.


ursg:
(Merging this before CI is even run through, because the only changes are anyway in the generate reference data Actions description.)

PR #1055, opened 2024-11-06
Limit tracing volume

ykempf:
Introduce cfg parameters to limit the volume of field tracing within x/y/z min/max. Needed to salvage tracing in certain production run.


ykempf:
No niin. Other compilers, other customs. But with the additional casting also the ionosphere miniapp tests compile in the CI system.

markusbattarbee (approved):
merg!

PR #1059, opened 2024-11-11
Add SIGTERM signal handler, for slurm preemption.

ursg:
When preempting, slurm sends a SIGTERM signal and gives the process two minutes time to gracefully terminate itself. Writing a restart immediately might work in that situation.

If the university IT department actually wants to take slurm preemption into use, we better be prepared. :)

Tested (by hand) with a small testpackage job.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,10 @@ COMPFLAGS += -DACC_SEMILAG_PQM -DTRANS_SEMILAG_PPM
 #May cause problems
 #COMPFLAGS += -DCATCH_FPE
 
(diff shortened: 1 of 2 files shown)


ursg:
Just tested this on mahti with a small job with 1 minute runtime limit. Works as intended, and wrote a restart.

But now that we have both recovers and restarts. What is it that we actually prefer here?
- On one hand: We write recovers "just in case" it would crash, whereas here we know that it will be killed. So that would favour a restart write.
- On the other hand: We know that we only have maybe a minute left, so writing a recover (potentially to a fast flash partition) might be the only thing that's actually possible.

markusbattarbee (approved):
I'd favour using the recover functionality.

PR #1060, opened 2024-11-12
Ghost translation dev merge

ursg:
This is a merge of #1006 with the current, updated dev branch after the merges of the latter days.

Compiles cleanly, but let's run CI to see if I broke anything.


parameters.cpp
```diff
@@ -422,7 +422,7 @@ bool P::addParameters() {
                         "ig_precipitation ig_deltaphi "+
                         "ig_inplanecurrent ig_b ig_e vg_drift vg_ionospherecoupling vg_connection vg_fluxrope fg_curvature "+
                         "vg_amr_drho vg_amr_du vg_amr_dpsq vg_amr_dbsq vg_amr_db vg_amr_alpha1 vg_amr_reflevel vg_amr_alpha2 "+
-                        "vg_gridcoordinates fg_gridcoordinates ");
+                        "vg_amr_translate_comm vg_gridcoordinates fg_gridcoordinates vg_pressure_anisotropy vg_amr_vorticity");
```
markusbattarbee: vg_amr_translate_comm is removed in 1006 so should not be re-introduced

markusbattarbee (changes requested):

ursg:
Ok, nope. This segfaults in Magnetosphere_polar_small in the testpackage, so I guess this needs to be merged more carefully.

ursg:
Ah yes, one probably reasonably important hunk had fallen under the table in the merge there:
```c++
   // Prepare ghost translation cell lists and build pencils for translation.
   if (doTranslationLists) {
      prepareAMRLists(mpiGrid);
   }
```

parameters.cpp
ursg: fixed in the force push below.

ursg:
And we're green across the board here. The testpackage has the same size mismatches for some AMR meshes (due to neighborhoods now being more snugly constructed), and in general the diffs look exactly the same as in #1006. So I am in favour of merging this now. Any objections?

PR #1061, opened 2024-11-12
Multipop pressure anistropy

lkotipal:
Continuation of https://github.com/fmihpc/vlasiator/pull/1022, support storing off-diagonal pressure in populations and use minimum of all pops and total for refinement.
Redo of #1052.

### DIFF
--- a/fieldsolver/derivatives.cpp
+++ b/fieldsolver/derivatives.cpp
@@ -740,6 +740,32 @@ static Real calculateU(SpatialCell* cell)
       (rho > EPS ? (pow(p[0], 2) + pow(p[1], 2) + pow(p[2], 2)) / (2.0 * cell->parameters[CellParams::RHOM]) : 0.0); // Kinetic energy
 }
 
+/*! \brief Calculates pressure anistotropy from B and P
+ *  \param P elements of pressure order in order: P_11, P_22, P_33, P_23, P_13, P_12
+ */
+static Real calculateAnistropy(const std::array<Real, 3>& B, const std::array<Real, 6>& P)
+{
+   // Now, rotation matrix to get parallel and perpendicular pressure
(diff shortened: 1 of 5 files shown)


ursg (approved):
The back- and forth handling of P as a matrix, then a bunch of floats, then an array is a bit unpretty.

But actually it is *less* unpretty than before.

lkotipal:
> The back- and forth handling of P as a matrix, then a bunch of floats, then an array is a bit unpretty.
> 
> But actually it is _less_ unpretty than before.

Considering changing a whole bunch of these variables to `std::vector`s for nicer initialization. I'm not sure whether there's any actual reason to use `std::array`, really.

fieldsolver/derivatives.cpp
```diff
@@ -855,11 +862,22 @@ void calculateScaledDeltas(
...
+   Real amr_vorticity {-1.0}; // Error value
    if (maxV > EPS) {
       amr_vorticity = vorticity * cell->parameters[CellParams::DX] / maxV;
    }
 
+   std::array<Real, 6> myPressure {cell->parameters[CellParams::P_11], cell->parameters[CellParams::P_22], cell->parameters[CellParams::P_33], cell->parameters[CellParams::P_23], cell->parameters[CellParams::P_13], cell->parameters[CellParams::P_12]};
+   Eigen::Matrix3d rot = Eigen::Quaterniond::FromTwoVectors(Eigen::Vector3d{myB[0], myB[1], myB[2]}, Eigen::Vector3d{0, 0, 1}).normalized().toRotationMatrix();
+   Real Panisotropy {calculateAnisotropy(rot, myPressure)};
+   for (const auto& pop : cell->get_populations()) {
+      // TODO I hate this. Change all this crap to std::vectors?
```
ursg, line 874: I still think that, instead of adding this comment, you should just change it. :)

```c++
std::array<Real, 6> popP = *reinterpret_cast<std::array<Real, 6> *>(pop.P);
```
Should do the trick, shouldn't it?

ursg (approved):

PR #1063, opened 2024-11-13
Update C++ standard from C++17 to C++20 in Makefile.lumi_gnu_2403

JonasSuni:


ykempf:
Unless you're trying to achieve some PR badge, you may upgrade other Makefiles in this same PR. :)

ursg (approved):

PR #1068, opened 2024-12-09
Inspection script mechanism to use srun overlap. Works on LUMI at least.

ykempf:
Runs once for every node as running on >1 node srun overlap fails on LUMI at the moment.


ursg (approved):
I like this!

tools/slurm/overlap_inspection.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+# Modify target JOBID below, then execute to run the inspector.sh on every node for every srun vlasiator instance.
+
+export JOBID=8513363
```
markusbattarbee: Could take the jobid as a command line parameter, would help with usability

PR #1070, opened 2024-12-17
Wrap all fsgrids in a FsGridWrapper struct, pass that around.

ursg:
In order to de-bloat many function signatures, and enable possible separation-of-concerns looping constructs that can simply operate on all fsgrids.

This is a massive change, touching basically all files that have anything to do with the fields, of course. But the code looks somewhat nicer afterwards!

Things to still reconsider:
 - The ```FsGridWrapper``` struct is now defined in the object_wrapper.h source file. Should it rather move to, say, fs_common.h?
 - Now that all grids are typically accessed as ```fsgrids.somethingGrid```, should they all be renamed to ```fsgrids.something```? Otherwise we are in the department of redundancy department.


ursg:
This PR is probably already superseded by [CSC's more comprehensive refactoring approach](https://github.com/cschpc/vlasiator/commits/fsgrid_refactoring/).

markusbattarbee:
Indeed, superceded.

PR #1072, opened 2024-12-19
Run carrington CI runs with proper slurm placement parameters.

ursg:
hopefully, this will make our CI runs faster!

### DIFF
--- a/testpackage/small_test_carrington.sh
+++ b/testpackage/small_test_carrington.sh
@@ -51,8 +51,8 @@ tasks_per_node=$(echo $units_per_node $t  | gawk '{print $1/$2}')
 export OMP_NUM_THREADS=$t
 
 #command for running stuff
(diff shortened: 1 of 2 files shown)


markusbattarbee:
Speedups in some situations, it seems. Nice to see placement output from mpirun, but that needs removing as it makes output too long and truncates the upload.

Also, the new launch needs to go also into the non-CI carrington run script. We might also want to see if the funny large mpirun shebang can be removed by running some multi-node tests with and without.

Also, I guess we could try to get srun placements working as well... the more options the merrier. Next year, I guess!

markusbattarbee:
Ah but according to 
https://version.helsinki.fi/it-for-science/hpc/-/issues/549
we should use `export UCX_NET_DEVICES=eth0` and launch with `srun`. Right? So this can be closed?

PR #1075, opened 2025-01-16
Use ARCH=arch in VLSV build

ykempf:
Or it still uses std=c++0x...


ykempf:
Merge #1068 first, I again wasn't able to do a clean branch.

markusbattarbee (approved):

PR #1087, opened 2025-01-30
Try building will all debug flags activated in CI to check the debug …

ykempf:
…code paths.


ykempf:
All right, the Ukko GPU testpackage build with debug flags runs too. Have I used all the relevant debug flags? I just copied the list from Makefile.

ykempf:
All right, these new checks work, so no rotten debug code behind these flags. Or at least, that code compiles...

ykempf:
up? :)

.github/workflows/github-ci.yml
```diff
@@ -182,6 +182,35 @@ jobs:
...
+      uses: actions/download-artifact@v4
+      with:
+        name: libraries
+    - name: Unpack libraries
+      run: tar --zstd -xvf libraries.tar.zstd
+    - uses: ursg/gcc-problem-matcher@master
+    - name: Compile vlasiator (Release build)
+      run: |
+        VLASIATOR_ARCH=github_actions make clean
+        VLASIATOR_ARCH=github_actions COMPFLAGS="-DDEBUG_VLASIATOR -DDEBUG_SOLVERS -DDEBUG_IONOSPHERE -DHASHINATOR_DEBUG -DDEBUG_SPATIAL_CELL -DDEBUG_VMESH -DDEBUG_VBC -DDEBUG_ACC " make -j 3
```
markusbattarbee, line 206: this does not need `-DHASHINATOR_DEBUG` (though it does no harm either)

.github/workflows/github-ci.yml
```diff
@@ -267,6 +296,46 @@ jobs:
           path: vlasiator
           if-no-files-found: error
 
+  build_testpackage_debug_flags_ukkoGPU:
```
markusbattarbee, line 299: I'd prefer this to be in line with #1100 so ukko_dgx and with appropriate modules.

Also, might be worth building two versions: with these flags, and also with an extra `-DUSE_WARPACCESSORS=1` to ensure the warp accessors code path also remains valid. These could be in the same job, as consequtive steps.

markusbattarbee (changes requested):

.github/workflows/github-ci.yml
ykempf: So as soon as #1100 is merged I guess?

PR #1092, opened 2025-02-05
Periodically overwriting restarts

ykempf:
Allow to write "recover" files (at the moment they are like restart files) at a set tstep interval, and they overwrite themselves periodically according to the number of files set to be used.
```
recover_tstep_interval = 3
number_of_recovers = 2
```

Includes a `touch RECO` to trigger manually a new one.

Does not include duplication of MPI-IO flags, striping etc, those are taken from restart parameters. This is changing soon if e.g. compression is allowed for the recovers, I suppose.

Functionality tested in testpackage in restart_write (but no checks implemented, not needed I suppose as long as they're identical to restarts).

### DIFF
--- a/common.h
+++ b/common.h
@@ -527,8 +527,8 @@ template<typename T> inline int sign(const T& value) {
  */
 struct globalflags {
    static int bailingOut; /*!< Global flag raised to true if a run bailout (write restart if requested/set and stop the simulation peacefully) is needed. */
-   static bool writeRestart; /*!< Global flag raised to true if a restart writing is needed (manually triggered). NOTE: used only by MASTER_RANK in vlasiator.cpp. */
-   static bool writeRecover; /*!< Global flag raised to true if a recover writing is needed (manually triggered). NOTE: used only by MASTER_RANK in vlasiator.cpp. */
+   static bool writeRestart; /*!< Global flag raised to true if a restart writing is needed. NOTE: used only by MASTER_RANK in vlasiator.cpp. */
+   static bool writeRecover; /*!< Global flag raised to true if a recover writing is needed. NOTE: used only by MASTER_RANK in vlasiator.cpp. */
    static bool balanceLoad; /*!< Global flag raised to true if a load balancing is needed. NOTE: used only by MASTER_RANK in vlasiator.cpp. */
    static bool doRefine; /*!< Global flag raised to true if a re-refine is needed. NOTE: used only by MASTER_RANK in vlasiator.cpp. */
    static bool ionosphereJustSolved; /*!< Flag used to notify that the ionosphere has been freshly solved, used to check whether the Vlasov boundary/bulk forcing need updating. */
--- a/common.h
+++ b/common.h
@@ -527,7 +527,8 @@ template<typename T> inline int sign(const T& value) {
  */
 struct globalflags {
    static int bailingOut; /*!< Global flag raised to true if a run bailout (write restart if requested/set and stop the simulation peacefully) is needed. */
-   static bool writeRestart; /*!< Global flag raised to true if a restart writing is needed (without bailout). NOTE: used only by MASTER_RANK in vlasiator.cpp. */
+   static bool writeRestart; /*!< Global flag raised to true if a restart writing is needed (manually triggered). NOTE: used only by MASTER_RANK in vlasiator.cpp. */
+   static bool writeRecover; /*!< Global flag raised to true if a recover writing is needed (manually triggered). NOTE: used only by MASTER_RANK in vlasiator.cpp. */
    static bool balanceLoad; /*!< Global flag raised to true if a load balancing is needed. NOTE: used only by MASTER_RANK in vlasiator.cpp. */
    static bool doRefine; /*!< Global flag raised to true if a re-refine is needed. NOTE: used only by MASTER_RANK in vlasiator.cpp. */
    static bool ionosphereJustSolved; /*!< Flag used to notify that the ionosphere has been freshly solved, used to check whether the Vlasov boundary/bulk forcing need updating. */
--- a/common.h
+++ b/common.h
@@ -534,6 +534,17 @@ struct globalflags {
    static bool ionosphereJustSolved; /*!< Flag used to notify that the ionosphere has been freshly solved, used to check whether the Vlasov boundary/bulk forcing need updating. */
 };
 
+namespace donow {
+   enum {
+      SAVE,
+      DORC,
+      DOLB,
+      DOMR,
+      N_DONOW
+   };
+}
+
+
 /*!
  * Name space for flags going into the project hook function.
  */
--- a/ioread.cpp
+++ b/ioread.cpp
@@ -92,6 +92,17 @@ void checkExternalCommands() {
       rename("SAVE", newName);
       return;
    }
+   if(stat("RECO", &tempStat) == 0) {
+      cerr << "Received an external RECO command. Writing a recover file." << endl;
+      globalflags::writeRecover = true;
+      char newName[80];
+      // Get the current time.
+      const time_t rawTime = time(NULL);
+      const struct tm * timeInfo = localtime(&rawTime);
+      strftime(newName, 80, "RECO_%F_%H-%M-%S", timeInfo);
+      rename("RECO", newName);
+      return;
+   }
    if(stat("DOLB", &tempStat) == 0) {
       cerr << "Received an external DOLB command. Balancing load." << endl;
       globalflags::balanceLoad = true;
--- a/ioread.cpp
+++ b/ioread.cpp
@@ -82,7 +82,7 @@ void checkExternalCommands() {
       return;
    }
    if(stat("SAVE", &tempStat) == 0) {
-      cerr << "Received an external SAVE command. Writing a restart file." << endl;
+      logFile << "Received an external SAVE command. Writing a restart file." << endl;
       globalflags::writeRestart = true;
       char newName[80];
       // Get the current time.
@@ -93,7 +93,7 @@ void checkExternalCommands() {
       return;
    }
    if(stat("DORC", &tempStat) == 0) {
-      cerr << "Received an external RECO command. Writing a recover file." << endl;
+      logFile << "Received an external DORC command. Writing a recover file." << endl;
       globalflags::writeRecover = true;
       char newName[80];
       // Get the current time.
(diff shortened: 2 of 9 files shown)


common.h
```diff
@@ -527,7 +527,8 @@ template<typename T> inline int sign(const T& value) {
  */
 struct globalflags {
    static int bailingOut; /*!< Global flag raised to true if a run bailout (write restart if requested/set and stop the simulation peacefully) is needed. */
-   static bool writeRestart; /*!< Global flag raised to true if a restart writing is needed (without bailout). NOTE: used only by MASTER_RANK in vlasiator.cpp. */
+   static bool writeRestart; /*!< Global flag raised to true if a restart writing is needed (manually triggered). NOTE: used only by MASTER_RANK in vlasiator.cpp. */
```
markusbattarbee: technically, this flag includes both manually triggered and walltime-based restarts, no?

common.h
```diff
@@ -527,7 +527,8 @@ template<typename T> inline int sign(const T& value) {
...
+   static bool writeRecover; /*!< Global flag raised to true if a recover writing is needed (manually triggered). NOTE: used only by MASTER_RANK in vlasiator.cpp. */
```
markusbattarbee: and this one tstep-based?

ioread.cpp
```diff
@@ -92,6 +92,17 @@ void checkExternalCommands() {
       rename("SAVE", newName);
       return;
    }
+   if(stat("RECO", &tempStat) == 0) {
```
markusbattarbee: I vote we call it `DORC` ;)

iowrite.cpp
```diff
@@ -1584,19 +1585,31 @@ bool writeRestart(
...
    }
-   MPI_Bcast(&currentDate,80,MPI_CHAR,MASTER_RANK,MPI_COMM_WORLD);
-   
    // Create a name for the output file and open it with VLSVWriter:
    stringstream fname;
-   fname << P::restartWritePath << "/" << name << ".";
-   fname.width(7);
-   fname.fill('0');
-   fname << fileIndex << "." << currentDate << ".vlsv";
+   if(dateInFileName) {
```
markusbattarbee, line 1601: simplify the logic - if dateinfilename is not set, `currentDate` can be an empty char array. (and if it is set, store also the dot into it)

parameters.cpp
```diff
@@ -264,6 +265,9 @@ bool P::addParameters() {
            "Save the complete simulation in given walltime intervals. Negative values disable writes.", -1.0);
    RP::add("io.number_of_restarts", "Exit the simulation after certain number of walltime-based restarts.",
            numeric_limits<uint>::max());
+   RP::add("io.recover_tstep_interval",
+           "Save the complete simulation in given tstep intervals. 0 disables writes.", 0);
+   RP::add("io.number_of_recovers", "Overwrite recovers cyclically after this number of recovers written.", 0);
```
markusbattarbee: Put default at 2?

common.h
```diff
@@ -527,7 +527,8 @@ template<typename T> inline int sign(const T& value) {
  */
 struct globalflags {
    static int bailingOut; /*!< Global flag raised to true if a run bailout (write restart if requested/set and stop the simulation peacefully) is needed. */
```
markusbattarbee, line 529: Couldn't these all be initialized false here instead of having to do it at the beginning of vlasiator.cpp and the ionosphere mini-app?

vlasiator.cpp
```diff
@@ -942,29 +946,43 @@ int simulate(int argn,char* args[]) {
...
                globalflags::doRefine = false;
             }
          }
       }
-      MPI_Bcast( &doNow, 3 , MPI_INT , MASTER_RANK ,MPI_COMM_WORLD);
+      MPI_Bcast( &doNow, 4 , MPI_INT , MASTER_RANK ,MPI_COMM_WORLD);
       writeRestartNow = doNow[0];
       doNow[0] = 0;
-      if (doNow[1] == 1) {
+      writeRecoverNow = doNow[1];
```
markusbattarbee: can't we just use doNow[0 / 1] further on in the code, with a comment after it, instead of duplicating the variable?

vlasiator.cpp
```diff
@@ -942,29 +946,43 @@ int simulate(int argn,char* args[]) {
...
                globalflags::writeRestart = false; // This flag is only used by MASTER_RANK here and it needs to be reset after a restart write has been issued.
             }
-         }
-         else {
+         } else {
             doNow[0] = 0;
          }
-         if (globalflags::balanceLoad || globalflags::doRefine) {
+         if (  P::saveRecoverTstepInterval > 0
+            && P::recoverFileCount > 0
```
markusbattarbee: I think this naming is not helpful. `P::recoverFileCount` sounds like it's something that gets incremented, whereas this is how many can exist at a time. `p::recoverMaxFiles`?

parameters.cpp
```diff
@@ -573,6 +584,13 @@ void Parameters::getParameters() {
       }
       P::restartWritePath = prefix;
    }
+   if (access(&(P::recoverWritePath[0]), W_OK) != 0) {
+      if (myRank == MASTER_RANK) {
+         cerr << "ERROR recover write path " << P::recoverWritePath << " not writeable, defaulting to local directory."
+              << endl;
+      }
+      P::recoverWritePath = prefix;
+   }
```
markusbattarbee, line 593: This should only trigger if recovers are actually active.

Do we want some sanity checking for the other recover parameters? I guess Boost will complain if someone tries to use negative numbers for those?

markusbattarbee (changes requested):

parameters.cpp
ykempf: Dunno if it will brute-cast it or complain.

ykempf: Actually, we want to check the path in case we `touch "DORC"`, no?

markusbattarbee: Ah, that's a good point.

Will the p::recoverMaxFiles be considered as 1 if it's 0, for DORC purposes?

common.h
ykempf: backgroundfield/../common.h:530:16: error: ISO C++ forbids in-class initialization of non-const static member ‘globalflags::writeRestart’
  530 |    static bool writeRestart=false; /*!< Global flag raised to true if a restart writing is needed. NOTE: used only by MASTER_RANK in vlasiator.cpp. */

parameters.cpp
ykempf: Aaaah, you found a failure mode. If it is set to 0 and a DORC triggers we get a division by zero. Great catch. So this should be checked for being >0 actually.

vlasiator.cpp
ykempf: You prefer
```
if (doNow[1] == 1){ // write recover now
```
below to
```
if (writeRecoverNow == 1){
```
?
Sure can do.

ykempf: Or do you want an enum? :grin:

ykempf:
Comments addressed but a manual check should be done before merging just to be sure I didn't do this too fast now.

ioread.cpp
```diff
@@ -92,6 +92,17 @@ void checkExternalCommands() {
       rename("SAVE", newName);
       return;
    }
+   if(stat("DORC", &tempStat) == 0) {
+      cerr << "Received an external RECO command. Writing a recover file." << endl;
```
markusbattarbee: DORC vs RECO in the output. Also, this should not go to cerr but stdout.

vlasiator.cpp
```diff
@@ -942,35 +944,44 @@ int simulate(int argn,char* args[]) {
                doNow[0] = 2; // Setting to 2 so as to not increment the restart count below.
                globalflags::writeRestart = false; // This flag is only used by MASTER_RANK here and it needs to be reset after a restart write has been issued.
             }
-         }
-         else {
+         } else {
             doNow[0] = 0;
```
markusbattarbee: I would suggest, for clarity:

on line 935, just for master rank, set all elements of `doNow` to zero. 
Remove the `else`-things from 948 and 959.

vlasiator.cpp
```diff
@@ -942,35 +944,44 @@ int simulate(int argn,char* args[]) {
...
       }
-      MPI_Bcast( &doNow, 3 , MPI_INT , MASTER_RANK ,MPI_COMM_WORLD);
-      writeRestartNow = doNow[0];
-      doNow[0] = 0;
-      if (doNow[1] == 1) {
+      MPI_Bcast( &doNow, 4 , MPI_INT , MASTER_RANK ,MPI_COMM_WORLD);
+      if (doNow[2] == 1) {
          P::prepareForRebalance = true;
-         doNow[1] = 0;
+         doNow[2] = 0;
```
markusbattarbee: then, remove resetting `doNow` flags here in the if-clauses. (also there's a mix of `0` and `false`...)

Instead, at the end of these checks at 1060 or so, set all flags again to zero.

markusbattarbee (commented):
Suggestion for logic clarification (plus an actual change request)

vlasiator.cpp
ykempf: *Someone* decided to add the 4th as a bool instead and is still passing it in the MPI broadcast as an MPI_INT. I disapprove but that was merged a while back.

vlasiator.cpp
```diff
@@ -791,10 +793,10 @@ int simulate(int argn,char* args[]) {
    // Invalidate cached cell lists just to be sure (might not be needed)
    P::meshRepartitioned = true;
 
-   unsigned int wallTimeRestartCounter=1;
+   uint wallTimeRestartCounter=1;
+   uint recoverCounter=0;
 
-   int doNow[3] = {0}; // 0: writeRestartNow, 1: balanceLoadNow, 2: refineNow ; declared outside main loop
-   int writeRestartNow; // declared outside main loop
+   int doNow[4] = {0}; // 0: writeRestartNow, 1: writeRecoverNow, 2: balanceLoadNow, 3: refineNow ; declared outside main loop
```
ursg: For code readability, I would actually suggest to have a struct
```c++
struct DoNow {
   bool writeRestart;
   bool writeRecover;
   bool balanceLoad;
   bool refine;
} doNow;
```
instead. Or at least make the array entries preprocessor macros. Or an enum.

Because the hardcoded array index numbers in the actual logic code further down are sure to confuse the uninitiated.

ursg (changes requested):
Functional approval, but that doNow array is starting to be big enough, that some sort of symbolic names would make sense there.

markusbattarbee (approved):

ursg (approved):

PR #1094, opened 2025-02-10
Reorder memory report lines in logfile.

ykempf:
I find it easier to have free and resident next to each other instead of HWM in between.

### DIFF
--- a/memory_report.cpp
+++ b/memory_report.cpp
@@ -24,7 +24,6 @@
 #include <cstdlib>
 #include <iostream>
 #include <sstream>
-#include <iomanip>
 #include <format>
 
 #ifdef _OPENMP
--- a/memory_report.cpp
+++ b/memory_report.cpp
@@ -120,7 +120,6 @@ void report_memory_consumption(
    int rank, nProcs, nodeRank, interRank;
    int nNodes;
    const double GiB = pow(2,30);
-   const double TiB = pow(2,40);
 
    std::hash<std::string> hasher;
    MPI_Comm nodeComm;
@@ -143,7 +142,7 @@ void report_memory_consumption(
    MPI_Comm_size(interComm, &nNodes);
 
    // string formatting
-   constexpr std::string_view formatString = "(MEM) tstep {} t {:.3g} {:<21} (GiB/node; avg, min, max): {:<7.5g} {:<7.5g} {:<7.5g} sum (TiB) {:<7.5g} on {} nodes\n";
+   constexpr std::string_view formatString = "(MEM) tstep {} t {:.3g} {:<21} (GiB/node; avg, min, max, sum): {:<8.3g} {:<8.3g} {:<8.3g} {:<8.3g} on {} nodes\n";
 
    // Report /proc/meminfo memory consumption first so we then get resident and HWM from Papi below, easier to compare by eye in logfile in this order.
    double mem_proc_free = (double)get_node_free_memory();
@@ -153,7 +152,7 @@ void report_memory_consumption(
    MPI_Reduce( &mem_proc_free, &total_mem_proc, numberOfParameters, MPI_DOUBLE, MPI_SUM, MASTER_RANK, MPI_COMM_WORLD );
    MPI_Reduce( &mem_proc_free, &min_free, numberOfParameters, MPI_DOUBLE, MPI_MIN, MASTER_RANK, MPI_COMM_WORLD );
    MPI_Reduce( &mem_proc_free, &max_free, numberOfParameters, MPI_DOUBLE, MPI_MAX, MASTER_RANK, MPI_COMM_WORLD );
-   logFile << std::format(formatString, P::tstep, P::t, "Free", total_mem_proc/nProcs/GiB, min_free/GiB, max_free/GiB, total_mem_proc/TiB, nNodes);
+   logFile << std::format(formatString, P::tstep, P::t, "Free", total_mem_proc/nProcs/GiB, min_free/GiB, max_free/GiB, total_mem_proc/GiB, nNodes);
 
 #ifdef PAPI_MEM
    /*If we have PAPI, we can report the resident usage of the process*/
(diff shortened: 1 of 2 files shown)


markusbattarbee (commented):
If you're editing the eye candy, how about aligning numbers in the output? :) 😇

ykempf:
So this introduces a "new" requirement that boost be not too old (1.55 is too old, 1.80s versions at least work), however since `boost/format` is header only it doesn't require anything but a reinstallation of a recent-enough version with `program_options`, no other tweaks needed.

It now compiles on Carrington after I updated there, mahti_gcc is fine, lumi_gnu_2403 as well. Do let me know if I should check other machines as part of this PR.

markusbattarbee (approved):

memory_report.cpp
```diff
@@ -33,6 +34,8 @@
 
 #include "memory_report.h"
 
+#include <boost/format.hpp>
```
lkotipal: As per Zulip, we can use `std::format` instead

memory_report.cpp
```diff
@@ -140,6 +143,29 @@ void report_memory_consumption(
...
+
+   // Report /proc/meminfo memory consumption first so we then get resident and HWM from Papi below, easier to compare by eye in logfile in this order.
+   double mem_proc_free = (double)get_node_free_memory();
+   double total_mem_proc = 0;
+   double min_free,max_free;
+   const int numberOfParameters = 1;
+   MPI_Reduce( &mem_proc_free, &total_mem_proc, numberOfParameters, MPI_DOUBLE, MPI_SUM, MASTER_RANK, MPI_COMM_WORLD );
+   MPI_Reduce( &mem_proc_free, &min_free, numberOfParameters, MPI_DOUBLE, MPI_MIN, MASTER_RANK, MPI_COMM_WORLD );
+   MPI_Reduce( &mem_proc_free, &max_free, numberOfParameters, MPI_DOUBLE, MPI_MAX, MASTER_RANK, MPI_COMM_WORLD );
+   logFile << "(MEM) tstep " << Parameters::tstep << " t " << Parameters::t << " Free             " << aligned_output_formatter % (total_mem_proc/nProcs / GiB) % (min_free / GiB) % (max_free / GiB) % (total_mem_proc/TiB) % nNodes;
```
lkotipal: Consider placing everything into the formatstring, i.e. (w/ `std::format` syntax) 
```c++
std::string formatString "(MEM) tstep {} t {} {<16} (GiB/node; avg, min, max): {<7.5g} {<7.5g} {<7.5g} sum (TiB) {<7.5g} on {} nodes\n");
logFile << std::format(formatString, Parameters::tstep, parameters::t, "Free", (total_mem_proc/nProcs / GiB), (min_free / GiB), (max_free / GiB), (total_mem_proc/TiB), nNodes)
```

lkotipal: Can also add a TODO for when we eventually get C++23 and `std::print`, which would just make it:
```
std::print(formatString, /* ... */)
```
Or `std::println` if we drop the newline from the formatstring.

spatial_cells/velocity_block_container.h
```diff
@@ -749,18 +749,18 @@ namespace vmesh {
...
-      return block_data.size()*WID3*sizeof(Realf) + parameters.size()*sizeof(Real);
+      return block_data.size()*sizeof(Realf) + parameters.size()*sizeof(Real);
       #else
       #ifdef DEBUG_VBC
       const size_t currentSize = block_data.size() / WID3;
       if (currentSize != cachedSize) {
          printf("VBC CHECK ERROR: cached size mismatch, %lu vs %lu in %s : %d\n",currentSize,cachedSize,__FILE__,__LINE__);
       }
       #endif
-      return cachedSize*WID3*sizeof(Realf) + cachedSize*BlockParams::N_VELOCITY_BLOCK_PARAMS*sizeof(Real);
```
markusbattarbee, line 760: Commented on this on the actual commit but that might have got lost - on this line WID3 is needed. 

`vector.size()` returns the actual count of elements inside the vector, whereas `cachedSize` is how many blocks are in there - each of size `WID3`.

memory_report.cpp
```diff
@@ -166,32 +189,29 @@ void report_memory_consumption(
          if (max_mem_papi[3] != 0.0) {
             logFile << "(MEM) Estimating increased high water mark from refinement" << std::endl;
          }
+#if __GNUC__ >= 13
+         logFile << std::format(formatString, P::tstep, P::t, "Resident",            sum_mem_papi[2]/nNodes/GiB, min_mem_papi[2]/GiB, max_mem_papi[2]/GiB, sum_mem_papi[2]/GiB, nNodes);
+         logFile << std::format(formatString, P::tstep, P::t, "High water mark \U0001F30A   ",     sum_mem_papi[0]/nNodes/GiB, min_mem_papi[0]/GiB, max_mem_papi[0]/GiB, sum_mem_papi[0]/GiB, nNodes);
```
markusbattarbee, line 194: trying to sneak in an emoji to break everything before skedaddling? 🤔

lkotipal: Drop the extra spaces too, might mess up padding

ykempf: I added these so it actually aligns in the terminal output of `less logfile.txt` (with a square instead of the water). If this is also unwanted, I just drop that attempt and stick to them in commit message. :innocent:

memory_report.cpp
```diff
@@ -24,6 +24,10 @@
 #include <cstdlib>
 #include <iostream>
 #include <sstream>
+#if __GNUC__ >= 13
+   #include <format>
+#endif
+
```
ursg, line 30: Ugh, I don't like this at all. Special treatment like this, because some platforms that we actually still run on can't handle the new code? When there is a perfectly working ```snprintif``` available, that has worked in every c compiler ever since 1969?

memory_report.cpp
```diff
@@ -140,6 +143,26 @@ void report_memory_consumption(
...
+   const int numberOfParameters = 1;
+   MPI_Reduce( &mem_proc_free, &total_mem_proc, numberOfParameters, MPI_DOUBLE, MPI_SUM, MASTER_RANK, MPI_COMM_WORLD );
+   MPI_Reduce( &mem_proc_free, &min_free, numberOfParameters, MPI_DOUBLE, MPI_MIN, MASTER_RANK, MPI_COMM_WORLD );
+   MPI_Reduce( &mem_proc_free, &max_free, numberOfParameters, MPI_DOUBLE, MPI_MAX, MASTER_RANK, MPI_COMM_WORLD );
+#if __GNUC__ >= 13
+   logFile << std::format(formatString, P::tstep, P::t, "Free", total_mem_proc/nProcs/GiB, min_free/GiB, max_free/GiB, total_mem_proc/GiB, nNodes);
+#else
+   logFile << "(MEM) tstep " << P::tstep << " t " << P::t << " Free             (GiB/node; avg, min, max): " << total_mem_proc/nProcs / GiB << " " << min_free / GiB << " " << max_free / GiB <<
+      " sum (GiB): " << total_mem_proc/GiB << " on "<< nNodes << " nodes" << std::endl;
+#endif
```
ursg, line 164: Why not just have the non-gcc-13 case feature-identical?

memory_report.cpp
```diff
@@ -140,6 +143,26 @@ void report_memory_consumption(
    MPI_Comm_rank(interComm, &interRank);
    MPI_Comm_size(interComm, &nNodes);
 
+#if __GNUC__ >= 13
+   // string formatting
+   constexpr std::string_view formatString = "(MEM) tstep {} t {:.3g} {:<21} (GiB/node; avg, min, max, sum): {:<8.3g} {:<8.3g} {:<8.3g} {:<8.3g} on {} nodes\n";
```
ursg, line 148: In particular, this format string uses none of the proper fancy std::format features, and is just a reformulation of a standard printf format in python style.

memory_report.cpp
ursg: My sense of trying to come up with a pathological example of where this might breaks comes up empty. So I would say this is fully ok!

memory_report.cpp
```diff
@@ -140,6 +143,26 @@ void report_memory_consumption(
...
+#endif
+
+   // Report /proc/meminfo memory consumption first so we then get resident and HWM from Papi below, easier to compare by eye in logfile in this order.
+   double mem_proc_free = (double)get_node_free_memory();
+   double total_mem_proc = 0;
+   double min_free,max_free;
+   const int numberOfParameters = 1;
+   MPI_Reduce( &mem_proc_free, &total_mem_proc, numberOfParameters, MPI_DOUBLE, MPI_SUM, MASTER_RANK, MPI_COMM_WORLD );
```
ursg, line 156: This is wrong. If running with n mpi tasks per node, it sums the same node n times.
(Just tested on 1 node with 4 tasks. With 8 GB free, it reports ```min 8, max 8, sum 32```).

I guess this needs to be performed on the interComm?

ursg: Also, the avg calculation then needs to divide by nNodes, not nProcs.

ursg (commented):
I don't like the introduction of std::format at all.
IMHO, std::format is unnecessarily verbose, badly documented, badly supported by compilers and brings no feature benefits over printf format strings.

In fact, I like it so little that I'll immediately send a PR to revert it, and replace it with trusty old printf strings.

ursg:
See #1101 for my suggested fixes.

PR #1095, opened 2025-02-10
Add support for multiple boxes where refinement is allowed

JonasSuni:
(Gaze upon my awful boolean logic and variable names and despair!)

This PR adds support for defining multiple coordinate axis-align boxes where refinement is allowed.

### DIFF
--- a/parameters.cpp
+++ b/parameters.cpp
@@ -493,7 +493,7 @@ bool P::addParameters() {
    RP::add("AMR.refine_cadence","Refine every nth load balance", 5);
    RP::add("AMR.refine_after","Start refinement after this many simulation seconds", 0.0);
    RP::add("AMR.refine_radius","Maximum distance from origin to allow refinement within. Only induced refinement allowed outside this radius.", LARGE_REAL);
-   RP::add("AMR.number_of_refine_boxes", "How many boxes in which to allow refinement, that number of box edges have to then be defined as well.", 0);
+   RP::add("AMR.number_of_refine_boxes", "How many boxes outside which to suppress refinement, that number of box edges have to then be defined as well. If more than 1 box is defined, refinement is suppressed outside the union of the volumes of all boxes.", 0);
    RP::addComposing("AMR.refinement_min_x", "Refinement minimum X coordinate, no refinement at x < this value (m) except induced refinement.");
    RP::addComposing("AMR.refinement_min_y", "Refinement minimum Y coordinate, no refinement at y < this value (m) except induced refinement.");
    RP::addComposing("AMR.refinement_min_z", "Refinement minimum Z coordinate, no refinement at z < this value (m) except induced refinement.");
--- a/parameters.cpp
+++ b/parameters.cpp
@@ -183,12 +183,13 @@ Real P::alphaDBWeight = 1.0;
 uint P::refineCadence = 5;
 Real P::refineAfter = 0.0;
 Real P::refineRadius = LARGE_REAL;
-Real P::refinementMinX = -LARGE_REAL;
-Real P::refinementMinY = -LARGE_REAL;
-Real P::refinementMinZ = -LARGE_REAL;
-Real P::refinementMaxX = LARGE_REAL;
-Real P::refinementMaxY = LARGE_REAL;
-Real P::refinementMaxZ = LARGE_REAL;
+int P::refineBoxNumber = 0;
+std::vector<Real> P::refinementMinX;
+std::vector<Real> P::refinementMinY;
+std::vector<Real> P::refinementMinZ;
+std::vector<Real> P::refinementMaxX;
+std::vector<Real> P::refinementMaxY;
+std::vector<Real> P::refinementMaxZ;
 int P::maxFilteringPasses = 0;
 int P::amrBoxNumber = 0;
 std::vector<uint> P::amrBoxHalfWidthX;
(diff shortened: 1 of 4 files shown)


JonasSuni:
Whoops, I should probably check that it compiles first

JonasSuni:
It does compile

projects/project.cpp
```diff
@@ -456,15 +456,22 @@ namespace projects {
...
+            || (xyz[2] < P::refinementMinZ.at(i)) || (xyz[2] > P::refinementMaxZ.at(i))) {
+            shouldNotRefine = shouldNotRefine && true;
+         } else {
+            shouldNotRefine = shouldNotRefine && false;
+         }
       }
 
+      shouldRefine = shouldRefine && !shouldNotRefine;
+
       return shouldRefine;
```
ykempf, line 474: So it will only refine if it is inside the intersection of all defined refinement boxes (and some criterion tells it to refine).

projects/project.cpp
```diff
@@ -493,16 +500,23 @@ namespace projects {
...
+            || (xyz[2] < P::refinementMinZ.at(i)) || (xyz[2] > P::refinementMaxZ.at(i))) {
+            shouldNotUnrefine = shouldNotUnrefine || true;
+         } else {
+            shouldNotUnrefine = shouldNotUnrefine || false;
+         }
       }
 
+      shouldUnrefine = shouldUnrefine || !shouldNotUnrefine;
+
       return shouldUnrefine;
```
ykempf, line 518: And this new mechanism says it shouldNotUnrefine if outside any of the defined boxes. Which confuses me. Why shouldn't it coarsen? Especially in your operational case at hand, you'll want things outside of the box to be coarsened, no?

ykempf (commented):
You got me confused, I think.

projects/project.cpp
```diff
@@ -456,15 +456,22 @@ namespace projects {
...
 
-      if(
-         // If this cell is planned to be refined, but is outside the allowed refinement region, cancel that refinement.
-         // Induced refinement still possible just beyond that limit.
-         (xyz[0] < P::refinementMinX) || (xyz[0] > P::refinementMaxX)
-         || (xyz[1] < P::refinementMinY) || (xyz[1] > P::refinementMaxY)
-         || (xyz[2] < P::refinementMinZ) || (xyz[2] > P::refinementMaxZ)) {
-         shouldRefine = false;
+      bool shouldNotRefine = true;
+      for (uint i = 0; i < P::refineBoxNumber; i++) {
```
ykempf: I'll side with the compiler warnings here. :grin:

projects/project.cpp
JonasSuni: No, shouldNotRefine will be false unless the point is outside all of the boxes. At least I think that is the case

projects/project.cpp
lkotipal: Specifically, change this to int, not the one in parameters to uint

JonasSuni: Yep, will do

projects/project.cpp
ykempf: All right, this was again a case of needing the explanation. OK with the upcoming changes.

projects/project.cpp
ykempf: Same here, we reviewed this together, I am less confused and the upcoming changes will make things more betterly clearer.

ykempf:
@markusbattarbee do we want to include a box in our legendary dAMR test?

markusbattarbee:
Will require changing the test and generating new reference data, but yes, a good idea.

markusbattarbee:
So on the first go, the Flowthrough_damr config needs to be updated to actually run. Then, when we get clean TP diffs, we can introduce the use of two boxes.

JonasSuni:
I updated the flowthrough damr test to use two boxes now, but of course that will have diffs compared to the reference data.

Should change it back to one box but with the parameter that makes it run properly?

markusbattarbee (approved):

parameters.cpp
```diff
@@ -492,12 +493,13 @@ bool P::addParameters() {
...
-   RP::add("AMR.refinement_min_x", "Refinement minimum X coordinate, no refinement at x < this value (m) except induced refinement.", -LARGE_REAL);
-   RP::add("AMR.refinement_min_y", "Refinement minimum Y coordinate, no refinement at y < this value (m) except induced refinement.", -LARGE_REAL);
-   RP::add("AMR.refinement_min_z", "Refinement minimum Z coordinate, no refinement at z < this value (m) except induced refinement.", -LARGE_REAL);
-   RP::add("AMR.refinement_max_x", "Refinement maximum X coordinate, no refinement at x > this value (m) except induced refinement.", LARGE_REAL);
-   RP::add("AMR.refinement_max_y", "Refinement maximum Y coordinate, no refinement at y > this value (m) except induced refinement.", LARGE_REAL);
-   RP::add("AMR.refinement_max_z", "Refinement maximum Z coordinate, no refinement at z > this value (m) except induced refinement.", LARGE_REAL);
+   RP::add("AMR.number_of_refine_boxes", "How many boxes in which to allow refinement, that number of box edges have to then be defined as well.", 0);
```
ykempf: Hullo there! Maybe explain that if there's more than 1 boxes defined, refinement is allowed inside the union of their volumes/suppressed outside? And *technically* you're not defining the "edges", but don't waste a precious commit just on that. ;)

JonasSuni: Good catch, I'll make that more clear

ykempf: Still not edges, but fine.

PR #1096, opened 2025-02-11
Bin pencils

lkotipal:
Attempt binning pencils to enable more threaded translation without race conditions

### DIFF
--- a/vlasovsolver/cpu_trans_map_amr.cpp
+++ b/vlasovsolver/cpu_trans_map_amr.cpp
@@ -335,12 +335,9 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
    int propagateTimerId = phiprof::initializeTimer("trans-amr-propagatePencil");
 
    std::vector<uint> activeBins;
-   for (auto& [bin, cells] : DimensionPencils[dimension].binsPencils) {
-      activeBins.push_back(bin);
-   }
 
    const size_t blocksSize {unionOfBlocks.size()};
-   const size_t binsSize {activeBins.size()};
+   const size_t binsSize {DimensionPencils[dimension].activeBins.size()};
 
    #pragma omp parallel
    {
@@ -361,7 +358,7 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
 
             phiprof::Timer loadTimer {loadTimerId};
             vmesh::GlobalID blockGID = unionOfBlocks[blocki];
-            for (uint pencili : DimensionPencils[dimension].binsPencils[activeBins[nBin]]) {
+            for (uint pencili : DimensionPencils[dimension].binsPencils[DimensionPencils[dimension].activeBins[nBin]]) {
                int nonEmptyBlocks = 0;
                int L = DimensionPencils[dimension].lengthOfPencils[pencili];
                int start = DimensionPencils[dimension].idsStart[pencili];
--- a/vlasovsolver/cpu_trans_map_amr.cpp
+++ b/vlasovsolver/cpu_trans_map_amr.cpp
@@ -357,7 +357,7 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
 
             phiprof::Timer loadTimer {loadTimerId};
             vmesh::GlobalID blockGID = unionOfBlocks[blocki];
-            for (uint pencili : DimensionPencils[dimension].binsPencils[currentBin]) {
+            for (uint pencili : DimensionPencils[dimension].pencilsInBin[currentBin]) {
                int nonEmptyBlocks = 0;
                int L = DimensionPencils[dimension].lengthOfPencils[pencili];
                int start = DimensionPencils[dimension].idsStart[pencili];
@@ -388,7 +388,7 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
 
             phiprof::Timer memsetTimer {memsetTimerId};
             // reset blocks in all non-sysboundary neighbor spatial cells for this block id
-            for (CellID target_cell_id: DimensionPencils[dimension].binsCells[currentBin]) {
+            for (CellID target_cell_id: DimensionPencils[dimension].targetCellsInBin[currentBin]) {
                SpatialCell* target_cell = mpiGrid[target_cell_id];
                if (target_cell) {
                   // Get local velocity block id
@@ -404,7 +404,7 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
             memsetTimer.stop();
 
             phiprof::Timer propagateTimer {propagateTimerId};
-            for (uint pencili : DimensionPencils[dimension].binsPencils[currentBin]) {
+            for (uint pencili : DimensionPencils[dimension].pencilsInBin[currentBin]) {
                // Skip pencils without blocks
                if (pencilBlocksCount.at(pencili) == 0) {
                   continue;
--- a/vlasovsolver/cpu_trans_map_amr.cpp
+++ b/vlasovsolver/cpu_trans_map_amr.cpp
@@ -393,17 +393,15 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
             phiprof::Timer memsetTimer {memsetTimerId};
             // reset blocks in all non-sysboundary neighbor spatial cells for this block id
             for (CellID target_cell_id: DimensionPencils[dimension].binsCells[activeBins[nBin]]) {
-               if (DimensionTargetCells[dimension].contains(target_cell_id)) {
-                  SpatialCell* target_cell = mpiGrid[target_cell_id];
-                  if (target_cell) {
-                     // Get local velocity block id
-                     const vmesh::LocalID blockLID = target_cell->get_velocity_block_local_id(blockGID, popID);
-                     // Check for invalid block id
-                     if (blockLID != vmesh::VelocityMesh::invalidLocalID()) {
-                        // Get a pointer to the block data
-                        Realf* blockData = target_cell->get_data(blockLID, popID);
-                        memset(blockData, 0, WID3*sizeof(Realf));
-                     }
+               SpatialCell* target_cell = mpiGrid[target_cell_id];
+               if (target_cell) {
+                  // Get local velocity block id
+                  const vmesh::LocalID blockLID = target_cell->get_velocity_block_local_id(blockGID, popID);
+                  // Check for invalid block id
+                  if (blockLID != vmesh::VelocityMesh::invalidLocalID()) {
+                     // Get a pointer to the block data
+                     Realf* blockData = target_cell->get_data(blockLID, popID);
+                     memset(blockData, 0, WID3*sizeof(Realf));
                   }
                }
             }
--- a/vlasovsolver/cpu_trans_map_amr.cpp
+++ b/vlasovsolver/cpu_trans_map_amr.cpp
@@ -339,6 +339,9 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
       activeBins.push_back(bin);
    }
 
+   const size_t blocksSize {unionOfBlocks.size()};
+   const size_t binsSize {activeBins.size()};
+
    #pragma omp parallel
    {
       phiprof::Timer mappingTimer {mappingTimerId}; // mapping (top-level)
@@ -352,8 +355,8 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
       // Get global id of the velocity block
       // Load data for pencils.
       #pragma omp for schedule(dynamic,1) collapse(2)
-      for(uint blocki = 0; blocki < unionOfBlocks.size(); blocki++) {
-         for (uint nBin = 0; nBin < activeBins.size(); ++nBin) {
+      for(uint blocki = 0; blocki < blocksSize; blocki++) {
+         for (uint nBin = 0; nBin < binsSize; ++nBin) {
             // For each block + bin we copy first copy each pencil's data into a buffer, clear the target blocks, and then sum the translated pencils in
 
             phiprof::Timer loadTimer {loadTimerId};
@@ -435,9 +438,9 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
                               pencilRatios,
                               vcell_transpose
                   );
-            }
-         }
-      }
+            } // Loop over pencils
+         } // Loop over bins
+      } // Loop over blocks
 
    } // closes pragma omp parallel
 
--- a/vlasovsolver/cpu_trans_map_amr.cpp
+++ b/vlasovsolver/cpu_trans_map_amr.cpp
@@ -353,10 +353,11 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
       for(uint blocki = 0; blocki < blocksSize; blocki++) {
          for (uint nBin = 0; nBin < binsSize; ++nBin) {
             // For each block + bin we copy first copy each pencil's data into a buffer, clear the target blocks, and then sum the translated pencils in
+            uint currentBin = DimensionPencils[dimension].activeBins[nBin];
 
             phiprof::Timer loadTimer {loadTimerId};
             vmesh::GlobalID blockGID = unionOfBlocks[blocki];
-            for (uint pencili : DimensionPencils[dimension].binsPencils[DimensionPencils[dimension].activeBins[nBin]]) {
+            for (uint pencili : DimensionPencils[dimension].binsPencils[currentBin]) {
                int nonEmptyBlocks = 0;
                int L = DimensionPencils[dimension].lengthOfPencils[pencili];
                int start = DimensionPencils[dimension].idsStart[pencili];
@@ -387,7 +388,7 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
 
             phiprof::Timer memsetTimer {memsetTimerId};
             // reset blocks in all non-sysboundary neighbor spatial cells for this block id
-            for (CellID target_cell_id: DimensionPencils[dimension].binsCells[DimensionPencils[dimension].activeBins[nBin]]) {
+            for (CellID target_cell_id: DimensionPencils[dimension].binsCells[currentBin]) {
                SpatialCell* target_cell = mpiGrid[target_cell_id];
                if (target_cell) {
                   // Get local velocity block id
@@ -403,7 +404,7 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
             memsetTimer.stop();
 
             phiprof::Timer propagateTimer {propagateTimerId};
-            for (uint pencili : DimensionPencils[dimension].binsPencils[DimensionPencils[dimension].activeBins[nBin]]) {
+            for (uint pencili : DimensionPencils[dimension].binsPencils[currentBin]) {
                // Skip pencils without blocks
                if (pencilBlocksCount.at(pencili) == 0) {
                   continue;
--- a/vlasovsolver/cpu_trans_map_amr.cpp
+++ b/vlasovsolver/cpu_trans_map_amr.cpp
@@ -335,7 +335,7 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
    int propagateTimerId = phiprof::initializeTimer("trans-amr-propagatePencil");
 
    std::vector<uint> activeBins;
-   for (auto& [bin, cells] : DimensionPencils[dimension].binCells) {
+   for (auto& [bin, cells] : DimensionPencils[dimension].binsPencils) {
       activeBins.push_back(bin);
    }
 
@@ -356,10 +356,7 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
          for (uint nBin = 0; nBin < activeBins.size(); ++nBin) {
             phiprof::Timer loadTimer {loadTimerId};
             vmesh::GlobalID blockGID = unionOfBlocks[blocki];
-            for (uint pencili = 0; pencili < DimensionPencils[dimension].N; ++pencili) {
-               if (DimensionPencils[dimension].bins[pencili] != activeBins[nBin])
-                  continue;
-
+            for (uint pencili : DimensionPencils[dimension].binsPencils[activeBins[nBin]]) {
                int nonEmptyBlocks = 0;
                int L = DimensionPencils[dimension].lengthOfPencils[pencili];
                int start = DimensionPencils[dimension].idsStart[pencili];
@@ -390,7 +387,7 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
 
             phiprof::Timer memsetTimer {memsetTimerId};
             // reset blocks in all non-sysboundary neighbor spatial cells for this block id
-            for (CellID target_cell_id: DimensionPencils[dimension].binCells[activeBins[nBin]]) {
+            for (CellID target_cell_id: DimensionPencils[dimension].binsCells[activeBins[nBin]]) {
                if (DimensionTargetCells[dimension].contains(target_cell_id)) {
                   SpatialCell* target_cell = mpiGrid[target_cell_id];
                   if (target_cell) {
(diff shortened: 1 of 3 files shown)


lkotipal:
algorithms-Spatial-space-time algorithms-Spatial-space-cells

Performance results from 20s of `samples/Magnetosphere3D`:
```
control
386.83680216471356 +- 0.6757976138800245
control cells/second compared to control: 1.0
control / control: KstestResult(statistic=0.0, pvalue=1.0, statistic_location=384343.7957763672, statistic_sign=1)

binned
403.1496047973633 +- 1.1022723703182713
binned cells/second compared to control: 1.0421697277543511
control / binned: KstestResult(statistic=1.0, pvalue=0.0021645021645021645, statistic_location=389477.9968261719, statistic_sign=1)
```

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -334,105 +334,116 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
...
-               );
          }
-         propagateTimer.stop();
-         mappingTimer.stop(); // mapping (top-level)
-      } // Closes loop over blocks
+      }
+
    } // closes pragma omp parallel
 
+   // TODO Where the hell does this belong?
```
ykempf: Indeed.

ykempf (commented):
So this also has a truncated ionsosphere report (fixed elsewhere) and a non-matching grid that should be flagged RED!

Any way this can help in sorting translation summation at AMR X task interfaces to get better reproducibility/less diffs in testpackage flags?

lkotipal:
Not sure. The order of pencils is deterministic, no?

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -97,6 +104,70 @@ struct setOfPencils {
...
    }
 
+   void binPencils() {
+      bins.resize(N);
+
+      for (uint i = 0; i < N; ++i) {
+         std::vector<uint> matchingBins;
+         for (auto& [bin, cells] : binsCells) {
+            for (auto id = ids.begin() + idsStart[i]; id < ids.begin() + idsStart[i] + lengthOfPencils[i]; ++id) {
+               if (*id && binsCells[i].contains(*id)) {
```
markusbattarbee: If I understand correctly, instead of `binCells[i].contains(*id)` it could just as well be `cells.contains(*id)`?

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -97,6 +104,70 @@ struct setOfPencils {
...
+
+         bins[i] = matchingBins[0];
+
+         for (auto id = ids.begin() + idsStart[i]; id < ids.begin() + idsStart[i] + lengthOfPencils[i]; ++id) {
+            if (*id) {
+               binsCells[matchingBins[0]].insert(*id);
+            }
+         }
+
+      }
```
markusbattarbee, line 138: `// end loop over pencils`

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -97,6 +104,70 @@ struct setOfPencils {
...
 
+   void binPencils() {
+      bins.resize(N);
+
+      for (uint i = 0; i < N; ++i) {
+         std::vector<uint> matchingBins;
+         for (auto& [bin, cells] : binsCells) {
+            for (auto id = ids.begin() + idsStart[i]; id < ids.begin() + idsStart[i] + lengthOfPencils[i]; ++id) {
+               if (*id && binsCells[i].contains(*id)) {
+                  matchingBins.push_back(bin);
```
markusbattarbee: And since in the rest of this pencil loop we only use `matchingBins[0]`, there could be a double-continue here? (and thus would need some refactoring)

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -97,6 +104,70 @@ struct setOfPencils {
...
+            auto& cells1 = bin1->second;
+            for (auto bin2 = binsCells.begin(); bin2 != binsCells.end(); ++bin2) {
+               if (bin1 == bin2 || binsToDelete.contains(bin2->first))
+                  continue;
+
+               for (auto cell : bin2->second) {
+                  if (cells1.contains(cell)) {
+                     binsToDelete.insert(bin2->first);
+                     cells1.insert(bin2->second.begin(), bin2->second.end());
+                     std::replace(bins.begin(), bins.end(), bin2->first, bin1->first);
```
markusbattarbee: This command could use a comment, took me a moment to figure out what's going on. 
Also, can add a `break;` at the end of this `if`-section.

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -97,6 +104,70 @@ struct setOfPencils {
...
+      }
+
+      // Super ugly!
+      // TODO If anyone knows how to do this more efficiently feel free to fix it
+      std::set<uint> binsToDelete;
+      do {
+         binsToDelete.clear();
+         for (auto bin1 = binsCells.begin(); bin1 != binsCells.end(); ++bin1) {
+            if (binsToDelete.contains(bin1->first))
+               continue;
```
markusbattarbee, line 147: curly braces for `if` here and also on line 147

vlasovsolver/cpu_trans_pencils.cpp
```diff
@@ -1273,8 +1273,51 @@ void printPencilsFunc(const setOfPencils& pencils, const uint dimension, const i
...
+      ss << "\n";
+   }
+
+   for (const auto& [bin, binsPencils] : pencils.binsPencils) {
+      const auto& cells = pencils.binsCells.at(bin);
+      std::set<uint64_t> collisions;
+
+      ss << "Bin " << bin << " pencils: ";
+      if (binsPencils.empty()) {
+         ss << "EMPTY ";
```
markusbattarbee, line 1284: Hopefully this won't ever trigger.

vlasovsolver/cpu_trans_pencils.cpp
```diff
@@ -1273,8 +1273,51 @@ void printPencilsFunc(const setOfPencils& pencils, const uint dimension, const i
...
+
+      for (auto pencil : binsPencils) {
+         ss << pencil << " ";
+      }
+
+      ss << "\n";
+
+      ss << "Bin " << bin << " cells: ";
+      if (cells.empty()) {
+         ss << "EMPTY ";
```
markusbattarbee, line 1295: Hopefully this won't ever trigger either.

vlasovsolver/cpu_trans_pencils.cpp
```diff
@@ -1273,8 +1273,51 @@ void printPencilsFunc(const setOfPencils& pencils, const uint dimension, const i
...
+      }
+
+      for (auto id : cells) {
+         ss << id << " ";
+         for (auto [bin2, cells2] : pencils.binsCells) {
+            if (bin != bin2 && cells2.contains(id)) {
+               collisions.insert(bin2);
```
markusbattarbee, line 1302: Perhaps collisions could store the bin and the cell and print both if any exist.

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -334,103 +334,111 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
    int memsetTimerId = phiprof::initializeTimer("trans-amr-MemSet");
    int propagateTimerId = phiprof::initializeTimer("trans-amr-propagatePencil");
 
-#pragma omp parallel
+   std::vector<uint> activeBins;
+   for (auto& [bin, cells] : DimensionPencils[dimension].binsPencils) {
```
markusbattarbee: This could be at the end of the `binPencils` function.

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -334,103 +334,111 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
...
       std::vector<Vec> blockDataBuffer(DimensionPencils[dimension].sumOfLengths*WID3/VECL);
       std::vector<Realf*> cellBlockData(DimensionPencils[dimension].sumOfLengths);
       std::vector<uint> pencilBlocksCount(DimensionPencils[dimension].N);
 
       // Loop over velocity space blocks (threaded).
-#pragma omp for schedule(dynamic,1)
+      // Get global id of the velocity block
+      // Load data for pencils.
+      #pragma omp for schedule(dynamic,1) collapse(2)
       for(uint blocki = 0; blocki < unionOfBlocks.size(); blocki++) {
```
markusbattarbee: Actually, I would recommend storing `unionOfBlocks.size()` and `activeBins.size()` in constant variables before this parallel region (or, placing `nActiveBins` in the pencils struct), so that the loops don't need to re-evaluate the container sizes every iteration.

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -334,103 +334,111 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
...
-               // Check for invalid block id
-               if (blockLID != vmesh::VelocityMesh::invalidLocalID()) {
-                  // Get a pointer to the block data
-                  Realf* blockData = target_cell->get_data(blockLID, popID);
-                  memset(blockData, 0, WID3*sizeof(Realf));
+            loadTimer.stop();
+
+            phiprof::Timer memsetTimer {memsetTimerId};
+            // reset blocks in all non-sysboundary neighbor spatial cells for this block id
+            for (CellID target_cell_id: DimensionPencils[dimension].binsCells[activeBins[nBin]]) {
```
markusbattarbee: There should be a per-bin collection of target cells to loop over here instead of cross-referencing the list of target cells and these sets. And that collection should be gathered once at the end of `binPencils`.

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -334,103 +334,111 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
...
-                            L,
-                            scalingthreshold,
-                            pencilBlockData,
-                            pencilRatios,
-                            vcell_transpose
-               );
          }
-         propagateTimer.stop();
-         mappingTimer.stop(); // mapping (top-level)
-      } // Closes loop over blocks
```
markusbattarbee, line 433: I'd request keeping these comments about which loop is closed where.

markusbattarbee (changes requested):
Thanks for looking into this! Looks like you figured out a way to do it, just some optimization and clarification requests.

markusbattarbee:
Might be good to get some debug printouts that the binning is actually happening properly - I would expect to see some improvement in the transtest amr cases, as those have only a single block, but a complicated mesh, so this would finally allow some parallelism in those.

vlasovsolver/cpu_trans_pencils.cpp
lkotipal: Yep, this is here for debugging

vlasovsolver/cpu_trans_pencils.cpp
lkotipal: Ditto

vlasovsolver/cpu_trans_map_amr.cpp
lkotipal: Are you sure? I think `vector::size` is just checking a variable, but there might indeed be some optimization if amount of iterations is definitely known before starting the loop

markusbattarbee: I can't quote anything exact right now, but I have a feeling that at least at some point of the GPU development this had a noticeable impact. Probably connected with GPU page faulting. So perhaps completely unnecessary here. Worth trying perhaps, still?

vlasovsolver/cpu_trans_map_amr.cpp
lkotipal: Indeed, I hadn't noticed `DimensionTargetCells` is filled before binning. So in fact I can limit binning to cells that are potentially written into as well

vlasovsolver/cpu_trans_pencils.hpp
lkotipal: Yes actually, that's a very clear bug. I think it just happens to work out because of the loop later on

vlasovsolver/cpu_trans_pencils.hpp
lkotipal: Yup, originally the entire set of matching bins was used to do some merging. Will indeed refactor into a function

vlasovsolver/cpu_trans_pencils.hpp
lkotipal: Now that I think of it, this intial bin matching could even be scrapped in favor of the later loopy thing. Not sure if that would be cleaner

vlasovsolver/cpu_trans_map_amr.cpp
lkotipal: From OpenMP documentation: "Values of the loop control expressions must be the same for all iterations of the loop." I guess it can't hurt to explicitly make it so, but I think OpenMP assumes it's constant anyway

vlasovsolver/cpu_trans_pencils.hpp
lkotipal: (which is exactly what I did)

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -334,103 +334,107 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
...
                }
             }
-         }
-         memsetTimer.stop();
+            memsetTimer.stop();
+
+            phiprof::Timer propagateTimer {propagateTimerId};
+            for (uint pencili : DimensionPencils[dimension].binsPencils[DimensionPencils[dimension].activeBins[nBin]]) {
+               // Skip pencils without blocks
+               if (pencilBlocksCount.at(pencili) == 0)
```
markusbattarbee: Please, let's not explicitly remove braces from if-clauses :) even if it's "trivial".

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -97,6 +106,69 @@ struct setOfPencils {
...
+   void binPencils(const std::unordered_set<CellID>& targetCells) {
+      bins.resize(N);
+
+      // Loop over pencils
+      for (uint i = 0; i < N; ++i) {
+         bins[i] = i;
+         binsCells[i] = {};
+
+         for (auto id = ids.begin() + idsStart[i]; id < ids.begin() + idsStart[i] + lengthOfPencils[i]; ++id) {
+            if (*id && targetCells.contains(*id)) {
```
markusbattarbee: So if we want to think in advanced terms about source and target cells, I think we should gather separate lists - for each bin, a list of source cells, and of target cells.

When flagging bins for deletion, we take the local bin source cells, and see if any of those are target cells of any other bins.

For this gathering of target cells, we cannot use `DimensionTargetCells[dimension]` as that's just a list over all cells (and in fact now not needed for anything). Rather, IMO it'd be better to actually gather a list of target cells when building the pencil itself. 

Thoughts?

lkotipal: `DimensionTargetCells[dimension]` should be the list of all cells that _any_ pencil writes into, so I think that's correct here. As far as I understand the data dependencies:

- Every pencil that reads a cell must read it before it's cleared
- Every pencil that writes into a cell must do it, serially, after it's been cleared

And the cells in a pencil are just a list of cells that are either read/written into with the ratio telling if it's written into at all. So from this I think the binning must be done considering _all cells that any pencil writes into_ which is what this bit should be doing.

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -334,103 +334,108 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
...
-               // Check for invalid block id
-               if (blockLID != vmesh::VelocityMesh::invalidLocalID()) {
-                  // Get a pointer to the block data
-                  Realf* blockData = target_cell->get_data(blockLID, popID);
-                  memset(blockData, 0, WID3*sizeof(Realf));
+            loadTimer.stop();
+
+            phiprof::Timer memsetTimer {memsetTimerId};
+            // reset blocks in all non-sysboundary neighbor spatial cells for this block id
+            for (CellID target_cell_id: DimensionPencils[dimension].binsCells[DimensionPencils[dimension].activeBins[nBin]]) {
```
markusbattarbee: ok here I see you're using `binCells` for clearing. Thus, it should be rather called `binTargetCells` so that the naming still is correct.

vlasovsolver/cpu_trans_map_amr.cpp
```diff
@@ -334,103 +334,108 @@ bool trans_map_1d_amr(const dccrg::Dccrg<spatial_cell::SpatialCell,dccrg::Cartes
...
-                  cellBlockData[start + b] = NULL;
+      // Get global id of the velocity block
+      // Load data for pencils.
+      #pragma omp for schedule(dynamic,1) collapse(2)
+      for(uint blocki = 0; blocki < blocksSize; blocki++) {
+         for (uint nBin = 0; nBin < binsSize; ++nBin) {
+            // For each block + bin we copy first copy each pencil's data into a buffer, clear the target blocks, and then sum the translated pencils in
+
+            phiprof::Timer loadTimer {loadTimerId};
+            vmesh::GlobalID blockGID = unionOfBlocks[blocki];
```
markusbattarbee, line 359: Line 359 is a doozy to read. How about having 
`const int currentActiveBin = DimensionPencils[dimension].activeBins[nBin];` on the above line, then having the loop as 
`for (uint pencili : DimensionPencils[dimension].binsPencils[currentActiveBin]) {` please?

lkotipal:
image
Numbers looking better after ignoring read-only cells in binning

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -43,6 +43,11 @@ struct setOfPencils {
    std::vector< bool > periodic;
    std::vector< std::vector<uint> > path; // Path taken through refinement levels
 
+   std::vector<uint> bins; // Bin of each pencil
```
markusbattarbee: renaming suggestions:
```
std::vector<uint> binOfPencil; // Bin of each pencil
std::map<uint, std::vector<uint>> PencilsInBin; // Vector of pencils in each bin
std::map<uint, std::set<CellID>> targetCellsInBin; // Set of cells in each bin 
std::vector<uint> activeBins; // list of set of active keys in the above two maps

File: vlasovsolver/cpu_trans_pencils.hpp (hunk 15 above)
Comment (markusbattarbee) [reply vlasovsolver/cpu_trans_pencils.hpp]: > `DimensionTargetCells[dimension]` should be the list of all cells that _any_ pencil writes into, so I think that's correct here. 

Yep - but I think we don't actually use that for anything anymore?


> As far as I understand the data dependencies:
> 
>     * Every pencil that reads a cell must read it before it's cleared
> 
>     * Every pencil that writes into a cell must do it, serially, after it's been cleared

Correct.

> And the cells in a pencil are just a list of cells that are either read/written into with the ratio telling if it's written into at all. So from this I think the binning must be done considering _all cells that any pencil writes into_ which is what this bit should be doing.

Ok so if we assume this one is called `binTargetCells`, then it's storing the correct contents. I'll continue reading below to see if it all still matches.

File: vlasovsolver/cpu_trans_pencils.hpp [hunk 19]
```diff
@@ -97,6 +106,69 @@ struct setOfPencils {
...
+      // TODO If anyone knows how to do this more efficiently feel free to fix it
+      std::set<uint> binsToDelete;
+      do {
+         binsToDelete.clear();
+         for (auto bin1 = binsCells.begin(); bin1 != binsCells.end(); ++bin1) {
+            if (binsToDelete.contains(bin1->first)) {
+               continue;
+            }
+
+            auto& cells1 = bin1->second;
```
markusbattarbee: this is now target cells of bin1

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -97,6 +106,69 @@ struct setOfPencils {
...
+            for (auto bin2 = binsCells.begin(); bin2 != binsCells.end(); ++bin2) {
+               if (bin1 == bin2 || binsToDelete.contains(bin2->first)) {
+                  continue;
+               }
+
+               // Check for overlapping cells
+               for (auto cell : bin2->second) {
+                  if (cells1.contains(cell)) {
```
markusbattarbee: and this is target cells of bin2. Thus, there isn't a check to see if there's an overlap of non-target source cells with the target cells of another pencil. Unless it happens somewhere else and I'm missing it?

markusbattarbee (commented):
Trying to do some brain-wrapping...

vlasovsolver/cpu_trans_pencils.hpp
lkotipal: > Yep - but I think we don't actually use that for anything anymore?

Only in binning, but indeed the scope can be reduced

> Ok so if we assume this one is called binTargetCells, then it's storing the correct contents. I'll continue reading below to see if it all still matches.

I think that's correct yes. But we don't need to separately consider targets and sources; if no pencil has a cell as a target it's irrelevant for binning, and if any pencil has it as a target it's considered for all cells. Thus `DimensionTargetCells`.

vlasovsolver/cpu_trans_pencils.hpp
lkotipal: Implemented

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -97,6 +106,85 @@ struct setOfPencils {
...
    }
 
+   void binPencils() {
+      binOfPencil.resize(N);
+
+      // Consider only cells which _any_ pencil writes into for binning,
+      // since read-only cells aren't affected by race conditions
+      std::unordered_set<CellID> targetCells = {};
+      #pragma omp parallel for
+      for (uint i = 0; i < N; ++i) {
```
markusbattarbee: This now does not loop over all cells... N is count of pencils, not of cells.

markusbattarbee:
A memo: It might be worth figuring out which of these bins require which ghost cells to be updated, so we could overlap MPI comms with computation, i.e. start processing some bins as soon as the required cells have been received (and all local cells which have are used by any other tasks as sources have been sent).

vlasovsolver/cpu_trans_pencils.hpp
markusbattarbee: UP?

lkotipal: ~~Both loops are over all bins, so it happens when the indices are swapped at a later iteration~~

`targetCellsInBin` somewhat confusingly means _all_ cells in the bin that are the target of _any_ pencil. So it's both pure sources (as long as it's the target of any pencil so it's relevant for race conditions) and target cells.

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -97,6 +106,85 @@ struct setOfPencils {
...
+            }
+         }
+      }
+
+      // Super ugly!
+      // TODO If anyone knows how to do this more efficiently feel free to fix it
+      std::set<uint> binsToDelete;
+      do {
+         binsToDelete.clear();
+         for (auto bin1 = targetCellsInBin.begin(); bin1 != targetCellsInBin.end(); ++bin1) {
```
markusbattarbee: could we use e.g. `[binIndex1, targetCellsInBin1]` so instead of `bin1->first` below, we'd have `binIndex`?

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -97,6 +106,85 @@ struct setOfPencils {
...
+            #pragma omp critical
+            targetCells.insert(targ);
+         }
+      }
+
+      // Loop over pencils
+      // TODO could be paralellized as well
+      for (uint i = 0; i < N; ++i) {
+         binOfPencil[i] = i;
+         targetCellsInBin[i] = {};
```
markusbattarbee, line 129: Let's please add a comment here clarifying what actually goes in this container. (and/or rename the container)

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -97,6 +106,85 @@ struct setOfPencils {
       path.push_back(pathIn);
    }
 
+   void binPencils() {
+      binOfPencil.resize(N);
+
+      // Consider only cells which _any_ pencil writes into for binning,
+      // since read-only cells aren't affected by race conditions
+      std::unordered_set<CellID> targetCells = {};
```
markusbattarbee: this could be called `allTargetCells`

vlasovsolver/cpu_trans_pencils.hpp
markusbattarbee: or rather, this is any source cells in bin1 which are also a target cell for any pencil. (Let's add a comment!)

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -97,6 +106,85 @@ struct setOfPencils {
...
+      std::set<uint> binsToDelete;
+      do {
+         binsToDelete.clear();
+         for (auto bin1 = targetCellsInBin.begin(); bin1 != targetCellsInBin.end(); ++bin1) {
+            if (binsToDelete.contains(bin1->first)) {
+               continue;
+            }
+
+            auto& cells1 = bin1->second;
+            for (auto bin2 = targetCellsInBin.begin(); bin2 != targetCellsInBin.end(); ++bin2) {
```
markusbattarbee: `for ([binIndex2, targetCellsInBin2]` here?

vlasovsolver/cpu_trans_pencils.hpp
markusbattarbee: ok, so the check is there, it was just too obfuscated for me to see it. Let's fix those namings and add descriptive comments please! How about doxygen-style? :)

markusbattarbee (changes requested):
Thanks for the explanation, good to know the algorithm is correct. Next up, clarifications so the next person scratching their head with this won't have the same questions!

vlasovsolver/cpu_trans_pencils.hpp
lkotipal: Doxygen'd this, at least I think; not sure what a better name would be (just cellsInBin?)

markusbattarbee (approved):

PR #1097, opened 2025-02-12
Fix order for running test printouts in CI

markusbattarbee:


ursg (approved):
Ah, good idea. This has indeed been annoying for a while.

ykempf (approved):

PR #1098, opened 2025-02-12
Reduce some specific timers that cause a wide spread of phiprof files.

ykempf:


sysboundary/sysboundary.cpp
```diff
@@ -124,7 +124,7 @@ void SysBoundary::getParameters() {
 void SysBoundary::addSysBoundary(SBC::SysBoundaryCondition* bc, Project& project, creal& t) {
    // Initialize the boundary condition
    stringstream timername;
-   timername<<"Initialize system boundary condition "<<bc->getName();
+   timername<<"Initialize system boundary condition";
```
markusbattarbee: I think this one and the one on line 633 should not cause phiprof divergence, as all tasks should make those calls.

sysboundary/sysboundarycondition.cpp
```diff
@@ -308,9 +308,7 @@ namespace SBC {
       if(closestCell == INVALID_CELLID) {
          abort_mpi("No closest cell found!", 1);
       }
-      phiprof::Timer boundaryTimer {"vlasovBoundaryCopyFromTheClosestNbr"};
```
markusbattarbee, line 311: approval to all changes in this file, those need to go, also as they are in threaded regions. Is this enough?

ykempf: I don't know, untested. Merge it and make a new PR next time if there's still tens of phiprofs in FIF?

sysboundary/sysboundary.cpp
ykempf: Indeed.

sysboundary/inflow.cpp
```diff
@@ -179,11 +179,9 @@ namespace SBC {
       setBFromTemplate(mpiGrid, perBGrid, BgBGrid);
 
       // Ensure up-to-date velocity block counts for all neighbours
-      phiprof::Timer ghostTimer {"transfer-ghost-blocks", {"MPI"}};
```
markusbattarbee, line 182: this would have been safe, I think? I'd imagine all processes have all boundary conditions existing in case LB does something interesting to them.

markusbattarbee (changes requested):

sysboundary/inflow.cpp
ykempf: Hmm yeah. But the naming is incorrect.

ykempf: Or well calling the same thing a different thing.

markusbattarbee: ok, I guess we can leave it out and then scratch our heads if it comes to that.

PR #1099, opened 2025-02-17
FsGrid & fieldsolver refactoring

cscjlan:
## FsGrid
Vlasiator has been updated to use the refactored FsGrid. This is visible everywhere FsGrid was used in. Some notable differences are dicussed below.

First, there's only a single `FsGrid` object and it's called `fsgrid`. It no longer stores any data. The data that used to be stored by the different grids are now owned by a structure called `FsData`, one for each ex-grid. The data is not passed around by passing the FsData structures, but rather by passing an `std::span`. You can get an `std::span` from `FsData` with the `view()` method: `auto technicalSpan = technicalData.view();`

`updateGhostCells` now takes in an `std::span`: `fsgrid.updateGhostCells(technical);` instead of `technicalGrid.updateGhostCells()`.

Data is no longer accessed through the `grid.get(i, j, k)` method, but rather by contructing an `FsStencil`, and using it to compute a 1D index from a 3D index:
```cpp
auto stencil = fsgrid.makeStencil(i, j, k);
// Get the center element ('ooo'), i.e. the one corresponding to i, j, k
const auto center = technical[stencil.ooo()];

// The element in the -x direction from the center, i.e. i - 1, j, k
const auto x_left = technical[stencil.moo()];
```
This has the benefit that some values that are constant throughout the simulation can be computed once and stored in the fsgrid object. A single `FsStencil` is constructed for each iteration of a loop (i.e. one for each i, j, k). The constant values can be accessed by the `FsStencil` when it computes the 1D value from the 3D index it was constructed with.

## Fieldsolver
Fieldsolver has been refactored to use the updated FsGrid. Some files have undergone heavier refactoring (`ldz_electric_field.cpp`), some less heavy. The underlying reason for all of the refactoring is to make it easier to change the fieldsolver to use the updated FsGrid: with less code repetition there's less work to be done (and thus fewer bugs) when changing the code to use the stencil and the spans. An additional benefit was a speed up especially in the electric field computation.

`derivatives.cpp` and `ldz_volume.cpp` use the newly added `fsgrid.parallel_for`, which is an architecture/processor independent way of doing the desired computation. It takes in a lambda, which it calls for each local cell on the fsgrid.


cscjlan:
Note that some of the diffs are completely unintelligible due to `clang-format`. To get a clearer diff, it's best to format some of the files and view the diffs locally.

Files:
- datareduction/datareducer.cpp
- fieldtracing/fieldtracing.cpp
- fieldtracing/fieldtracing.h
- grid.cpp
- ioread.cpp
- sysboundary/copysphere.cpp
- sysboundary/inflow.cpp
- sysboundary/ionosphere.cpp
- sysboundary/outflow.cpp
- tools/vlsvdiff.cpp

This script creates a complete diff locally:
```bash
#!/bin/bash

set -exu

git clone http://github.com/fmihpc/vlasiator.git
cd vlasiator/
git remote add csc https://github.com/cschpc/vlasiator.git
git fetch csc
git checkout -b clang-format-dev origin/dev

declare -a to_format=(
    "datareduction/datareducer.cpp"
    "fieldtracing/fieldtracing.cpp"
    "fieldtracing/fieldtracing.h"
    "grid.cpp"
    "ioread.cpp"
    "sysboundary/copysphere.cpp"
    "sysboundary/inflow.cpp"
    "sysboundary/ionosphere.cpp"
    "sysboundary/outflow.cpp"
    "tools/vlsvdiff.cpp"
)

for file in "${to_format[@]}"; do
    clang-format --style=file:.clang-format -i "${file}"
done

git add .
git commit -m "format some files"
git diff HEAD csc/fsgrid_merge_without_override > git_diff
```
Save in a file, execute it, then view the file `vlasiator/git_diff`.

backgroundfield/backgroundfield.cpp
```diff
@@ -20,142 +20,126 @@
...
+// clang-format on
 
-//FieldFunction should be initialized
-void setBackgroundField(
-   const FieldFunction& bgFunction,
-   FsGrid< std::array<Real, fsgrids::bgbfield::N_BGB>, FS_STENCIL_WIDTH> & BgBGrid,
-   bool append
-   ) {
+// FieldFunction should be initialized
+void setBackgroundField(const FieldFunction& bgFunction, std::span<std::array<Real, fsgrids::bgbfield::N_BGB>> bgb,
```
markusbattarbee: Would it be beneficial to define, inside the `fsgrids` namespace, an alias like `fsgrids::bgbspan` which would signify `std::span<std::array<Real, fsgrids::bgbfield::N_BGB>>` ?

And similarly for all other fsgrid object spans.

I see that on line 34 there is indeed a much cleaner `std::span<fsgrids::technical>`. That approach could also be nice?

backgroundfield/backgroundfield.cpp
```diff
@@ -20,142 +20,126 @@
...
+                        std::span<fsgrids::technical> technical, FieldSolverGrid &fsgrid, bool append) {
```
markusbattarbee: is this really a clang-formatted syntax for arguments? putting as many on a single line as fits, then wrapping, resulting now in 2 per line?

backgroundfield/backgroundfield.cpp
```diff
@@ -20,142 +20,126 @@
...
-               dx[0] = BgBGrid.DX;
-               dx[1] = BgBGrid.DY;
-               dx[2] = BgBGrid.DZ;
-               double end[3];
-               end[0]=start[0]+dx[0];
-               end[1]=start[1]+dx[1];
-               end[2]=start[2]+dx[2];
+// These are threaded now that the dipole field is threadsafe
+#pragma omp parallel for collapse(2)
+      for (auto z = 0; z < localSize[2]; ++z) {
```
markusbattarbee: So this PR does not yet convert all FSgrid operations to use `fsgrid.parallel_for()`? I'd think with all the changes going in now, the parallel loops might as well also go in now.

backgroundfield/backgroundfield.h
```diff
@@ -23,114 +23,90 @@
...
-   FsGrid< std::array<Real, numFields>, FS_STENCIL_WIDTH> & BGrid,
-   int offset=fsgrids::bfield::PERBX
-   ) {
-   std::array<FsGridTools::FsIndex_t,3> localSize = BGrid.getLocalSize();
-
-   #pragma omp parallel for collapse(2)
-   for (FsGridTools::FsIndex_t z = 0; z < localSize[2]; ++z) {
-      for (FsGridTools::FsIndex_t y = 0; y < localSize[1]; ++y) {
-         for (FsGridTools::FsIndex_t x = 0; x < localSize[0]; ++x) {
-            // This is still N_BFIELD (==3) instead of numFields
```
markusbattarbee, line 58: Let's please not lose important comments like this

datareduction/datareducer.cpp
```diff
@@ -47,492 +44,494 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
-                        retval[3*(gridSize[1]*gridSize[0]*z + gridSize[0]*y + x) + 1] = (*BgBGrid.get(x,y,z))[fsgrids::BGBY];
-                        retval[3*(gridSize[1]*gridSize[0]*z + gridSize[0]*y + x) + 2] = (*BgBGrid.get(x,y,z))[fsgrids::BGBZ];
-                     }
-                  }
-               }
-               return retval;
-         }
-         ));
+      for (auto& c : lowercase)
+         c = tolower(c);
```
markusbattarbee, line 48: let's add curly braces here.

datareduction/datareducer.cpp
```diff
@@ -47,492 +44,494 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
+
+      if (P::systemWriteAllDROs || lowercase == "fg_b" ||
+          lowercase == "b") { // Bulk magnetic field at Yee-Lattice locations
+         outputReducer->addOperator(new DRO::DataReductionOperatorFsGrid(
+             "fg_b", [](const FieldSolverData& fieldSolverData) -> std::vector<double> {
+                const auto* localSize = &fieldSolverData.fsgrid.getLocalSize()[0];
+                std::vector<double> retval(localSize[0] * localSize[1] * localSize[2] * 3);
+
+                // Iterate through fsgrid cells and extract total magnetic field
+                for (int z = 0; z < localSize[2]; z++) {
```
markusbattarbee: Since this file is also changing a lot, might as well move to the fsgrid parallel_for construct straight away.

datareduction/datareductionoperator.cpp
```diff
@@ -126,56 +127,40 @@ namespace DRO {
...
-      attribs["unitLaTeX"]=unitLaTeX;
-      attribs["unitConversion"]=unitConversion;
-      attribs["variableLaTeX"]=variableLaTeX;
-
-      std::vector<double> varBuffer =
-         lambda(perBGrid,EGrid,EHallGrid,EGradPeGrid,momentsGrid,dPerBGrid,dMomentsGrid,BgBGrid,volGrid,technicalGrid);
+   bool DataReductionOperatorFsGrid::writeFsGridData(const FieldSolverData& fieldSolverData,
+                                                     const std::string& meshName, vlsv::Writer& vlsvWriter,
+                                                     const bool writeAsFloat) {
+      const std::map<std::string, std::string> attribs = {
```
markusbattarbee, line 134: nice, cleaner syntax for initializing this 👍

markusbattarbee (commented):
I started reviewing, but it really is a massive PR. I wonder, would it be feasible to have another branch with no actual code changes, but the same comment / re-ordering / whitespace / line break changes as here, so one could compare against that to see what's actually changed?

markusbattarbee:
> I started reviewing, but it really is a massive PR. I wonder, would it be feasible to have another branch with no actual code changes, but the same comment / re-ordering / whitespace / line break changes as here, so one could compare against that to see what's actually changed?

And of course there were instructions for that. Web-diffs available at:
https://github.com/markusbattarbee/vlasiator-public/pull/24/files?diff=split&w=1

Evaluating those now.

vlasiator.cpp
```diff
@@ -1173,37 +1069,22 @@ int simulate(int argn,char* args[]) {
 
          phiprof::Timer couplingInTimer {"fsgrid-coupling-in"};
          // Copy moments over into the fsgrid.
-         //setupTechnicalFsGrid(mpiGrid, cells, technicalGrid);
-         feedMomentsIntoFsGrid(mpiGrid, cells, momentsGrid, technicalGrid, false);
-         feedMomentsIntoFsGrid(mpiGrid, cells, momentsDt2Grid, technicalGrid, true);
+         //setupTechnicalFsGrid(mpiGrid, cells, technical, fsgrid);
```
ykempf: This particular line can go.

vlasiator.cpp
```diff
@@ -1173,37 +1069,22 @@ int simulate(int argn,char* args[]) {
...
-            technicalGrid,
-            sysBoundaryContainer,
-            P::dt,
-            P::fieldSolverSubcycles
-         );
+
+         propagateFields(perb.view(), perbdt2.view(), e.view(), edt2.view(), ehall.view(), egradpe.view(),
+                         egradpedt2.view(), moments.view(), momentsdt2.view(), dperb.view(), dmoments.view(),
+                         dmomentsdt2.view(), bgb.view(), vol.view(), technical.view(), fsgrid, sysBoundaryContainer,
+                         P::dt, P::fieldSolverSubcycles);
```
ykempf, line 1174: I don't think that this formatting style makes it more readable. Neither does it when a one-liner is split in two.

vlasiator.cpp
```diff
@@ -1173,37 +1069,22 @@ int simulate(int argn,char* args[]) {
...
 
          phiprof::Timer getFieldsTimer {"getFieldsFromFsGrid"};
          // Copy results back from fsgrid.
-         volGrid.updateGhostCells();
-         technicalGrid.updateGhostCells();
-         getFieldsFromFsGrid(volGrid, BgBGrid, EGradPeGrid, dMomentsGrid, technicalGrid, mpiGrid, cells);
+         fsgrid.updateGhostCells(vol.view());
+         fsgrid.updateGhostCells(technical.view());
+         getFieldsFromFsGrid(vol.view(), bgb.view(), egradpe.view(), dmoments.view(), technical.view(), fsgrid, mpiGrid,
+                             cells);
```
ykempf, line 1181: Like, ugh!

vlasiator.cpp
```diff
@@ -1282,7 +1164,7 @@ int simulate(int argn,char* args[]) {
       propagateTimer.stop(computedCells,"Cells");
       
       phiprof::Timer endStepTimer {"Project endTimeStep"};
-      project->hook(hook::END_OF_TIME_STEP, mpiGrid, perBGrid);
+      project->hook(hook::END_OF_TIME_STEP, mpiGrid, perb.view(), technical.view(), fsgrid);
```
ykempf, line 1269: This one needs technical now when it didn't before?

vlasiator.cpp
```diff
@@ -109,9 +111,9 @@ void addTimedBarrier(string name){
    MPI_Barrier(MPI_COMM_WORLD);
 }
 
-void computeNewTimeStep(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
-			FsGrid< fsgrids::technical, FS_STENCIL_WIDTH> & technicalGrid, Real &newDt, bool &isChanged) {
-
+void computeNewTimeStep(dccrg::Dccrg<SpatialCell, dccrg::Cartesian_Geometry>& mpiGrid,
+                        std::span<fsgrids::technical> technical, FieldSolverGrid &fsgrid, Real& newDt,
```
ykempf: Taking this one as an example (I now only looked through this file) – I would find it more logical to have fsgrid first, then the views. But that'll be a tedious change of now instead consistently fsgrid is always last in such lists. So if the Vlasiator team is fine with this, let's go with it and go into "proper" things to do here in this PR.

ykempf (commented):
Went through vlasiator.cpp, let's see how I should proceed contributing, we can discuss on zulip or here.

fieldsolver/derivatives.cpp
```diff
@@ -22,482 +22,345 @@
...
-         dMoments->at(fsgrids::dmoments::drhoqdz) = (rghtMoments->at(fsgrids::moments::RHOQ)-leftMoments->at(fsgrids::moments::RHOQ))/2;
-         dMoments->at(fsgrids::dmoments::dp11dz) = (rghtMoments->at(fsgrids::moments::P_11)-leftMoments->at(fsgrids::moments::P_11))/2;
-         dMoments->at(fsgrids::dmoments::dp22dz) = (rghtMoments->at(fsgrids::moments::P_22)-leftMoments->at(fsgrids::moments::P_22))/2;
-         dMoments->at(fsgrids::dmoments::dp33dz) = (rghtMoments->at(fsgrids::moments::P_33)-leftMoments->at(fsgrids::moments::P_33))/2;
-         dMoments->at(fsgrids::dmoments::dVxdz)  = (rghtMoments->at(fsgrids::moments::VX)-leftMoments->at(fsgrids::moments::VX))/2;
-         dMoments->at(fsgrids::dmoments::dVydz)  = (rghtMoments->at(fsgrids::moments::VY)-leftMoments->at(fsgrids::moments::VY))/2;
-         dMoments->at(fsgrids::dmoments::dVzdz)  = (rghtMoments->at(fsgrids::moments::VZ)-leftMoments->at(fsgrids::moments::VZ))/2;
+      for (size_t i = 0; i < moms.size(); i++) {
+         dMoments[dmiz[i]] = computeDiff(moms[i], momData.oop, momData.oom);
```
ykempf, line 114: Why three consecutive loops here?

fieldsolver/derivatives.cpp
```diff
@@ -22,482 +22,345 @@
...
-      if(calculateMoments) {
-         dMoments->at(fsgrids::dmoments::drhomdy) = (rghtMoments->at(fsgrids::moments::RHOM)-leftMoments->at(fsgrids::moments::RHOM))/2;
-         dMoments->at(fsgrids::dmoments::drhoqdy) = (rghtMoments->at(fsgrids::moments::RHOQ)-leftMoments->at(fsgrids::moments::RHOQ))/2;
-         dMoments->at(fsgrids::dmoments::dp11dy) = (rghtMoments->at(fsgrids::moments::P_11)-leftMoments->at(fsgrids::moments::P_11))/2;
-         dMoments->at(fsgrids::dmoments::dp22dy) = (rghtMoments->at(fsgrids::moments::P_22)-leftMoments->at(fsgrids::moments::P_22))/2;
-         dMoments->at(fsgrids::dmoments::dp33dy) = (rghtMoments->at(fsgrids::moments::P_33)-leftMoments->at(fsgrids::moments::P_33))/2;
-         dMoments->at(fsgrids::dmoments::dVxdy)  = (rghtMoments->at(fsgrids::moments::VX)-leftMoments->at(fsgrids::moments::VX))/2;
-         dMoments->at(fsgrids::dmoments::dVydy)  = (rghtMoments->at(fsgrids::moments::VY)-leftMoments->at(fsgrids::moments::VY))/2;
-         dMoments->at(fsgrids::dmoments::dVzdy)  = (rghtMoments->at(fsgrids::moments::VZ)-leftMoments->at(fsgrids::moments::VZ))/2;
+   if (notSysBoundary) {
```
ykempf: This now does finite differencing instead of limiter-derivatives `if(notSysBoundary)` . I need to look up what `notSysBoundary` stands for but "orally" we should do limiters if we're not at boundaries, and finite differences if we're at boundaries L1 or just inside the domain in L2 cells.

fieldsolver/derivatives.cpp
```diff
@@ -22,482 +22,345 @@
...
-      dPerB->at(fsgrids::dperb::dPERBydxx) = leftPerB->at(fsgrids::bfield::PERBY) + rghtPerB->at(fsgrids::bfield::PERBY) - 2.0*centPerB->at(fsgrids::bfield::PERBY);
-      dPerB->at(fsgrids::dperb::dPERBzdxx) = leftPerB->at(fsgrids::bfield::PERBZ) + rghtPerB->at(fsgrids::bfield::PERBZ) - 2.0*centPerB->at(fsgrids::bfield::PERBZ);
+      const auto& rv = momData.poo[mom::RHOM];
+      if (rv <= 0) {
+         std::cerr << __FILE__ << ":" << __LINE__ << (rv < 0 ? " Negative" : " Zero") << " density in spatial cell"
+                   << std::endl;
+         abort();
+      }
+#endif
    }
```
ykempf, line 97: Why do we need a { } scope around this ifdef?

fieldsolver/derivatives.cpp
```diff
@@ -22,482 +22,345 @@
...
+                          std::span<std::array<Real, fsgrids::dperb::N_DPERB>> dperb,
+                          std::span<std::array<Real, fsgrids::dmoments::N_DMOMENTS>> dmoments,
+                          const fsgrid::FsStencil& stencil, cuint sysBoundaryFlag, cuint sysBoundaryLayer,
+                          const bool doMoments) {
+   /*
+    * For sysBoundaryLayer 1 or 2, we are near a boundary, and we wish to use regular centered differences instead of
+    * slope limiter-adjusted values. This is to minimize oscillations as a smooth behaviour is required near artificial
+    * boundaries, unlike at boundaries and shocks inside the simulation domain.
+    */
+   const bool notSysBoundary =
```
ykempf: All right, found it here, I think this should be renamed. Agreed?

fieldsolver/derivatives.cpp
ykempf: To collapse a block in the editor, no further technical reason.

ykempf:
In this latest commit the lambda copies by reference `[&]` as the `[=]` caused it to shout along the lines of
```
fieldsolver/ldz_electric_field.cpp:1254:63: error: binding reference of type ‘SysBoundary&’ to ‘const SysBoundary’ discards qualifiers 
```

ykempf:
Same "issue" in the next one – not an issue I'm sure, just noting it here.

backgroundfield/backgroundfield.cpp
```diff
@@ -20,142 +20,114 @@
...
-   for (FsGridTools::FsIndex_t z = 0; z < localSize[2]; ++z) {
-      for (FsGridTools::FsIndex_t y = 0; y < localSize[1]; ++y) {
-         for (FsGridTools::FsIndex_t x = 0; x < localSize[0]; ++x) {
-            for (FsGridTools::FsIndex_t i = 0; i < fsgrids::bgbfield::N_BGB; ++i) {
-               BgBGrid.get(x,y,z)->at(i) = 0;
-            }
-         }
-      }
+void setBackgroundFieldToZero(std::span<std::array<Real, fsgrids::bgbfield::N_BGB>> bgb) {
+#pragma omp parallel for
```
ykempf: Ehm. Should this also get a lambda?

backgroundfield/backgroundfield.h
```diff
@@ -23,114 +23,86 @@
...
-            for (int i = 0; i < fsgrids::bfield::N_BFIELD; ++i) {
-               BGrid.get(x,y,z)->at(offset+i) = 0;
-            }
-         }
+template <long unsigned int numFields>
+void setPerturbedFieldToZero(std::span<std::array<Real, numFields>> b, int offset = fsgrids::bfield::PERBX) {
+#pragma omp parallel for
```
ykempf: Ah, same here? I had grepped for nested loops, but I guess this should then also be ported?

backgroundfield/backgroundfield.cpp
markusbattarbee: up

fieldsolver/derivatives.cpp
ykempf: Renaming in an upcoming commit.

fieldsolver/derivatives.cpp
ykempf: Upcoming

backgroundfield/backgroundfield.cpp
ykempf: boom

backgroundfield/backgroundfield.h
markusbattarbee: UP!

ykempf: Right. So should this now use the template value then? Or do we keep that comment in then?

markusbattarbee: It's an important comment - the comment needs to stay! It's used to set the perturbed field to zero, but it's also used in some cases with the backgroundfield, and needs to only affect 3 components at once.

ykempf: Yeah but no, it's only ever used in the one call 10 lines below.

vlasiator.cpp
ykempf: Ah, OK. It still doesn't, but if someone were to call for a parallel_for in there it would be needed. Shall I remove the passing of technical, or do we keep it here? Doesn't hurt I guess...

markusbattarbee: Let's keep it,  imo

backgroundfield/backgroundfield.h
markusbattarbee: pfft, when did that change... 👀

ykempf: There's both setPerturbed* and setBackground*, one in .h and the other in .cpp. The background one is called in a bunch of places.

backgroundfield/backgroundfield.cpp
cscjlan: Clang-format has very many options that can be modified, this is one of them. https://clang.llvm.org/docs/ClangFormatStyleOptions.html#binpackarguments

fieldsolver/fs_common.cpp
```diff
@@ -207,83 +213,70 @@ void reconstructionCoefficients(
...
+   cuint cellSysBoundaryFlag = technical[stencil.ooo()].sysBoundaryFlag;
    if (cellSysBoundaryFlag != sysboundarytype::NOT_SYSBOUNDARY) {
-      std::array<Real, 3> zero = {0,0,0};
-      return zero;
+      return {0, 0, 0};
    }
 
    // Balsara reconstruction formulas: x,y,z are in [-1/2, 1/2] local coordinates
-   std::array<Real, 3> xLocal = getFractionalFsGridCellForCoord(technicalGrid, x);
+   std::array<Real, 3> xLocal = fsgrid.physicalToFractionalGlobal(x[0], x[1], x[2]);
```
ykempf: Oh this is not named correctly, the function returns a local fractional coordinate. So it does what it should, just with an incorrect function name. @markusbattarbee if you confirm I'll change that.

fieldsolver/ldz_electric_field.cpp
```diff
@@ -20,57 +20,141 @@
...
+   template <size_t N>
+   std::tuple<Real, Real> compute(const std::array<Real, N>& nbr_arr, const std::array<Real, N>& arr, size_t i,
+                                  size_t j, Real mul0 = 1.0, Real mul1 = 1.0) const {
+      const Real a = nbr_arr[i] + nbr_bgb[j];
+      const Real b = arr[i] + bgb[j];
+
+      return {mul0 * (a + b), mul1 * (a - b)};
+   }
+
+   Real momentsCoeff(Real dir0, Real dir1, size_t i, size_t j, size_t k, Real min, Real max) const {
```
ykempf: So this is interpolating (volume-averaged i.e. ~cell-centred) `moment[i]` by half a cell width in two directions to obtain a cell-edge value. Would a name like `interpolateToCellEdge` be better (more verbose but more explicit)? Or just `interpolate` (with a comment)?

ykempf: All right I went through this file and renamed some of the things, and I added some documentation like the Balsara and Londrillo & del Zanna references.

datareduction/datareducer.cpp
```diff
@@ -47,317 +44,250 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
+         outputReducer->addOperator(new DRO::DataReductionOperatorFsGrid(
+            "fg_b", [](const FieldSolverData& fieldSolverData) -> std::vector<double> {
+            const auto* localSize = &fieldSolverData.fsgrid.getLocalSize()[0];
+            std::vector<double> retval(localSize[0] * localSize[1] * localSize[2] * 3);
+
+            // Iterate through fsgrid cells and extract total magnetic field
+            fieldSolverData.fsgrid.serial_for_ijk([](int timerId) -> phiprof::Timer { return phiprof::Timer{timerId}; },
+                                                  phiprof::initializeTimer("DRO_fg"), fieldSolverData.technical,
+                                                  [=, &retval](const fsgrid::FsStencil& stencil, cuint sysBoundaryFlag, cuint sysBoundaryLayer, const fsgrid::FsIndex_t i, const fsgrid::FsIndex_t j, const fsgrid::FsIndex_t k) {
+               const auto lid = stencil.ooo();
```
markusbattarbee, line 62: Here and in many places below - aren't `lid` and `ri` the same value? Or was it that `lid` is the index accounting for ghost regions? In that case, isn't here an accessor for getting `ri` without calculating it in the lambda?

datareduction/datareducer.cpp
```diff
@@ -3754,7 +3025,8 @@ bool DataReducer::addOperator(DRO::DataReductionOperator* op) {
  * @return Name of the operator.
  */
 std::string DataReducer::getName(const unsigned int& operatorID) const {
-   if (operatorID >= operators.size()) return "";
+   if (operatorID >= operators.size())
+      return "";
```
markusbattarbee, line 3029: curlies

datareduction/datareducer.cpp
ykempf: I don't think so, that lid is maybe ill-named but it's an index into the span I believe whereas that ri is a "traditional" local ID into the rectangular output array.

fieldsolver/derivatives.cpp
```diff
@@ -22,482 +22,342 @@
...
-         botRght = perBGrid.get(i+1,j,k-1);
-         topLeft = perBGrid.get(i-1,j,k+1);
-         topRght = perBGrid.get(i+1,j,k+1);
-         dPerB->at(fsgrids::dperb::dPERBydxz) = FOURTH * (botLeft->at(fsgrids::bfield::PERBY) + topRght->at(fsgrids::bfield::PERBY) - botRght->at(fsgrids::bfield::PERBY) - topLeft->at(fsgrids::bfield::PERBY));
-      } else {
-         SBC::SysBoundaryCondition::setCellDerivativesToZero(dPerBGrid, dMomentsGrid, i, j, k, 4);
-      }
+   auto computeDiff = [](const auto& i, const auto& right, const auto& left) { return 0.5 * (right[i] - left[i]); };
+
+   auto computeLimiter = [](const auto& i, const auto& right, const auto& left, const auto& center) {
```
markusbattarbee, line 142: I kinda dislike this ordering - limiter takes left,center,right, but `computeLimiter` takes right, left, center...

fieldsolver/derivatives.cpp
```diff
@@ -22,482 +22,342 @@
...
+             compute(stencil.mom(), stencil.pom(), stencil.mop(), stencil.pop(), bfi::PERBY);
+         dPerB[dpb::dPERBzdxy] =
+             compute(stencil.mmo(), stencil.pmo(), stencil.mpo(), stencil.ppo(), bfi::PERBZ);
       }
    }
 }
 
+/*! \brief Low-level spatial derivatives calculation.
+ *
+ * For the cell with ID cellID calculate the spatial derivatives or apply the derivative boundary conditions defined in
```
markusbattarbee: outdated comments

fieldsolver/derivatives.cpp
```diff
@@ -559,205 +405,131 @@ void calculateBVOLDerivativesSimple(
...
-      rght_z_bz /= rght_z_bnorm;
-
-      vol->at(fsgrids::volfields::CURVATUREX) = bx * 0.5*(rght_x_bx-left_x_bx) / technicalGrid.DX + by * 0.5*(rght_y_bx-left_y_bx) / technicalGrid.DY + bz * 0.5*(rght_z_bx-left_z_bx) / technicalGrid.DZ;
-      vol->at(fsgrids::volfields::CURVATUREY) = bx * 0.5*(rght_x_by-left_x_by) / technicalGrid.DX + by * 0.5*(rght_y_by-left_y_by) / technicalGrid.DY + bz * 0.5*(rght_z_by-left_z_by) / technicalGrid.DZ;
-      vol->at(fsgrids::volfields::CURVATUREZ) = bx * 0.5*(rght_x_bz-left_x_bz) / technicalGrid.DX + by * 0.5*(rght_y_bz-left_y_bz) / technicalGrid.DY + bz * 0.5*(rght_z_bz-left_z_bz) / technicalGrid.DZ;
-   }
+void calculateCurvature(fsgrids::volspan vol,
+                        fsgrids::constbgbspan bgb,
+                        const fsgrid::FsStencil& stencil, const std::array<Real, 3>& gridSpacing) {
+   auto compute = [&vol, &bgb](auto i) -> std::array<Real, 3> {
```
markusbattarbee: this should be renamed to normalizeVector or something similar.

fieldsolver/derivatives.cpp
```diff
@@ -559,205 +405,131 @@ void calculateBVOLDerivativesSimple(
...
+   fsgrid.updateGhostCells(vol);
+   commTimer.stop(numCells, "Spatial Cells");
+
+   fsgrid.parallel_for([](int timerId) -> phiprof::Timer { return phiprof::Timer{timerId}; },
+                       phiprof::initializeTimer("Calculate curvature compute cells"), technical,
+                       [=](const fsgrid::FsStencil& stencil, cuint sysBoundaryFlag, cuint sysBoundaryLayer) {
+                          const bool compute = (sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY &&
+                                     sysBoundaryLayer != 1 && sysBoundaryLayer != 2);
+                          if (compute) {
+                             calculateCurvature(vol, bgb, stencil, gridSpacing);
```
markusbattarbee: should there be an `else set to zero` part here as well?

ykempf: Is it not zero at init? I'd rather do a zero initialization in that case and not bother the rest of the time?

datareduction/datareducer.cpp
ykempf: As for accessing, no it'd be recomputed, it could of course be hidden into yet another `parallel_for`. I put a `parallel_for_cellid` that computes a global id, I could add one providing a local id to the lambda instead but the computation would have to be there nevertheless.

fieldsolver/ldz_magnetic_field.cpp
```diff
@@ -21,136 +21,131 @@
...
    if (doX == true) {
       switch (RKCase) {
-         case RK_ORDER1:
-            EGrid0 = EGrid.get(i,j,k);
-            EGrid1 = EGrid.get(i,j+1,k);
-            EGrid2 = EGrid.get(i,j,k+1);
-            perBGrid0->at(fsgrids::bfield::PERBX) += dt/dz*(EGrid2->at(fsgrids::efield::EY) - EGrid0->at(fsgrids::efield::EY)) + dt/dy*(EGrid0->at(fsgrids::efield::EZ) - EGrid1->at(fsgrids::efield::EZ));
-            break;
+      case RK_ORDER1: {
+         const auto& EGrid0 = e[stencil.ooo()]; // i,j,k;
```
ykempf: Do we need these (misaligned) comments? I already got used to poo et al.

datareduction/datareducer.cpp
cscjlan: I originally named them `lid` and `ri` as they were 'a local index' and 'return value index'.

Looking at the code below, `volGrid` loops over the `gridSize` (which, if anything makes any sense, is the local index as the `technicalGrid` function is called `getLocalSize()`) and it stores values to the `retval` array.

Of course it's possible you have different semantics for what is a local index and what is not, but I decided to call them `lid` and `ri` based on what they do in this context, if that makes sense.
```cpp
std::array<FsGridTools::FsIndex_t,3>& gridSize = technicalGrid.getLocalSize();
for(int z=0; z<gridSize[2]; z++) {
    for(int y=0; y<gridSize[1]; y++) {
        for(int x=0; x<gridSize[0]; x++) {
            retval[3*(gridSize[1]*gridSize[0]*z + gridSize[0]*y + x)] =     (*volGrid.get(x,y,z))[fsgrids::volfields::CURVATUREX];
            retval[3*(gridSize[1]*gridSize[0]*z + gridSize[0]*y + x) + 1] = (*volGrid.get(x,y,z))[fsgrids::volfields::CURVATUREY];
            retval[3*(gridSize[1]*gridSize[0]*z + gridSize[0]*y + x) + 2] = (*volGrid.get(x,y,z))[fsgrids::volfields::CURVATUREZ];
        }
    }
}
```

And yes, `lid` is the more complicated index computed originally in the `fsgrid.get()` method, which is now partially precomputed by the stencil.

fieldsolver/derivatives.cpp
cscjlan: I remember being dissatisfied with this as well. I left it that way due to the symmetry of `computeDiff`, `computeLimiter` and `compute2ndDerivative`: the arguments to those are now in the same order:

```cpp
dPerB[dpb::dPERBydx] = computeDiff(bfi::PERBY, perbData.poo, perbData.moo);
```
```cpp
dPerB[dpb::dPERBydx] = computeLimiter(bfi::PERBY, perbData.poo, perbData.moo, perbData.ooo);
```

```cpp
dPerB[dpb::dPERBydxx] = compute2ndDerivative(bfi::PERBY, perbData.poo, perbData.moo, perbData.ooo);
```

I do agree it's more sensible to move from the left, through the center to the right, but it breaks down when there's only left and right. In that case I find it easier to follow, when the common arguments are in the same order and the extra arguments come after the common ones.

sysboundary/copysphere.cpp
```diff
@@ -248,508 +263,391 @@ namespace SBC {
...
-                     }
-                  }
-               }
+         }
+      };
+   
+      auto averageAllNeighbours = [&stencil, &technical, &b, &perbComponent](auto predicateLambda, auto& sum,
+                                                                             auto& nCells) {
+         for (const auto& i : stencil.indices()) {
+            if (predicateLambda(technical[i])) {
```
ykempf, line 550: What is this `predicateLambda`??

sysboundary/inflow.cpp
```diff
@@ -175,153 +107,155 @@ namespace SBC {
...
+   
                // TODO: This code up to determineFace() should be in a separate
                // function, it gets called in a lot of places.
                // Shift to the center of the fsgrid cell
                auto cellCenterCoords = coords;
-               cellCenterCoords[0] += 0.5 * perBGrid.DX;
-               cellCenterCoords[1] += 0.5 * perBGrid.DY;
-               cellCenterCoords[2] += 0.5 * perBGrid.DZ;
-
+               cellCenterCoords[0] += 0.5 * gridSpacing[0];
```
ykempf: You agree that this is a beautiful brainfart @markusbattarbee?

ykempf: Ah the block selection went somewhere. I meant down to the mpiGrid reflevel query and what we do with that just thereafter.

markusbattarbee: Hm? That's the way we determine if the fsgrid cell is associated with a sysb-face vlasov cell...

ykempf: No, this just extracts dx as a 2^-refLvl instead of using fsgrid's gridSpacing.

ykempf: Ah no ugh. We get an equivalent mpiGrid dx. duh.

ykempf: Can't we use technical's refLevel?

ykempf:
All right this latest one should work (TM) but I had to do two copies of the same class-member-free function, which is horrible. Suggestions to fix that welcome. I can think of inserting that as a preprocessor macro but that'd be ugly too I suspect. I gotta stop now, let's see if CI survives this, and discuss tomorrow if there's more elegant suggestions.

fieldsolver/ldz_magnetic_field.cpp
cscjlan: I guess just remove them? I think I left them there originally while changing them, so I could make sure I changed them correctly.

sysboundary/copysphere.cpp
cscjlan: It's just some lambda that this `averageAllNeighbours` takes in to check whether to include the component in the sum. It's used at (new, added) lines 571 and 575 with different predicates, namely
```cpp
auto bitFieldSet = [&bitfield](auto& tech) { return (tech.SOLVE & bitfield) == bitfield; };
auto sbLayerIsOne = [](auto& tech) { return tech.sysBoundaryLayer == 1; };
```

which correspond to the if checks at (old, removed) lines 622 and 646.

fieldsolver/fs_common.cpp
```diff
@@ -294,200 +287,82 @@ std::array<Real, 3> interpolatePerturbedB(
...
-
-            Real coupling = abs(xoffset - (cell[0]-fsc[0])) * abs(yoffset - (cell[1]-fsc[1])) * abs(zoffset - (cell[2]-fsc[2]));
-
-            // Only couple to actual simulation cells
-            if(technicalGrid.get(lfsc[0]+xoffset,lfsc[1]+yoffset,lfsc[2]+zoffset)->sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY) {
-               couplingSum += coupling;
-            } else {
-               continue;
+   const std::array<Real, Rec::N_REC_COEFFICIENTS> rc = [&stencil, &reconstructionCoefficientsCache, &perb, &dperb]() {
+      std::array<int, 3> cellIds = {stencil.i, stencil.j, stencil.k};
```
cscjlan, line 325: Example of using `i, j, k` from stencil to access the loop variables. So no need for `parallel_for_ijk`.

markusbattarbee:
We had discussion about some remaining omp parallel for loops over fsgrid values during e.g. initialization, and challenges related to passing the RNG object inside the lambda. My suggestion is: 
 No RNG allowed inside fsgrid loops. (thus, no passing of RNG objects to FS parallel_for lambdas)
 
 This is because cell-based RNG of any fields will lead to fine-grained noise, which will result in plenty of local minima or maxima, which will not be propagated at all due to the slope limiters used by our field solver. Thus, it is never (tm) a good idea to apply cell-based randomization to field data.
 
 Disturbances in project init should thus be instructed to always randomize densities or bulkv values, not fields. And by preventing this action, we're actually guiding new users to design projects in the correct manner.

fieldsolver/fs_common.cpp
ykempf: yay! :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada:

ykempf: So that works, thanks for the tip!
Any similar tip regarding accessing the global size/coordinates and the local domain's start so I can reconstruct local-to-global conversions? Or what you suggest for me to be able to access the conversion functions from the loop body? IIRC it was failing with "plain" fsgrid.<function> calls inside the loop body, which had led me to create these variants of `parallel_for`.

ykempf: (pushing later, once the UH CI is fixed)

ykempf: Ah well I can call the fsgrid.* functions before the lambda to get the parameters needed and then work with that inside the lambda, using these parameters and the stencil.i/j/k to compute what I need.

cscjlan: You can pass the individual parameters if you want, but passing the `Coordinates` struct was what we discussed yesterday. I'd say make the lambdas accept a `const Coordinates &coordinates`, and pass in `coordinates` in the `parallel_for`.

cscjlan: Something like

```cpp
// In vlasiator
fsgrid.parallel_for([](int timerId) -> phiprof::Timer { return phiprof::Timer{timerId}; },
                       phiprof::initializeTimer("FS derivatives compute cells"), technical,
                       [=](const fsgrid::Coordinates &coordinates, const fsgrid::FsStencil& stencil, cuint sysBoundaryFlag, cuint sysBoundaryLayer) {
                          calculateDerivatives(perb, moments, dperb, dmoments, stencil, sysBoundaryFlag,
                                               sysBoundaryLayer, doMoments);
                       });

// In fsgrid
template <typename Lambda, typename TimerCallBack, typename T>
   void parallel_for(TimerCallBack timerCallBack, int timerId, std::span<T> technical, Lambda loop_body) {
      // Using raw pointer for localSize;
      // Workaround intel compiler bug in collapsed openmp loops
      // see https://github.com/fmihpc/vlasiator/commit/604c81142729c5025a0073cd5dc64a24882f1675
      const FsIndex_t* localSize = &coordinates.localSize[0];

#pragma omp parallel
      {
         auto timer = timerCallBack(timerId);
#pragma omp for collapse(2)
         for (auto k = 0; k < localSize[2]; k++) {
            for (auto j = 0; j < localSize[1]; j++) {
               for (auto i = 0; i < localSize[0]; i++) {
                  const auto s = makeStencil(i, j, k);
                  const auto& tech = technical[s.ooo()];
                  const auto sysBoundaryFlag = tech.sysBoundaryFlag;
                  const auto sysBoundaryLayer = tech.sysBoundaryLayer;
                  // GPU pointer to coordinates for GPU parallel_for
                  loop_body(coordinates, s, sysBoundaryFlag, sysBoundaryLayer);
               }
            }
         }
         timer.stop(getNumCells(), "Spatial Cells");
      }
   }
```

cscjlan: Just "thinking out loud" here about the GPU version and what does it mean for the stencil & coordinates:

```cpp
template <typename Lambda, typename T>
__global__ void loop_kernel(
    Coordinates *coordinates,
    StencilConstants *stencil_constants,
    std::span<T> technical,
    Lambda loop_body)
{
    const FsIndex tidx = threadIdx.x + blockIdx.x * blockDim.x;
    const FsIndex tidy = threadIdx.y + blockIdx.y * blockDim.y;
    const FsIndex tidz = threadIdx.z + blockIdx.z * blockDim.z;

    const FsIndex stridex = blockDim.x * gridDim.x;
    const FsIndex stridey = blockDim.y * gridDim.y;
    const FsIndex stridez = blockDim.z * gridDim.z;

    for (FsIndex k = tidz; k < coordinates->localSize[2]; k += stridez) {
        for (FsIndex j = tidy; j < coordinates->localSize[1]; j += stridey) {
            for (FsIndex i = tidx; i < coordinates->localSize[0]; i += stridex) {
                const FsStencil s(i, j, k, *stencil_constants);
                const auto& tech = technical[s.ooo()];
                const auto sysBoundaryFlag = tech.sysBoundaryFlag;
                const auto sysBoundaryLayer = tech.sysBoundaryLayer;
                loop_body(*coordinates, s, sysBoundaryFlag, sysBoundaryLayer);
            }
        }
    }
}

// In fsgrid
template <typename Lambda, typename TimerCallBack, typename T>
   void parallel_for(TimerCallBack timerCallBack, int timerId, std::span<T> technical, Lambda loop_body) {
#ifdef CPU_STUFF
    // The CPU loop
#elif defined(GPU_STUFF)
    // Do these timers make sense, as the gpu kernel is async?
    auto timer = timerCallBack(timerId);
    loop_kernel<<<gridSize, blockSize>>>(gpu_coordinates, gpu_stencil_constants, technical, loop_body);
    timer.stop(getNumCells(), "Spatial Cells");
#endif
   }
```

ykempf: Since we'll now be passing `coordinates` we can use that to get e.g. `gridSpacing` in the lambda body when needed instead of having it extracted ahead of the lambda and captured. Does it make a big difference, should I do that change then too?

ykempf:
All right I seem to have botched something(s). However nominally the last 3 commits should have dealt with
- getting rid of the numerous variants of parallel_for_* loop interfaces
- removing RNG inits for perturbed B
- removing capturing gridSpacing by using the newly passed `coordinates`.

ykempf:
Oh yeah forgot to push out the fsgrid commit. :melting_face:

backgroundfield/backgroundfield.cpp
```diff
@@ -20,142 +20,120 @@
...
-         for (FsGridTools::FsIndex_t x = 0; x < localSize[0]; ++x) {
-            for (FsGridTools::FsIndex_t i = 0; i < fsgrids::bgbfield::N_BGB; ++i) {
-               BgBGrid.get(x,y,z)->at(i) = 0;
-            }
-         }
+   fsgrid.parallel_for([](int timerId) -> phiprof::Timer { return phiprof::Timer{timerId}; },
+   phiprof::initializeTimer("setBackgroundFieldToZero"), technical,
+   [=](const fsgrid::Coordinates &coordinates, const fsgrid::FsStencil& stencil, cuint sysBoundaryFlag, cuint sysBoundaryLayer) {
+      for (size_t i = 0; i < bgb[stencil.ooo()].size(); i++) {
+         bgb[stencil.ooo()][i] == 0.0;
```
ykempf: Ooooof thanks HILE compiler!

vlasiator.cpp
```diff
@@ -151,14 +153,16 @@ void computeNewTimeStep(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpi
...
-   for (FsGridTools::FsIndex_t k = 0; k < gridDims[2]; k++) {
-      for (FsGridTools::FsIndex_t j = 0; j < gridDims[1]; j++) {
-         for (FsGridTools::FsIndex_t i = 0; i < gridDims[0]; i++) {
-            fsgrids::technical* cell = technicalGrid.get(i, j, k);
-            if (cell->sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY ||
-               (cell->sysBoundaryLayer == 1 && cell->sysBoundaryFlag != sysboundarytype::NOT_SYSBOUNDARY)) {
-               dtMaxLocal[2] = min(dtMaxLocal[2], cell->maxFsDt);
+   // not performance-critical so no acute need to parallelise (parallel_for)
+   const auto* localSize = &fsgrid.getLocalSize()[0];
+   for (auto k = 0; k < localSize[2]; k++) {
```
ykempf: Note to self – this is one of the spots to look at/implement a reduction tool for.

fieldsolver/ldz_main.cpp
```diff
@@ -38,377 +38,272 @@
...
 
-         std::array<FsGridTools::FsIndex_t, 3>& localSize = technicalGrid.getLocalSize();
-         for(FsGridTools::FsIndex_t z=0; z<localSize[2]; z++) {
-            for(FsGridTools::FsIndex_t y=0; y<localSize[1]; y++) {
-               for(FsGridTools::FsIndex_t x=0; x<localSize[0]; x++) {
-                  fsgrids::technical* cell = technicalGrid.get(x,y,z);
-                  if ( cell->sysBoundaryFlag == sysboundarytype::NOT_SYSBOUNDARY ||
-                        (cell->sysBoundaryLayer == 1 && cell->sysBoundaryFlag != sysboundarytype::NOT_SYSBOUNDARY )) {
-                     dtMaxLocal=min(dtMaxLocal, cell->maxFsDt);
+         // No urgent need to parallelise, see below.
```
ykempf: Note to self – this is another spot where the reduction would be needed.
@cscjlan @trossi should this be fancy enough to take a function (min, max, what else?) or should this just do the min that's needed at the moment?

grid.cpp
```diff
@@ -1195,72 +1196,76 @@ void initializeStencils(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpi
...
-            const dccrg::Types<3>::indices_t  indices = {{(uint64_t)mapIndices[0],(uint64_t)mapIndices[1],(uint64_t)mapIndices[2]}}; //cast to avoid warnings
-            CellID dccrgCellID2 = mpiGrid.get_existing_cell(indices, 0, mpiGrid.mapping.get_maximum_refinement_level());
-            int amrLevel= mpiGrid.get_refinement_level(dccrgCellID2);
-            technicalGrid.get(i, j, k)-> refLevel =amrLevel ;
+void mapRefinement(dccrg::Dccrg<SpatialCell, dccrg::Cartesian_Geometry>& mpiGrid,
+                   fsgrids::technicalspan technical, FieldSolverGrid &fsgrid) {
+   phiprof::Timer timer{"Map Refinement Level to FsGrid"};
+   const auto* localSize = &fsgrid.getLocalSize()[0];
+
+   for (auto k = 0; k < localSize[2]; k++) {
```
ykempf: Here's a lambda-less loop, not touched so far due to having mpiGrid.

ioread.cpp
```diff
@@ -881,67 +914,73 @@ template<unsigned long int N> bool readFsGridVariable(
...
 
       // Assign buffer into fsgrid
-      int index=0;
-      for(FsGridTools::FsIndex_t z=0; z<localSize[2]; z++) {
-         for(FsGridTools::FsIndex_t y=0; y<localSize[1]; y++) {
-            for(FsGridTools::FsIndex_t x=0; x<localSize[0]; x++) {
-               memcpy(targetGrid.get(x,y,z), &buffer[index], N*sizeof(Real));
+      // Could be parallelised (parallel_for) but this is done once at restart so not critical
+      int index = 0;
+      for (auto z = 0; z < localSize[2]; z++) {
```
ykempf: Note to self – TBD

iowrite.cpp
```diff
@@ -983,18 +965,20 @@ bool writeFsGridMetadata(FsGrid< fsgrids::technical, FS_STENCIL_WIDTH> & technic
...
+  xmlAttributes["yperiodic"]=fsgrid.getPeriodic()[1]?"yes":"no";
+  xmlAttributes["zperiodic"]=fsgrid.getPeriodic()[2]?"yes":"no";
 
   if (writeIDs) {
      // Write cell "globalID" numbers, which are just the global array indices.
-     std::vector<FsGridTools::FsIndex_t> globalIds(localSize[0]*localSize[1]*localSize[2]);
+     std::vector<fsgrid::FsSize_t> globalIds(
+         static_cast<fsgrid::FsSize_t>(localSize[0] * localSize[1] * localSize[2]));
      int i=0;
+     // Could be parallelised (parallel_for) but not critical.
```
ykempf: Note to self – TBD

fieldsolver/ldz_magnetic_field.cpp
```diff
@@ -159,115 +154,106 @@ void propagateMagneticField(
...
+      fsgrid.updateGhostCells(perbdt2);
    }
    mpiTimer.stop();
 
    // Propagate B on system boundary/process inner cells
    phiprof::Timer sysBoundaryTimer {sysBoundaryTimerId};
    // L1 pass, gather which faces to solve
    std::vector<std::array<int,4>> L1Solve;
-   #pragma omp parallel
+#pragma omp parallel
```
ykempf: And so here @markusbattarbee @ursg the consensus is to revert part of this to use `parallel_for` as well, right?

fieldsolver/ldz_main.cpp
ykempf: Now I look at it actually I don't think it would be easy to generalise as the current one would do the reduction with the pragma omp, so I'll implement the `parallel_for_min` that's needed at the moment.

ykempf:
The rabbit hole of #1161 convinced me that the testpackage diffs are very unlikely to have arisen from this PR. I will thus proceed with the remaining open things here.

fieldsolver/ldz_main.cpp
ykempf: This was done.

vlasiator.cpp
ykempf: Done.

grid.cpp
ykempf: This I can "fix" by capturing with `[=, &mpiGrid]` otherwise the mpiGrid's communicator is wrecked. Please advise if this should be fixed differently @cscjlan @trossi @ursg.

ykempf: I pushed this now in a version with parallel_for whereas the original was a serial triple-nested loop. *Should work*. #TM

fieldsolver/ldz_magnetic_field.cpp
ykempf: up

grid.cpp
cscjlan: I don't know much about the implementation of the mpiGrid, but does it work directly on the GPU as well? If not, then capturing it doesn't work here. Then the option would be to refactor the stuff that's needed on the GPU away from the mpiGrid to a separate structure, which mpiGrid uses internally, and pass it to the lambda, or not do these computations on the GPU.

iowrite.cpp
```diff
@@ -1717,57 +1672,42 @@ bool writeRestart(
...
-            }
-            return retval;
-         }
-   ));
+   restartReducer.addOperator(
+       new DRO::DataReductionOperatorFsGrid("fg_E", [](const FieldSolverData& fieldSolverData) -> std::vector<Real> {
+          const std::array<fsgrid::FsIndex_t, 3>& gridSize = fieldSolverData.fsgrid.getLocalSize();
+          std::vector<Real> retval(gridSize[0] * gridSize[1] * gridSize[2] * fsgrids::efield::N_EFIELD);
+          int index = 0;
+          for (fsgrid::FsIndex_t z = 0; z < gridSize[2]; z++) {
```
ykempf: Gah. More loops.

fieldsolver/ldz_magnetic_field.cpp
ykempf: Done yesterday.

iowrite.cpp
ykempf: done

ioread.cpp
ykempf: done

iowrite.cpp
ykempf: done

backgroundfield/backgroundfield.cpp
```diff
@@ -20,142 +20,120 @@
...
-         for (FsGridTools::FsIndex_t x = 0; x < localSize[0]; ++x) {
-            for (FsGridTools::FsIndex_t i = 0; i < fsgrids::bgbfield::N_BGB; ++i) {
-               BgBGrid.get(x,y,z)->at(i) = 0;
-            }
-         }
+   fsgrid.parallel_for([](int timerId) -> phiprof::Timer { return phiprof::Timer{timerId}; },
+   phiprof::initializeTimer("setBackgroundFieldToZero"), technical,
+   [=](const fsgrid::Coordinates &coordinates, const fsgrid::FsStencil& stencil, cuint sysBoundaryFlag, cuint sysBoundaryLayer) {
+      for (size_t i = 0; i < bgb[stencil.ooo()].size(); i++) {
+         bgb[stencil.ooo()][i] = 0.0;
```
ursg, line 136: This parallel for has inconsistent indentation with the others above.

datareduction/datareducer.cpp
```diff
@@ -47,317 +44,250 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
-               }
-               return retval;
-         }
-         ));
+      for (auto& c : lowercase) {
+         c = tolower(c);
+      }
+
+      if (P::systemWriteAllDROs || lowercase == "fg_b" ||
+          lowercase == "b") { // Bulk magnetic field at Yee-Lattice locations
```
ursg, line 52: I'm completely ok with most of your autoreformattings in this PR, but breaking all these lines in the datareducer file *actually* makes the whole source file harder to read. Is this something you can semi-easily revert by hand, or should I do that?

datareduction/datareducer.cpp
```diff
@@ -47,317 +44,250 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
+          lowercase == "populations_vg_rho") { // Per-population particle number density
+         for (unsigned int i = 0; i < getObjectWrapper().particleSpecies.size(); i++) {
+            species::Species& species = getObjectWrapper().particleSpecies[i];
             const std::string& pop = species.name;
-            outputReducer->addOperator(new DRO::DataReductionOperatorPopulations<Real>(pop + "/vg_rho", i, offsetof(spatial_cell::Population, RHO), 1));
-            outputReducer->addMetadata(outputReducer->size()-1,"1/m^3","$\\mathrm{m}^{-3}$","$n_\\mathrm{"+pop+"}$","1.0");
+            outputReducer->addOperator(new DRO::DataReductionOperatorPopulations<Real>(
+                pop + "/vg_rho", i, offsetof(spatial_cell::Population, RHO), 1));
+            outputReducer->addMetadata(outputReducer->size() - 1, "1/m^3", "$\\mathrm{m}^{-3}$",
+                                       "$n_\\mathrm{" + pop + "}$", "1.0");
```
ursg, line 253: likewise, breaking these unit lines just makes things unneccesarily hard to read.

datareduction/datareducer.cpp
```diff
@@ -560,1789 +533,1274 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
+         outputReducer->addMetadata(outputReducer->size() - 5, "T/m", "$\\mathrm{T}\\,\\mathrm{m}^{-1}$",
+                                    "$\\Delta B_{Y,\\mathrm{per,vol,vg}} (\\Delta Y)^{-1}$", "1.0");
+         outputReducer->addMetadata(outputReducer->size() - 4, "T/m", "$\\mathrm{T}\\,\\mathrm{m}^{-1}$",
+                                    "$\\Delta B_{Y,\\mathrm{per,vol,vg}} (\\Delta Z)^{-1}$", "1.0");
+         outputReducer->addMetadata(outputReducer->size() - 3, "T/m", "$\\mathrm{T}\\,\\mathrm{m}^{-1}$",
+                                    "$\\Delta B_{Z,\\mathrm{per,vol,vg}} (\\Delta X)^{-1}$", "1.0");
+         outputReducer->addMetadata(outputReducer->size() - 2, "T/m", "$\\mathrm{T}\\,\\mathrm{m}^{-1}$",
+                                    "$\\Delta B_{Z,\\mathrm{per,vol,vg}} (\\Delta Y)^{-1}$", "1.0");
+         outputReducer->addMetadata(outputReducer->size() - 1, "T/m", "$\\mathrm{T}\\,\\mathrm{m}^{-1}$",
+                                    "$\\Delta B_{Z,\\mathrm{per,vol,vg}} (\\Delta Z)^{-1}$", "1.0");
```
ursg, line 909: Also this is unnecessary, IMHO

datareduction/datareducer.cpp
```diff
@@ -2353,1368 +1811,1181 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
+      if (P::systemWriteAllDROs || lowercase == "ig_chi0") {
+         outputReducer->addOperator(new DRO::DataReductionOperatorIonosphereNode(
+             "ig_chi0", [](SBC::SphericalTriGrid& grid) -> std::vector<Real> {
+                std::vector<Real> retval(grid.nodes.size());
 
-                     std::vector<Real> retval(grid.nodes.size());
+                for (uint n = 0; n < grid.nodes.size(); n++) {
+                   Real theta = acos(grid.nodes[n].x[2] / sqrt(grid.nodes[n].x[0] * grid.nodes[n].x[0] +
+                                                               grid.nodes[n].x[1] * grid.nodes[n].x[1] +
+                                                               grid.nodes[n].x[2] * grid.nodes[n].x[2])); // Latitude
```
ursg, line 2288: Here's an exception, that I actually appreciate! :)

datareduction/datareducer.cpp
```diff
@@ -2353,1368 +1811,1181 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
+                       physicalconstants::K_B * grid.nodes[i].electronTemperature() / physicalconstants::CHARGE / 1000;
+
+                   for (int p = 0; p < SBC::productionNumParticleEnergies; p++) {
+                      Real energyparam = (particle_energy[p] - accenergy) / temp_keV; // = E_p / (kB T)
+                      Real deltaE =
+                          (particle_energy[p + 1] - particle_energy[p]) * 1e3 * physicalconstants::CHARGE; // dE in J
+                      retval[i] += grid.nodes[i].parameters[ionosphereParameters::RHON] *
+                                   sqrt(1. / (2. * M_PI * physicalconstants::MASS_ELECTRON)) * particle_energy[p] /
+                                   temp_keV / sqrt(temp_keV * 1e3 * physicalconstants::CHARGE) * deltaE *
+                                   exp(-energyparam); // Flux 1/m^2/s
```
ursg, line 2576: This has now become very messy, compared to before.

datareduction/datareducer.cpp
```diff
@@ -2353,1368 +1811,1181 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
+                       physicalconstants::K_B * grid.nodes[i].electronTemperature() / physicalconstants::CHARGE / 1000;
+
+                   for (int p = 0; p < SBC::productionNumParticleEnergies; p++) {
+                      Real energyparam = (particle_energy[p] - accenergy) / temp_keV; // = E_p / (kB T)
+                      Real deltaE =
+                          (particle_energy[p + 1] - particle_energy[p]) * 1e3 * physicalconstants::CHARGE; // dE in J
+                      numberFlux += grid.nodes[i].parameters[ionosphereParameters::RHON] *
+                                    sqrt(1. / (2. * M_PI * physicalconstants::MASS_ELECTRON)) * particle_energy[p] /
+                                    temp_keV / sqrt(temp_keV * 1e3 * physicalconstants::CHARGE) * deltaE *
+                                    exp(-energyparam); // Flux 1/m^2/s
```
ursg, line 2619: Same here.

fieldsolver/derivatives.cpp
```diff
@@ -809,43 +581,46 @@ void calculateScaledDeltas(
          dU = std::max(fabs(myU - otherU) / maxU, dU);
          dBsq = std::max(deltaBsq / (2 * physicalconstants::MU_0 * maxU), dBsq);
          if (myRho > EPS) {
-            dPsq = std::max((pow(myP[0] - otherP[0], 2) + pow(myP[1] - otherP[1], 2) + pow(myP[2] - otherP[2], 2)) / (2 * myRho * maxU), dPsq);
+            dPsq = std::max((pow(myP[0] - otherP[0], 2) + pow(myP[1] - otherP[1], 2) + pow(myP[2] - otherP[2], 2)) /
+                                (2 * myRho * maxU),
+                            dPsq);
```
ursg, line 587: Here's another particularly unpretty formatting example.

fieldsolver/fs_common.cpp
```diff
@@ -126,77 +109,99 @@ void reconstructionCoefficients(
...
+
+      perturbedResult[Rec::c_xx] =
+          HALF * (der_i1j1k2[fsgrids::dperb::dPERBzdxx] + der_i1j1k1[fsgrids::dperb::dPERBzdxx]);
+      perturbedResult[Rec::c_xy] =
+          HALF * (der_i1j1k2[fsgrids::dperb::dPERBzdxy] + der_i1j1k1[fsgrids::dperb::dPERBzdxy]);
+      perturbedResult[Rec::c_yy] =
+          HALF * (der_i1j1k2[fsgrids::dperb::dPERBzdyy] + der_i1j1k1[fsgrids::dperb::dPERBzdyy]);
+      perturbedResult[Rec::c_xxz] = (der_i1j1k2[fsgrids::dperb::dPERBzdxx] - der_i1j1k1[fsgrids::dperb::dPERBzdxx]);
+      perturbedResult[Rec::c_xyz] = (der_i1j1k2[fsgrids::dperb::dPERBzdxy] - der_i1j1k1[fsgrids::dperb::dPERBzdxy]);
+      perturbedResult[Rec::c_yyz] = (der_i1j1k2[fsgrids::dperb::dPERBzdyy] - der_i1j1k1[fsgrids::dperb::dPERBzdyy]);
```
ursg, line 140: And here, only ~50% of the lines have been broken now.

ursg: (This is anyway code that probably nobody will ever read through with any intent or detail... so why isn't this just autogenerated anyway?)

fieldsolver/fs_common.h
```diff
@@ -62,26 +62,22 @@ static creal EPS = 1.0e-30;
...
+                     fsgrids::egradpespan egradpe,
+                     fsgrids::egradpespan egradpedt2,
+                     fsgrids::momentsspan moments,
+                     fsgrids::momentsspan momentsdt2,
+                     fsgrids::dperbspan dperb,
+                     fsgrids::dmomentsspan dmoments,
+                     fsgrids::dmomentsspan dmomentsdt2,
+                     fsgrids::bgbspan bgb,
+                     fsgrids::volspan vol,
+                     fsgrids::technicalspan technical, FieldSolverGrid &fsgrid, SysBoundary& sysBoundaries,
```
ursg, line 79: Why aren't all of these spans just collected together into one proper "fsgrids" object by now? I mean, it's great that they live in their own namespace, but for all intents and purposes, they might as well be handed back and forth in a single pointer, if they were sharing a common struct.

ursg: (Also, the fieldsolver functions could then be methods of that object)

fieldsolver/ldz_hall.cpp
ursg, line 1: Hooray, my least favourite source file got even messier now. :)

ursg (approved):
Incomplete review so far, I'll go through some more later. But functional approval, of course.
I'm not fully happy with the style in a number of places, but those are things that can... settle over time.

fieldsolver/fs_common.h
ursg: And now I saw further down that FsGridSolverData *does* exist. So how about passing that one here?

datareduction/datareducer.cpp
ykempf: These were all done automatically by Juhana's software, according to the repo's style file. In things I did manually, I tried to stick to some sort of "what was done so far in the team", but I did not revert Juhana's automatic formatting. I can provide the style file I came up with when trying to improve it. I would really strongly suggest from this side of the bay that pliiis, deploy some CI tool doing all formatting for everyone, or at least a style file that works for all. However I stumbled upon options that changed in recentish versions of clang formatting, and/or git-clang-format was not available on major platforms we use (LUMI), so I think the more robust way would be a CI pre-commit or similar feature holding the ground truth and making everything consistent. At the price of having to create that ground truth file...

fieldsolver/ldz_hall.cpp
ykempf: Show it some love, come on! I assume the hyperresistivity one will not be much different. :)

fieldsolver/fs_common.h
ykempf: I think so far the idea was to pass the needed spans down the chain, which can transparently point to device or host memory in the end.

ykempf:
By my eye, most if not all of the lines you disliked in terms of formatting are due to the max character width setting in the config file.

fieldsolver/fs_common.h
cscjlan: IIRC I kept the function signatures on the vlasiator side as close to the original as possible if it wasn't necessary to change them. But definitely one could just pass `FieldSolverData` here.

-----------------------

I would keep the functionality separate from the `FieldSolverData` object, i.e. not implement any methods for it. First of  all, the `FieldSolverData` doesn't even own the data, it's just a wrapper over multiple `std::span`s -- i.e. pointers & lengths -- the original purpose of which was to make the tens of lambda expressions simpler in `datareducer.cpp` to simplifiy changing the underlying objects, because they can be changed in one place.

Secondly keeping data and functionality separated was one of the major points of the refactoring on the FsGrid side. This way e.g. the initialization of `FsData` and `FieldSolverGrid` can happen separately from each other and there are no partially initialized objects: "on this line the MPI initialization has happened for object A, but the data is garbage, but after these lines also the data has been initialized for object A". With separation, `FieldSolverGrid` is a fully functioning object after construction and likewise for each `FsData` (albeit they're full of zeroes).

This separation also simplifies e.g. testing of `FsData`: when it's implemented to work on the GPU, testing it can be done without any references to MPI/OpenMP, since it neither knows nor cares about those.

datareduction/datareducer.cpp
cscjlan: As @ykempf mentioned in a comment down the line, many of these could be "reverted" quite easily by increasing the allowed line width.

ursg: "They're more like guidelines anyway".

I'm all in favour of allowing for longer lines here, because the readability *actually* benefits from the better formatting in many cases.

fieldsolver/fs_common.h
ursg: I see. Ok then, given what we have learned about memory handling in hybrid architectures, it makes a lot of sense to handle the spans individually and *carefully*.

fieldsolver/fs_common.cpp
ykempf: This is inherited directly from Arto's hand-crafted first implementation out of the paper, I/we only changed the arrays, namespaces etc in hiostory but the indices are still as typed by Him in the olden days.

datareduction/datareducer.cpp
ykempf: So! Set up some way or other for this to be rolled out automatically, fix the formatting options to your desire, and off we go! Some options allow or disallow e.g. one parameter per line when breaking, but also some options were recent enough not to be available on all platforms, so I suppose having one central reference instance will be easiest to maintain (I *tried* installing git-clang-format on some machine *without* having to build the whole of LLVM and failed with that...).

ykempf:
OK I fixed merge conflicts but I know that I broke at least one of the new features so TBC.

PR #1100, opened 2025-02-17
GPU improved launch parameters, memory usage, etc.

markusbattarbee:
Differentiates between Ukko DGX (V100 nodes) and A100 nodes. Updates to up-to-date modules. DGX has much better queue times, but performance is still a bit too slow for CI (about 1h 20 mins). A100 nodes suffer, in addition to terrible queue times, from terrible core/thread/GPU placement, awaiting fixes from IT4sci.

GPU initial memory buffers were changed to scale with WID value so WID=8 binaries don't hog up crazy amounts of memory. Also fixed a mis-indexing in GPU memory reporting. Some tests still report more GPU memory use than expected, pending investigation.

Cleans away some broken timers from GPU branch.

Improves readability of batch operation launch parameters (use of `std::max`)

Improves launch parameters for GPU kernels, especially when not using warp accessors (which is the default).


markusbattarbee:
Now it would be possible to go back to a single ukko CUDA makefile, compiling and linking the binary to support both V100 and A100 cards (architectures 70 and 80) but compilation will take longer, and compiling for an older architecture should be safe. Thoughts?

spatial_cells/velocity_block_container.h
```diff
@@ -760,7 +793,7 @@ namespace vmesh {
       return cachedSize*WID3*sizeof(Realf) + cachedSize*BlockParams::N_VELOCITY_BLOCK_PARAMS*sizeof(Real);
```
ykempf: In my branch I removed this WID3* too.

ykempf (approved):
If it otherwise works... But apart from this one comment I have not looked in detail.

ykempf:
And here too the output file is truncated mid-Ionos

ykempf (approved):
All right I looked at the testpackage changes too they do look good as well.
Would we save file size by using more tabs than spaces? Though a quick scroll shows me we do have a lot of \t already.

ursg (approved):
Heh, I read through the complete diff now and found *nothing* to comment. That has never happened before!

Obvious red flag. But I can't help but approve. 😃

PR #1106, opened 2025-03-04
Improve documentation of EgradPe term in field solver code.

markusbattarbee:


fieldsolver/derivatives.cpp
```diff
@@ -67,10 +67,10 @@ void calculateDerivatives(
    cuint sysBoundaryFlag  = technicalGrid.get(i,j,k)->sysBoundaryFlag;
    cuint sysBoundaryLayer = technicalGrid.get(i,j,k)->sysBoundaryLayer;
 
-   // Constants for electron pressure derivatives
-   // Upstream pressure
-   Real Peupstream = Parameters::electronTemperature * Parameters::electronDensity * physicalconstants::K_B;
-   Real Peconst = Peupstream * pow(Parameters::electronDensity, -Parameters::electronPTindex);
+   // Constants for electron pressure derivatives, see also ldz_gradpe.cpp
+   // Calculate anchor point constants: First the pressure, then a derived value.
```
alhom: Since the PR is about improved documentation... :)
The phrase "then a derived value." should IMO rather be "then the derived constant", since it is a singular constant in the simulation and referred to as "that constant" in ldz_gradpe.cpp commentary.

PR #1107, opened 2025-03-07
Hile makefiles

markusbattarbee:
I split up the library building into first a file fetch script, then an actual build script, so one does not need to set up tunneling to compute nodes.

This also includes scripts to build and run on hile as well as a crucial ionosphere timer fix.

On a single HILE-C node (with 128 cores), testpackage completes now in about 23 minutes which is quite nice. I'm generating reference data on Kappa as well now, so we could consider moving CI over to Hile from Carrington.

Hile-G compiles and runs but I haven't yet sorted out performance with that.


ursg (approved):
Cool. I like the fetch and build split, too.

My first thought reaction actually was "should we just absorb them into the Makefile as separate targets?".
But then that's probably for another PR.

PR #1110, opened 2025-03-28
Alternative Fieldsolver magnetic field boundary cell parallel approach

markusbattarbee:
Do a "quick" loop of gathering which cells/faces need computing, then do a separate parallel loop to do the actual computation.

### DIFF
--- a/fieldsolver/ldz_electric_field.cpp
+++ b/fieldsolver/ldz_electric_field.cpp
@@ -1693,7 +1693,7 @@ void calculateUpwindedElectricFieldSimple(
    
    mpiTimer.stop();
    
-   // Calculate upwinded electric field on inner cells
+   // Calculate upwinded electric field
    #pragma omp parallel
    {
       phiprof::Timer computeTimer {computeTimerID};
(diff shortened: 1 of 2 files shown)


valentin-seitz:
I applied these changes to my `FIC` branch (https://github.com/valentin-seitz/vlasiator/commit/20f4955b06901f61867584faa535d7ea1612ce87) and just ran the production case you shared with me. 
For the 4. timestep after the restart i see a ~10% speedup of the fieldsolver in the production case. 

I also checked on a smaller example using the example Magnetosphere3D input file and saw on 32 Nodes a similar ~10% speedup in the `Propagate Fields` region. 

For the production case the maximum time across processes dropped from `3.59s` to `3.25s`.

Note, that most of the speedup was actually gained in the iterative field-solver calls, where this PR drastically reduces the "combing" effect we saw before: 

## Trace of 12 fieldsolver iterations on `FIC` branch (https://github.com/valentin-seitz/vlasiator/commit/85e7622f8878ab8e8ece7a172c499a75df626507)
**Total Duration of 12 Iterations**: `2.24s`

image



## Trace of 12 fieldsolver iterations on `FIC` with the PR applied (https://github.com/valentin-seitz/vlasiator/commit/20f4955b06901f61867584faa535d7ea1612ce87)

**Total Duration of 12 Iterations**: `1.75s`
image


**How to read the trace**: 
X-Axis: Time
Y-Axis: Processes and Threads (7000 MPI processes and 8 OpenMP Threads each)
Color: How much time is spent in "Useful" computation between calls to parallel Runtimes (MPI or OpenMP). Black means, that the CPU is not in user-code, but in runtime code probably spin-locking.

ursg (approved):
It's not the most elegant code, pulling the loops apart like that, but it does its job well, so what's not to like!

PR #1112, opened 2025-04-03
Reorder acceleration preparations so that always-on-cpu tasks are sep…

markusbattarbee:
Preparation for moving to batch-style GPU acceleration handling. Using PR for CI.

### DIFF
--- a/vlasovsolver/cpu_acc_semilag.hpp
+++ b/vlasovsolver/cpu_acc_semilag.hpp
@@ -26,7 +26,6 @@
 #include "../common.h"
 #include "../spatial_cells/spatial_cell_wrapper.hpp"
 
-using namespace std;
 using namespace spatial_cell;
 
 void cpu_accelerate_cells(
--- a/vlasovsolver/cpu_acc_semilag.hpp
+++ b/vlasovsolver/cpu_acc_semilag.hpp
@@ -26,12 +26,17 @@
 #include "../common.h"
 #include "../spatial_cells/spatial_cell_wrapper.hpp"
 
-void cpu_accelerate_cell(
-        spatial_cell::SpatialCell* spatial_cell,
-        const uint popID,
-        const uint map_order,
-        const Real& dt,
-        int intersections_id,
-        int mappings_id);
+using namespace std;
+using namespace spatial_cell;
+
+void cpu_accelerate_cells(
+   dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
+   const std::vector<CellID>& acceleratedCells,
+   const uint popID,
+   const uint map_order);
+
+void cpu_accelerate_cell(SpatialCell* spatial_cell,
+                         const uint popID,
+                         const uint map_order);
 
 #endif
--- a/vlasovsolver/cpu_acc_semilag.hpp
+++ b/vlasovsolver/cpu_acc_semilag.hpp
@@ -26,16 +26,12 @@
 #include "../common.h"
 #include "../spatial_cells/spatial_cell_wrapper.hpp"
 
-void prepareAccelerateCell(spatial_cell::SpatialCell* spatial_cell, const uint popID);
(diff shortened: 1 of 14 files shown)


vlasovsolver/cpu_acc_semilag.hpp
```diff
@@ -26,16 +26,17 @@
 #include "../common.h"
 #include "../spatial_cells/spatial_cell_wrapper.hpp"
 
-void prepareAccelerateCell(spatial_cell::SpatialCell* spatial_cell, const uint popID);
-uint getAccelerationSubcycles(spatial_cell::SpatialCell* spatial_cell, Real dt, const uint popID);
+using namespace std;
```
ursg: This is a bit of bad style, to have the std namespace included in a header file that potentially gets used somewhere else.

vlasovsolver/gpu_acc_semilag.hpp
```diff
@@ -26,14 +26,17 @@
 #include "../common.h"
 #include "../spatial_cells/spatial_cell_wrapper.hpp"
 
-void prepareAccelerateCell(spatial_cell::SpatialCell* spatial_cell, const uint popID);
-uint getAccelerationSubcycles(spatial_cell::SpatialCell* spatial_cell, Real dt, const uint popID);
+using namespace std;
```
ursg: here too, the namespace std inclusion is bad style

ursg (approved):
The actual core of this PR is a bit buried in all of those formatting and documentation changes, but when I found it, it looked solid.

Two small coding style remarks, otherwise thumbs up!

alhom (approved):
I'm a bit cautious about having the temporary and per-population intersections stored externally from the acceleration phase where they are used, but I assume there is a good reason for it!

I would still add an explicit comment to spatial_cell_cpu.hpp and _gpu.hpp next to the declarations that these are for temporary storage of intermediate values and used in these files and/or functions.

markusbattarbee:
> I'm a bit cautious about having the temporary and per-population intersections stored externally from the acceleration phase where they are used, but I assume there is a good reason for it!

Yeah, I might as well explain this. For the GPU branch, we'll want to perform these cpu-only quick calculations all in advance, then start on the vlasov side of the operations by doing one small subset at a time, for all spatial cells at once. Thus, there needs to be some temporary storage for these intersections. For the actual GPU operations, there will be a buffer where they'll be stored so they can be easily uploaded to device, but I thought that making a separate buffer for CPU-only does not make sense. 

One could of course store these intersections and dt in the population variable, but as we're still performing acceleration on only one population at a time, I thought it was easier to place it here (and in fact it will be removed from the `spatial_cell_gpu.hpp` in the future. But if placing the variables in population would be seen as better, then can do that.

> I would still add an explicit comment to spatial_cell_cpu.hpp and _gpu.hpp next to the declarations that these are for temporary storage of intermediate values and used in these files and/or functions.

Sure, can elaborate.

What I was wanting to hear from you @alhom was that hopefully this won't totally torpedo timeclasses?

alhom:
Now for _that_ there needs to be a bit of adaptation, but that is to be expected. Since there are, occasionally, multiple VDFs of the same population being accelerated within the same cell, and at various dts.. yes, a bit of an issue. But, it is easy to account for by... having even more temporary storage! One set of intersections per requested dt, and we're constructing these acceleration-payloads already ahead of the intersection calculations, so I don't see major issues.

markusbattarbee:
> Now for _that_ there needs to be a bit of adaptation, but that is to be expected. Since there are, occasionally, multiple VDFs of the same population being accelerated within the same cell, and at various dts.. yes, a bit of an issue. But, it is easy to account for by... having even more temporary storage! One set of intersections per requested dt, and we're constructing these acceleration-payloads already ahead of the intersection calculations, so I don't see major issues.

OK, but those will then be... different population IDs or what? I can perfectly well place those temporary variables in the population container instead, if that would help.

alhom:
Population containers would work well with that,  indeed!

alhom (approved):

PR #1113, opened 2025-04-07
Spell checker for cfg file

kstppd:
QOL thingy I had built a couple months ago.  When an input in the cfg is mistyped we calculate the Levenshtein distance with a list of known variables and print a suggestion to the user.


datareduction/datareducer.cpp
```diff
@@ -3695,7 +3738,14 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
       int myRank;
       MPI_Comm_rank(MPI_COMM_WORLD,&myRank);
       if (myRank == MASTER_RANK) {
-         std::cerr << __FILE__ << ":" << __LINE__ << ": The diagnostic variable " << *it << " is not defined." << std::endl;
+         if (const auto best_match = spell_check(*it)) {
```
markusbattarbee, line 3741: This uses now an incorrect list of output variables. Here you should check only for accepted diagnostic outputs.

markusbattarbee (changes requested):
So now there's duplication of accepted output variables, here and in 
https://github.com/fmihpc/vlasiator/blob/6780dce9404fee8deb758b6722783dfc097da2d0/parameters.cpp#L388
up to line 455. These lists should be somehow merged - a better way might be to use parameters.cpp to read this new header file (`variable_dictionary.h`) and use that list - but then that dict should be amended to also have a list of up-to-date diagnostic output variables. Thought that needs to be done anyway so that the heuristic suggestion for diagnostic output is valid.

In fact, what you could do for diagnostic output on line 3745 or so is to see if there's a zero-diff match in output variables but not in diagnostic variables, and then let the user know that output is not supported as a diagnostic one. Only if no output variables are a zero-diff match, try to suggest a better spelling from the diagnostic variables list.

kstppd:
Ok so there are valid points here as to why this needs some fixing and I also do not like the whole approach with the extra dictionary. I may find a better way to incorporate this so I will leave this open for now.

datareduction/datareducer.cpp
kstppd: True!

PR #1115, opened 2025-04-15
Return to considering has_content blocks of spatial neighbours at all…

markusbattarbee:
Reverts #919  as it was found to hinder foreshock formation

### DIFF
--- a/grid.cpp
+++ b/grid.cpp
@@ -761,20 +761,18 @@ bool adjustVelocityBlocks(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& m
    // Batch call
    update_velocity_block_content_lists(mpiGrid,cells,popID);
 
-   if (doPrepareToReceiveBlocks) {
-      // We are in the last substep of acceleration, so need to account for neighbours
-      phiprof::Timer transferTimer {"Transfer with_content_list", {"MPI"}};
-      SpatialCell::set_mpi_transfer_type(Transfer::VEL_BLOCK_WITH_CONTENT_STAGE1 );
-      mpiGrid.update_copies_of_remote_neighbors(Neighborhoods::NEAREST);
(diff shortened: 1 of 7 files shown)


ursg:
Is this now an expected outcome of this change, that the flux function outputs show diffs?

markusbattarbee:
> Is this now an expected outcome of this change, that the flux function outputs show diffs?

Well, polar_small output is indeed different, so not surprising as such. Previously for polar_small we had diffs only in the diagonal/offdiagonal pressure components, but now it reads:
```
-----------
ref-time | new-time | speedup |
--------------------------------
20.33      21.41      0.949556
----------
variable                                     | absolute diff | relative diff |
----------
Comparing file /wrk-vakka/group/spacephysics/proj/CI/turso-carrington-mjalho/_work/vlasiator/vlasiator/testpackage/run_2025.04.15_11.02.57/Magnetosphere_polar_small/bulk.0000001.vlsv against reference
proton/vg_rho_0                                1.82            1.13e-06
proton/vg_v_0                                  1.11            1.47e-06
proton/vg_v_1                                  0.375           8.99e-05
proton/vg_v_2                                  0.542           6.46e-05
fg_b_0                                         1.72e-15        8.35e-09
fg_b_1                                         6.86e-16        5.58e-05
fg_b_2                                         1.91e-15        3.62e-11
fg_e_0                                         8.51e-10        8.76e-05
fg_e_1                                         4.14e-09        1.17e-06
fg_e_2                                         7.94e-10        6.48e-05
```
So small relative diffs but diffs still. Not surprising, as the polar run does have small Alfvén wings forming and that's where this change is seen - or even moreso because polar_small is the only non-single-cell test which has acceleration subcycling active. (Should we activate it for M3D and Iono3D?)

ursg (approved):

PR #1117, opened 2025-04-22
Fix generate ref data on Carrington

markusbattarbee:
The basic Carrington job script had recently been updated to default to comparing against reference data instead of creating new data. This updates the github workflow for generating new data to also use `sed` on `small_test_carrington.sh` for creating new data instead of comparing against existing data.


ursg (approved):
Ah, interesting omission. Yep, I suspect this works!

PR #1118, opened 2025-04-30
prevent VBC shrink on AMD-G (still not working), switch lumi_hipcc to…

markusbattarbee:
… use library building script. Also squelch ukko_dgx warnings.


ursg (approved):
Cool. Good to move away from potentially-stale system library piles and go for the build script on more platforms.

This of course doesn't make those scripts any prettier. But getting a nicer solution for that shall be a matter to develop somewhere else than this PR.

PR #1119, opened 2025-05-06
Fix pop_1dmuspace output reducer which was corrupting some testpackag…

markusbattarbee:
…e data


ursg (approved):
Heh, was that a *this capture falling out of scope? I'm surprised this ever worked at all!

markusbattarbee:
> Heh, was that a *this capture falling out of scope? I'm surprised this ever worked at all!

Nah, that was just issuing a warning and only when I implemented the object member `nBins`. The issue was the lack of `std::max(0,std::min(muindex,nBins_lambda-1));` - that led to mis-indexing and data corruption. Arithmetically it should never overflow, but apparently floating point implementations can sometimes round it up to `nBins_lambda==nBins` anyway.

PR #1120, opened 2025-05-06
Add CI action for generating reference testpackage data on Ukko DGX

markusbattarbee:


ursg:
Fair enough. But this needs to go into master, too, otherwise the run trigger won't be visible, right?

markusbattarbee:
> Fair enough. But this needs to go into master, too, otherwise the run trigger won't be visible, right?

Yes, won't be available until then - but might as well have it ready for the next release.

PR #1124, opened 2025-06-03
Fix gpu moments

markusbattarbee:
Changes introduced by #1049 hadn't been properly incorporated into GPU moments calculation. Other minor QoL improvements.


ursg (approved):
That ```initialCompute``` variable name is a bit unclear. Sure, it implies that stuff gets calculated differntly on first round, but a little explanatory comment in the function definition would help clear things up.

Otherwise good!

PR #1125, opened 2025-06-04
Rewritten Acceleration solver for GPUs

markusbattarbee:
This PR introduces a completely re-written acceleration solver for GPUs.
Instead of sorting blocks and constructing columns serially from that list, this approach utilizes a "probe cube" into which active blocks are inserted, with that cube then flattened and efficient GPU scan algorithms used to allow parallel computation of all columnsets. Acceleration is now computed in chunks (with the chunk size defined as a config parameter) where larger chunks allow for greater parallelism but require more memory for temporary buffers. This chunking approach replaces performing accelereation in several concurrent streams, and thus better phiprof timers have been introduced for the chunking approach.

There are some parallelism improvements which could be done to column construction, but profiling suggests the time spent in those parts is miniscule and thus not worth putting effort into at this time.

The bank conflict avoidance scheme used in the scan kernel can issue warnings during compilation on some hardware but was seen to operate correctly nevertheless on NVIDIA hardware. On AMD hardware (or rather, when using HIP) it is deactivated through pragmas.

Future tasks include implementing better and cleaned up memory management, but these were not included in this PR. 

Also includes #1124


markusbattarbee:
Output from Mahti TP run
[slurm-4620717.txt](https://github.com/user-attachments/files/20590763/slurm-4620717.txt)

arch/gpu_base.hpp
```diff
@@ -91,91 +94,166 @@ void gpu_trans_deallocate();
...
+      //    + columnNumBlocks.capacity() * sizeof(uint)
+      //    + setColumnOffsets.capacity() * sizeof(uint)
+      //    + setNumColumns.capacity() * sizeof(uint)
+      //    + minBlockK.capacity() * sizeof(int)
+      //    + maxBlockK.capacity() * sizeof(int)
+      //    + kBegin.capacity() * sizeof(int)
+      //    + i.capacity() * sizeof(int)
+      //    + j.capacity() * sizeof(int)
+      //    + 4 * sizeof(split::SplitVector<uint>)
+      //    + 5 * sizeof(split::SplitVector<int>);
```
ykempf: This still needed?

spatial_cells/block_adjust_gpu.cpp
```diff
@@ -754,4 +614,266 @@ void adjust_velocity_blocks_in_cells(
...
+   const uint popID
+   ) {
+
+   const uint nCells = cellsToAdjust.size();
+   if (nCells == 0) {
+      return;
+   }
+   const gpuStream_t baseStream = gpu_getStream();
+
+   // Resizes are faster this way with larger grid and single thread
```
ykempf: I trust you tested, but it's probably going to depend massively on the system simulated and be very fickle. And there'll be other kernels. But should there be some interface to allow the user to set these at runtime without recompiling? Say to go through parameters (semi-)automatically and determine useful grid and block sizes for a given setup? One thing I noticed in a recent GPU course I took was to not think of these parameters too much as a function of actual problem size/cell count etc but in a more generic way...

spatial_cells/block_adjust_gpu.cpp
```diff
@@ -754,4 +614,266 @@ void adjust_velocity_blocks_in_cells(
...
+   out_largestBlocksToChange = largestBlocksToChange;
+   out_largestBlocksBeforeOrAfter = largestBlocksBeforeOrAfter;
+
+   // Do we actually have any changes to perform?
+   if (largestBlocksToChange > 0) {
+      phiprof::Timer addRemoveKernelTimer {"GPU batch add and remove blocks kernel"};
+      // Third argument specifies the number of bytes in *shared memory* that is
+      // dynamically allocated per block for this call in addition to the statically allocated memory.
+      dim3 grid_addremove(largestBlocksToChange,nCells,1);
+      batch_update_velocity_blocks_kernel<<<grid_addremove, WID3, 0, baseStream>>> (
```
ykempf: In a similar vein as above, for the sake of future optimisation hunters, any comments with wisdom as to the choices of grids and blocks would be very welcome. Now here it is kind of self-explanatory I guess, but nevertheless.

vlasovsolver/gpu_acc_map.cpp
```diff
@@ -20,70 +20,510 @@
...
-#include "cpu_1d_ppm.hpp"
-#include "cpu_1d_plm.hpp"
-
-#ifdef DEBUG_VLASIATOR
-   #ifndef DEBUG_ACC
-   #define DEBUG_ACC
-   #endif
+#include "../spatial_cells/block_adjust_gpu.hpp"
+
+/* Fills the target probe block with the invalid value for vmesh::LocalID
```
ykempf: Use doxygen format when comments are obviously in spots for these? :)

vlasovsolver/gpu_acc_map.cpp
```diff
@@ -20,70 +20,510 @@
...
+   const uint cellOffset = parallelOffsetIndex + cumulativeOffset;
+   vmesh::VelocityBlockContainer* blockContainer = blockContainers[cellOffset];
+   const size_t VBC_size = blockContainer->size() * WID3;
+   Realf *blockData = blockContainer->getData();
+
+   const size_t ind = blockIdx.x * blockDim.x + threadIdx.x;
+   for (size_t i = ind; i < VBC_size; i += gridDim.x * blockDim.x) {
+      blockData[i] = 0;
+   }
+}
```
ykempf: Related to my comments above about kernel launch parameters – if (obviously here, *since*) kernel launch parameters (grid, block parameters) are intimately tied to the strategy used in the kernel, comments at the "launch site"  warning not to change blindly, and matching comments at/in the kernel will help future readers.

This does look like some optimised versions of certain exercises in certain upcoming summer school, but if someone's fancy is tickled to rejig that for better memory access or whatever, they'd better do that carefully and it'll be easier with some guides :).

ykempf (commented):
Some comments, ole hyvä. Do with them what you like. :)

arch/gpu_base.cpp
```diff
@@ -53,24 +53,16 @@ Real *host_returnReal[MAXCPUTHREADS];
...
-vmesh::GlobalID *gpu_GIDlist[MAXCPUTHREADS];
-vmesh::LocalID *gpu_LIDlist[MAXCPUTHREADS];
-vmesh::GlobalID *gpu_BlocksID_mapped[MAXCPUTHREADS];
-vmesh::GlobalID *gpu_BlocksID_mapped_sorted[MAXCPUTHREADS];
-vmesh::LocalID *gpu_LIDlist_unsorted[MAXCPUTHREADS];
-vmesh::LocalID *gpu_columnNBlocks[MAXCPUTHREADS] = {0};
-vmesh::GlobalID *invalidGIDpointer = 0;
-void *gpu_RadixSortTemp[MAXCPUTHREADS] = {0};
-size_t gpu_acc_RadixSortTempSize[MAXCPUTHREADS] = {0};
+ColumnOffsets *host_columnOffsetData = NULL, *dev_columnOffsetData = NULL;
```
ursg: Fundamentally and code-structure wise, this doesn't seem like the right spot to declare the individual solver data structures.

Purely from a separation-of-concerns point of view, shouldn't the arch/ subdirectory contain only high-level computation abstractions, and the solvers' datastructures should be declared in their corresponding header files?

(Of course that'll be a large-ish refactor and maybe not fit into this PR's scope)

arch/gpu_base.cpp
```diff
@@ -282,7 +269,9 @@ __host__ int gpu_getDevice() {
    return device;
 }
 
-
+__host__ uint gpu_getAllocationCount() {
+   return allocationCount;
+}
 /* Memory reporting function
  */
 int gpu_reportMemory(const size_t local_cells_capacity, const size_t ghost_cells_capacity,
```
ursg, line 269: Related to the above comment, it of course makes sense to have the datastructures declared here, for this function to be able to actually sum their sizes.
But that, in itself, is of course an inelegant monstrosity.

Ah well, such is GPU life.

arch/gpu_base.cpp
```diff
@@ -322,20 +307,17 @@ int gpu_reportMemory(const size_t local_cells_capacity, const size_t ghost_cells
       + 2 * sizeof(Hashinator::Hashmap<vmesh::GlobalID,vmesh::LocalID>*) // dev_allMaps
       + 2 * sizeof(split::SplitVector<vmesh::GlobalID>*) // dev_vbwcl_vec, dev_lists_with_replace_new
       + 3 * sizeof(split::SplitVector<Hashinator::hash_pair<vmesh::GlobalID,vmesh::LocalID>>*) // dev_lists_delete, dev_lists_to_replace, dev_lists_with_replace_old
-      + 5 * sizeof(vmesh::LocalID) // dev_contentSizes
+      + 6 * sizeof(vmesh::LocalID) // dev_nWithContent, dev_nBefore, dev_nAfter, dev_nBlocksToChange, dev_resizeSuccess, dev_overflownElements
```
ursg: These hardcoded numbers give me headaches. If we were to wrap all the pointers that the acceleration solver needs into a struct, and give it its own method to report its size, we'd already have a point of failure less for future code changes.

arch/gpu_base.cpp
```diff
@@ -401,83 +383,93 @@ int gpu_reportMemory(const size_t local_cells_capacity, const size_t ghost_cells
...
+      std::array<uint, 3> s = {c0,c1,c2};
+      std::sort(s.begin(), s.end());
+      const size_t probeCubeExtentsFull = s[0]*s[1]*s[2];
+      size_t probeCubeExtentsFlat = s[1]*s[2];
+      probeCubeExtentsFlat = 2*Hashinator::defaults::MAX_BLOCKSIZE * (1 + ((probeCubeExtentsFlat - 1) / (2*Hashinator::defaults::MAX_BLOCKSIZE)));
+      blockDataAllocation = std::max(blockDataAllocation,probeCubeExtentsFull*sizeof(vmesh::LocalID) + probeCubeExtentsFlat*GPU_PROBEFLAT_N*sizeof(vmesh::LocalID));
+   }
+   // Mallocs should be in increments of 256 bytes. WID3 is at least 64, and len(Realf) is at least 4, so this is true. Still, ensure that
+   // probe cube managing does not break this.
+   blockDataAllocation = (1 + ((blockDataAllocation - 1) / 256)) * 256;
```
ursg, line 455: Also, I suspect this is a magic number that may change in future hardware generations?

arch/gpu_base.cpp
```diff
@@ -560,93 +576,69 @@ __host__ void gpu_batch_deallocate(bool first, bool second) {
...
-   const uint c2 = (*vmesh::getMeshWrapper()->velocityMeshes)[0].gridLength[2];
-   std::array<uint, 3> s = {c0,c1,c2};
-   std::sort(s.begin(), s.end());
-   gpu_acc_columnContainerSize = s[1]*s[2];
-   CHK_ERR( gpuMallocAsync((void**)&gpu_columnNBlocks[cpuThreadID], gpu_acc_columnContainerSize*sizeof(vmesh::LocalID), stream) );
-}
+   uint columnAllocationCount;
+   if (columnSetAllocationCount==0) {
+      // Estimate column count from maxblockcount, non-critical if ends up being too small.
+      columnAllocationCount = BLOCK_ALLOCATION_PADDING * std::pow(firstAllocationCount,0.666);
```
ursg, line 635: Another fun magical constant that might rather want to be a ```#define``` (or a ```const static double``` in global scope).

arch/gpu_base.hpp
```diff
@@ -50,11 +50,14 @@ static const uint VLASOV_BUFFER_MINCOLUMNS = 2000/WID;
 static const double BLOCK_ALLOCATION_PADDING = 1.2;
 static const double BLOCK_ALLOCATION_FACTOR = 1.1;
 
+// Used in acceleration column construction
+static const int GPU_PROBEFLAT_N = 5;
```
ursg, line 52: Is this dependent on reconstruction polynomial order?

vlasovsolver/gpu_acc_map.cpp
```diff
@@ -20,70 +20,510 @@
...
+   ) {
+   const uint parallelOffsetIndex = blockIdx.y;
+   const uint cellOffset = parallelOffsetIndex + cumulativeOffset;
+   vmesh::VelocityBlockContainer* blockContainer = blockContainers[cellOffset];
+   const size_t VBC_size = blockContainer->size() * WID3;
+   Realf *blockData = blockContainer->getData();
+
+   const size_t ind = blockIdx.x * blockDim.x + threadIdx.x;
+   for (size_t i = ind; i < VBC_size; i += gridDim.x * blockDim.x) {
+      blockData[i] = 0;
```
ursg, line 118: I'm surprised a simple cudaMemset isn't more efficient here. Because all this does is set blockdata to zero, in parallel, right?

vlasovsolver/gpu_acc_map.cpp
```diff
@@ -554,339 +919,479 @@ __global__ void __launch_bounds__(VECL,4) acceleration_kernel(
...
+
+  Then, utilizing pre-calculated SLICE-3D intersections, a kernel is launched to evaluate
+  the extents of velocity space after each column(set) has been accelerated as requested.
+  Block adjustment functions are called to batch-update the velocity space of the cell
+  in question to match the new requirements (adding and removing cells as needed). Then,
+  the velocity block container is cleared.
+
+  Finally the acceleration kernel itself is called. It reads the aligned velocity space data
+  from the intermediate buffer, advects the columns according to the SLICE-3D intersections,
+  and stores the resultant phase space density back into the velocity block container.
```
ursg: Big thumbs up for this high-level overview comment!

If anybody actually ever *read* our doxygen documentation, this would maybe be the best example in all of our codebase. :)

vlasovsolver/gpu_acc_map.cpp
```diff
@@ -554,339 +919,479 @@ __global__ void __launch_bounds__(VECL,4) acceleration_kernel(
...
+   flattenTimer.stop();
+
+   // This kernel performs an exclusive prefix scan to get offsets for storing
+   // data from potential columns into the columnData container. Also gives us the total
+   // counts of columns, columnsets, and blocks, and uses the first two to resize
+   // our splitvector containers inside columnData.
+
+   // A proper prefix scan needs to be a two-phase process, thus two kernels,
+   // but here we do an iterative loop processing MAX_BLOCKSIZE elements at once.
+   // Not as efficient but simpler, and will be parallelized over spatial cells.
```
ursg: The literal code style of this function, with its generous interspersed comments, is a joy to read.

It *does* have some redundancy with the functions' and kernels' individual comment blocks though, and I'm worried they might drift apart (and away from the actual reality) if the code develops further.

But I like it a lot. Let's keep it like this.

ursg (approved):
I like this a lot, and it is clearly visible that a proper work-through of the entire solver made it much more readable. Awesome!

My only comments are on code structure, and specifically the places where datastructures are declared and kept. This should be made more compact, as not to completely lose track of things in the near future. But that logically will be a separate PR.

arch/gpu_base.cpp
markusbattarbee: it's pow two thirds :) Take a v-space, assume it's roughly cubic, and find the cross-section. But yes, the description should be more verbose.

arch/gpu_base.hpp
markusbattarbee: No, the count of those is described in gpu_acc_map.cpp (but yes, should be referenced or explained here)

vlasovsolver/gpu_acc_map.cpp
markusbattarbee: It's a matter of launching one kernel to do all memsets vs calling one memset per spatial cell (64 launches). Probably not a huge difference in performance, but less looping in the host-side function, and the addressing is already available on the GPU.

gpuMemSet calls actually internally launch a kernel to do their work, so should be the same on-device performance.

Can update the documentation to reflect this choice.

vlasovsolver/gpu_acc_map.cpp
ykempf: I've been advocating for using doxygen style as if you ever want to go forward with some sphinx + doxygen automagication of online documentation, that'll be already there to some extent.

vlasovsolver/gpu_acc_map.cpp
ursg: Ah, then it makes sense.
And I could imagine that gpuMemset also somehow links into the driver's memory allocation locking chain, or otherwise serializes things in a way that we wouldn't care about.

spatial_cells/block_adjust_gpu.cpp
markusbattarbee: Launch parameters indeed depend on what each kernel does. And this is something which may change over different hardware iterations. However, too many knobs to dial are also not useful, and then it's a question of "well if that can be fine-tuned then why not also this and this and this?" 
I'd say that when we have code which is ready for a release, then we can evaluate (a bit more at scale) which kernels take significant amounts of time and then consider adding tuning knobs to those specifically, not every one.

For this kernel in particular, I wanted to write down that I *did* test it, and found that one GPU block per spatial cell was faster.

vlasovsolver/gpu_acc_map.cpp
markusbattarbee: "Don't change launch parameters unless you know what you're doing."
There, I wrote it :)
No but really, the kernels themselves utilize the launch parameters quite heavily in indexing into input parameter and pointer buffers. I can try to improve the documentation to that effect, but I don't think it's worth filling the source code with warnings about "don't change things unless you know what you're doing" everywhere - that has applied to our solvers always! :)

arch/gpu_base.cpp
markusbattarbee: That is a fair point, but I wanted to keep the bulk of GPU-based memory management (allocations etc) in one place. And instead of making a separate GPUmem sub-directory, I used this one. Can be changed, but that's semantics, and I'd rather leave it to the future "clean up GPU buffers into some actually neat and managed container objects" PR.

arch/gpu_base.cpp
markusbattarbee: For now. I agree that we want to clean things up, but one thing at a time.

arch/gpu_base.cpp
markusbattarbee: yep, deffo on the table for a future PR. Luckily this is now just for reporting.

arch/gpu_base.cpp
markusbattarbee: Addressed in updated comments.

vlasovsolver/gpu_acc_map.cpp
markusbattarbee: OK, will make function & kernel descriptions doxygenned tomorrow.

vlasovsolver/gpu_acc_map.cpp
ykempf: Sure. But if you explain a little bit the choices, the next time some externally auditing GPU guru looks at your code they'll have an easier time figuring whether there's potential in changing things or whether it's "locked" by the rest of the GP-fu you deployed.

PR #1127, opened 2025-06-10
 Implement Vectorclass Supporting SVE Instructions 

lnghrdntcr:
This PR adds support for a new type of Vectorclass, named VectorclassSVE. It serves as a drop-in replacement for both the default VectorclassFallback (used on non-x86 architectures) and Agner’s Vectorclass (used on x86).
VectorclassSVE provides the same feature set as both VectorclassFallback and Agner’s Vectorclass, as used within Vlasiator. However, unlike the other two, VectorclassSVE supports arbitrary power-of-two vector lengths, eliminating the need to choose vector length based on the data type.
The implementation of VectorclassSVE is primarily composed of three classes:
* VecSVEBase: The base class, which provides common operations and storage.
* VecSVE: The main class for performing arithmetic operations.
* VecSVEMask: The class used for and returned from logical operations.

At the core of this implementation are the `sve::functions::transform` higher-order-functions that apply SVE intrinsics to one or more arguments, returning either a VecSVE or VecSVEMask, depending on the operation. Such transform functions use both inlining and RVO to allow efficient chaining of multiple operations without creating intermediate vectors.

Benchmark results from the provided test cases show a median speedup of 13%. This improvement is however negatively affected by shorter test cases, which experience an average slowdown of 20%, and also higher performance variablity. In contrast, longer-running test cases (>10s) show an average speedup of 41%.


vlasovsolver/cpu_acc_load_blocks.cpp
```diff
@@ -165,7 +165,7 @@ cog.outl("#endif")
    if(dimension == 0) {
       for (vmesh::LocalID block_k=0; block_k<n_blocks; ++block_k) {
          Realf* __restrict__ data = blockContainer->getData(vmesh->getLocalID(blocks[block_k]));
-   #if (defined(VEC4D_AGNER) || defined(VEC4F_AGNER) || defined(VEC_FALLBACK_GENERIC)) && VECL == 4
+   #if (defined(VEC4D_AGNER) || defined(VEC4F_AGNER) || defined(VEC_FALLBACK_GENERIC) || defined(VEC4F_SVE) || defined(VEC4D_SVE)) && VECL == 4
```
ursg, line 168: Note: This file is autogenerated by the cog code at the top. You now modified the generated code, but did *not* update the cog generator above!

ursg (approved):
This looks good, and I can merge this as-is, but consider sending another PR later to also correspondingly update the cog generator (see detailed comment below).

Realistically though, we haven't regenerated that part of the code in maybe a decade, so we might never touch it again. Still, better to be safe than sorry.

PR #1128, opened 2025-06-10
GPU pitch angle diffusion

MikaelHuppunen:
Made a GPU version of the pitch angle diffusion calculations.


.gitignore
```diff
@@ -22,3 +22,7 @@ vlasiator+pat
 nbproject
 vlsvdiff_DP
 vlsvextract_DP
+libraries*/
+library-build/
+run_*/
+!run_tests.sh
```
markusbattarbee: Please add a newline at the end of the file here

arch/arch_device_cuda.h
```diff
@@ -88,6 +88,7 @@
 #define gpuKernelBallot(mask, input)     __ballot_sync(mask, input)
 #define gpuKernelAny(mask, input)        __any_sync(mask, input)
 #define gpuKernelShfl(input, source, mask)  __shfl_sync(mask, input, source)
+#define __shfl_down_gpu(val, offset) __shfl_down_sync(0xffffffff, val, offset) //0xffffffff is a mask that tells cuda to include all threads in the warp
```
markusbattarbee: Let's please keep a consistent naming scheme here - so `gpuKernelShflDown` (and same for _hip.h)

vlasiator.cpp
```diff
@@ -47,8 +47,15 @@
 #include "readparameters.h"
 #include "spatial_cells/spatial_cell_wrapper.hpp"
 #include "datareduction/datareducer.h"
+
 #include "sysboundary/sysboundary.h"
+#include "vlasovsolver/common_pitch_angle_diffusion.hpp"
+#ifdef USE_GPU
```
markusbattarbee: let's have these ifdefs (and inclusion of the correct `???_pitch_angle_diffusion.h??`) in `common_pitch_angle_diffusion.hpp` to keep the main cpp file a bit cleaner

vlasovsolver/common_pitch_angle_diffusion.hpp
```diff
@@ -0,0 +1,41 @@
...
+extern std::vector<Real> TanisoArray;
+extern std::vector<Real> nu0Array;
+extern size_t n_betaPara;
+extern size_t n_Taniso;
+extern bool nuArrayRead;
+
+void readNuArrayFromFile();
+
+Realf interpolateNuFromArray(
+   const Real Taniso, const Real betaParallel);
```
markusbattarbee: newline at end of file

vlasovsolver/common_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,180 @@
+/*
+ * This file is part of Vlasiator.
+ * Copyright 2010-2020 University of Helsinki
```
markusbattarbee: Generic comment about the copyright headers: Many of them are out of date, but let's take this opportunity to change the edited files to read:
`Copyright 2010-2025 Finnish Meteorological Institute and University of Helsinki`

vlasovsolver/common_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,180 @@
...
+      const Real nu022   = nu0Array[(betaIndx+1)*n_Taniso+TanisoIndx+1];
+      // Weights
+      const Real w11 = (beta2 - betaParallel)*(Taniso2 - Taniso)  / ( (beta2 - beta1)*(Taniso2-Taniso1) );
+      const Real w12 = (beta2 - betaParallel)*(Taniso  - Taniso1) / ( (beta2 - beta1)*(Taniso2-Taniso1) );
+      const Real w21 = (betaParallel - beta1)*(Taniso2 - Taniso)  / ( (beta2 - beta1)*(Taniso2-Taniso1) );
+      const Real w22 = (betaParallel - beta1)*(Taniso  - Taniso1) / ( (beta2 - beta1)*(Taniso2-Taniso1) );
+      // Linear interpolation (with fudge factor divisor)
+      return (w11*nu011 + w12*nu012 + w21*nu021 + w22*nu022)/Parameters::PADfudge;
+   }
+}
```
markusbattarbee: newline

vlasovsolver/common_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,180 @@
...
+   https://doi.org/10.1063/5.0176376
+ */
+std::vector<Real> betaParaArray;
+std::vector<Real> TanisoArray;
+std::vector<Real> nu0Array;
+size_t n_betaPara = 0;
+size_t n_Taniso = 0;
+bool nuArrayRead = false;
+
+void readNuArrayFromFile() {
```
markusbattarbee, line 53: I'm guessing these two functions are just directly moved over from cpu_pitch_angle_diffusion.cpp with no modifications?

vlasovsolver/cpu_pitch_angle_diffusion.cpp
```diff
@@ -31,6 +31,7 @@
 #include <Eigen/Geometry>
 #include "vec.h"
 #include "cpu_pitch_angle_diffusion.h"
```
markusbattarbee: This line 33 won't then be needed if the common header selectively includes also the architecture-specific one.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,743 @@
...
+#include <iostream>
+#include <fstream>
+#include <iomanip>
+#include <iterator>
+#include <Eigen/Geometry>
+#include <thrust/device_vector.h>
+#include <thrust/reduce.h>
+#include <thrust/functional.h>
+#include "vec.h"
+#include "gpu_pitch_angle_diffusion.hpp"
```
markusbattarbee: This line 39 won't then be needed if the common header selectively includes also the architecture-specific one.

vlasovsolver/gpu_pitch_angle_diffusion.hpp
```diff
@@ -0,0 +1,31 @@
...
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include <zoltan.h>
+#include <dccrg.hpp>
+#include "../common.h"
+#include "../spatial_cells/spatial_cell_wrapper.hpp"
+#include <dccrg_cartesian_geometry.hpp>
+
+void pitchAngleDiffusion(dccrg::Dccrg<spatial_cell::SpatialCell, dccrg::Cartesian_Geometry>& mpiGrid,const uint popID);
```
markusbattarbee: endline

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,743 @@
...
+            Vec vectorAdjust;
+            for (size_t i=0; i<cell.get_number_of_velocity_blocks(popID)*WID3/VECL; ++i) {
+               vectorAdjust.load(&cell.get_data(popID)[i*VECL]);
+               vectorAdjust *= adjustRatio;
+               vectorAdjust.store(&cell.get_data(popID)[i*VECL]);
+            }
+         }
+      }
+   } // End spatial cell loop
+} // End function
```
markusbattarbee: endline

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,743 @@
...
+//#include <cmath> // NaN Inf checks
+#include <iostream>
+#include <fstream>
+#include <iomanip>
+#include <iterator>
+#include <Eigen/Geometry>
+#include <thrust/device_vector.h>
+#include <thrust/reduce.h>
+#include <thrust/functional.h>
+#include "vec.h"
```
markusbattarbee: vec.h not needed once the mass conservation calculations are GPUified.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,743 @@
...
+   const Real Vmu = dev_dVbins[cellIdx] * (Vindex+0.5); // Take value at the center of the mu cell
+   int muindex = static_cast<int>(std::nearbyint(floor((mu+1.0) / dmubins)));
+
+   const Realf cellValue = dev_velocityBlockContainer[cellIdx]->getData()[velocityIdx*WID3+k*WID2+j*WID+i];
+   const Real increment = 2.0 * M_PI * Vmu*Vmu * cellValue;
+   // Safety check to handle edge case where mu = exactly 1.0
+   const int mui = std::max(0,std::min(muindex,nbins_mu-1));
+   const int vi = std::max(0,std::min(Vindex,nbins_v-1));
+
+   // TODO: can this be done without atomicAdd while avoiding race conditions?
```
markusbattarbee, line 104: Indeed.

When a large amount of gathering like this is required, it may be preferential for the block to first loop over a lot of input data, gathering into a local buffer for each thread, then gather that into shared memory (using staggered writes and `__syncthreads()` into an array of size nbins_mu * nbins_v). OR gather directly into shared memory, but there needs indeed to be staggering or you risk race conditions and data corruption.

After that, the shared memory buffer can be used to output into the actual output data. If there's only one block writing, you don't even need atomics there, but if there are several, then you can have threads index into the shared array and only `atomicadd` if the value is nonzero.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,743 @@
...
+   // resonance gap filling coefficient, not needed assuming even number of bins in mu-space
+   const Real epsilon = 0.0;
+
+   phiprof::Timer diffusionTimer {"pitch-angle-diffusion"};
+
+   const auto LocalCells=getLocalCells();
+
+   size_t numberOfLocalCells = LocalCells.size();
+
+   std::vector<Real> host_bValues (3*numberOfLocalCells, 0.0);
```
markusbattarbee: I'd suggest making these gpuHostMallocs (pinned memory, faster transfers)

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,743 @@
...
+         }
+
+         // Perform Eigen rotation to find parallel and perpendicular pressure
+         Eigen::Matrix3d rot = Eigen::Quaterniond::FromTwoVectors(Eigen::Vector3d{b[0], b[1], b[2]}, Eigen::Vector3d{0, 0, 1}).normalized().toRotationMatrix();
+         Eigen::Matrix3d Ptensor {
+            {cell.parameters[CellParams::P_11], cell.parameters[CellParams::P_12], cell.parameters[CellParams::P_13]},
+            {cell.parameters[CellParams::P_12], cell.parameters[CellParams::P_22], cell.parameters[CellParams::P_23]},
+            {cell.parameters[CellParams::P_13], cell.parameters[CellParams::P_23], cell.parameters[CellParams::P_33]},
+         };
+         Eigen::Matrix3d transposerot = rot.transpose();
```
markusbattarbee: These are the Eigen commands, so you should be safe to remove the `namespace Eigen` at the top.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,743 @@
...
+   } // End spatial cell loop
+   
+   // Create device arrays
+   Real *dev_bValues, *dev_nu0Values, *dev_sparsity, *dev_dVbins, *dev_bulkVX, *dev_bulkVY, *dev_bulkVZ;
+   Realf *dev_fmu, *dev_dfdt_mu;
+   int *dev_fcount;
+   vmesh::VelocityBlockContainer **dev_velocityBlockContainer;
+
+   // Allocate memory
+   CHK_ERR( gpuMalloc((void**)&dev_bValues, 3*numberOfLocalCells*sizeof(Real)) );
```
markusbattarbee: These allocations could also be made persistent - the number of local cells does not change very often, so no need to constantly allocate/deallocate. See the allocation commands in `arch/gpu_base.cpp`.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,743 @@
...
+         }
+
+         const auto CellID                  = LocalCells[CellIdx];
+         SpatialCell& cell                  = *mpiGrid[CellID];
+
+         maxBlockIndex += cell.get_number_of_velocity_blocks(popID);
+         maxCellIndex++;
+      } // End spatial cell loop
+
+      int maxThreadsPerBlock = 1024;
```
markusbattarbee: `Hashinator::defaults::MAX_BLOCKSIZE`

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,743 @@
...
+      // Copy data to device
+      CHK_ERR( gpuMemcpy(dev_cellIdxStartCutoff, host_cellIdxStartCutoff.data(), maxCellIndex*sizeof(size_t), gpuMemcpyHostToDevice) );
+      CHK_ERR( gpuMemcpy(dev_smallCellIdxArray, host_smallCellIdxArray.data(), maxCellIndex*sizeof(size_t), gpuMemcpyHostToDevice) );
+      CHK_ERR( gpuMemcpy(dev_remappedCellIdxArray, host_remappedCellIdxArray.data(), numberOfLocalCells*sizeof(size_t), gpuMemcpyHostToDevice) );
+
+      // Initialize with zero values
+      CHK_ERR( gpuMemset(dev_fmu, 0.0, numberOfLocalCells*nbins_v*nbins_mu*sizeof(Realf)) );
+      CHK_ERR( gpuMemset(dev_fcount, 0, numberOfLocalCells*nbins_v*nbins_mu*sizeof(int)) );
+
+      int totalThreadsPerBlock = 512;
```
markusbattarbee: What's this magic number? :)

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,743 @@
...
+      int maxThreadIndex = totalNumberOfVelocityBlocks;
+      int blocksPerGrid = (maxThreadIndex+totalThreadsPerBlock-1)/totalThreadsPerBlock;
+
+      // Find cell indices
+      getCellIndexArray_kernel<<<blocksPerGrid, totalThreadsPerBlock>>>(
+         dev_cellIdxArray, dev_velocityIdxArray, dev_cellIdxStartCutoff, totalNumberOfVelocityBlocks, maxCellIndex
+      );
+
+      dim3 threadsPerBlock(WID, WID, WID);
```
markusbattarbee: Can you please verify it still runs (and looks sane) when running with WID=8?

markusbattarbee (changes requested):
First cycle of comments, will delve deeper into the actual GPU implementation soon. :)

MikaelHuppunen:
Noticed a few places where 'cuda...' commands were used instead of 'gpu...'. Will fix.

markusbattarbee:
> Noticed a few places where 'cuda...' commands were used instead of 'gpu...'. Will fix.

Also the Thrust calls needs to be made agnostic for https://github.com/ROCm/rocThrust

vlasovsolver/common_pitch_angle_diffusion.cpp
MikaelHuppunen: That's correct

MikaelHuppunen:
> Noticed a few places where 'cuda...' commands were used instead of 'gpu...'. Will fix.

This is fixed.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,725 @@
...
+   int nextIndexInsideBlock = (blockIdx.x+1)%blocksPerVelocityCell;
+   int idx = indexInsideBlock*blockDim.x+threadIdx.x;
+   int threadIndex = threadIdx.x;
+   int indv = idx%nbins_v;
+   int indmu = (idx/nbins_v)%nbins_mu;
+   size_t cellIdx = dev_smallCellIdxArray[spatialBlockIndex];
+   
+   // Initiate shared memory values
+   extern __shared__ Real localDdtValues[];
+   localDdtValues[threadIndex] = std::numeric_limits<Real>::max();
```
kstppd, line 196: probably a syncthread missing here?

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,727 @@
...
+   
+   int totalBlockIndex = blockIdx.x*blockDim.x + threadIdx.x;
+   
+   if(totalBlockIndex >= totalNumberOfVelocityBlocks){return;}
+   
+   // Binary search
+   int left = 0;
+   int right = maxCellIndex - 1;
+   int cellIndex = 0;
+
```
kstppd, line 335: assert(right>left) .  Just for debug builds?

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,727 @@
...
+         out_keys, out_values,     // output
+         thrust::equal_to<int>(),              // binary predicate for keys
+         thrust::minimum<Real>()              // reduction operator
+      );
+      
+      CHK_ERR( gpuPeekAtLastError() );
+      CHK_ERR( gpuDeviceSynchronize() );
+
+      Real *dev_Ddt_values = thrust::raw_pointer_cast(out_values);
+      gpuMemcpy(host_Ddt.data(), dev_Ddt_values, maxCellIndex * sizeof(Real), gpuMemcpyDeviceToHost);
```
kstppd: check return value

vlasovsolver/gpu_pitch_angle_diffusion.cpp
MikaelHuppunen: Which debug flag should it be tied to?

vlasovsolver/gpu_pitch_angle_diffusion.cpp
MikaelHuppunen: I suppose something like that could be done. Do you think such a system should be implemented as this is already quite fast? I think it's quite nontrivial to implemented it in an efficient manner since nbins_mu * nbins_v tends to be much larger than the maximum number of threads per block and might not be much smaller than the number of velocity cells per spatial cell, which makes efficiently reducing the data harder but allows atomicAdd to operate quite smoothly.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
kstppd: We set -NDEBUG in production builds so that assert would be noop.

markusbattarbee: DEBUG_SOLVERS

vlasovsolver/gpu_pitch_angle_diffusion.cpp
MikaelHuppunen: Works fine

arch/gpu_base.cpp
```diff
@@ -125,6 +125,15 @@ uint gpu_acc_foundColumnsCount = 0;
 // split::SplitInfo *info_4[MAXCPUTHREADS];
 // Hashinator::MapInfo *info_m[MAXCPUTHREADS];
 
+// Pointers used in pitch angle diffusion
```
markusbattarbee, line 117: I recommend initializing these pointers as `NULL`, so on deallocation one can check and only deallocate if the pointer at that time is non-null.

arch/gpu_base.cpp
```diff
@@ -813,3 +823,49 @@ __host__ void gpu_trans_deallocate() {
...
+   CHK_ERR( gpuMalloc((void**)&dev_fmu, numberOfLocalCells*nbins_v*nbins_mu*sizeof(Realf)) );
+   CHK_ERR( gpuMalloc((void**)&dev_dVbins, numberOfLocalCells*sizeof(Real)) );
+   CHK_ERR( gpuMalloc((void**)&dev_bulkVX, numberOfLocalCells*sizeof(Real)) );
+   CHK_ERR( gpuMalloc((void**)&dev_bulkVY, numberOfLocalCells*sizeof(Real)) );
+   CHK_ERR( gpuMalloc((void**)&dev_bulkVZ, numberOfLocalCells*sizeof(Real)) );
+}
+
+void gpu_pitch_angle_diffusion_deallocate() {
+   // Free memory
+   CHK_ERR( gpuFree(dev_bValues) );
```
markusbattarbee: so checks here:
```
if (dev_bValues) {
   CHK_ERR( gpuFree(dev_bValues) );
}
```
etc

arch/gpu_base.cpp
```diff
@@ -813,3 +823,49 @@ __host__ void gpu_trans_deallocate() {
       }
    }
 }
+
+void gpu_pitch_angle_diffusion_allocate(size_t numberOfLocalCells, int nbins_v, int nbins_mu) {
+   // Allocate host memory
```
markusbattarbee, line 850: we know that `nbins_v` and `nbins_mu` won't change during runtime, but `numberOfLocalCells` may. Thus, you should have a counter to see for how many cells these buffers exist. When called, if the argument ncells is larger than the existing allocation, first de-allocate, then make a new allocation. If it's the same or smaller, then you can return without action.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
markusbattarbee: So far we've used `vbins=200` and `mubins=30`, which gives 6k elements. For the subgrid testpackage tests, we've had about 6700 blocks per spatial cell, so that's a 70-fold difference. 6k elements is still small enough to fit as as a shared memory buffer, as we can trust there's always at least 64k of shared memory available.

I did a test, running subgrid_2 with 8x8x8 spatial cells and turning on the other solvers as well. It's true that PAD does not take a huge amount of time there, but it's also non-negligible at 4% of propagation:
```
 | 255 | 3   |     |     Propagate                         | 30    | 424     | 99.1      | 0     | 1  |        |        | 1.562e+07 Cells/s |
 | 257 | 4   |     |       Spatial-space                   | 30    | 105.2   | 24.81     | 0     | 1  |        |        | 6.296e+07 Cells/s |
 | 258 | 5   |     |         semilag-trans                 | 30    | 105.2   | 100       | 0     | 1  |        |        |                   |
 | 260 | 6   |     |           translate proton            | 30    | 99.33   | 94.43     | 0     | 1  |        |        |                   |
 | 263 | 7   |     |             compute-mapping-z         | 30    | 33.59   | 33.81     | 0     | 1  |        |        |                   |
 | 273 | 8   |     |               trans-amr-mapping       | 30    | 33.27   | 99.06     | 0     | 1  |        |        |                   |
 | 278 | 7   |     |             compute-mapping-x         | 30    | 32.83   | 33.05     | 0     | 1  |        |        |                   |
 | 288 | 8   |     |               trans-amr-mapping       | 30    | 32.65   | 99.43     | 0     | 1  |        |        |                   |
 | 293 | 7   |     |             compute-mapping-y         | 30    | 32.91   | 33.13     | 0     | 1  |        |        |                   |
 | 303 | 8   |     |               trans-amr-mapping       | 30    | 32.75   | 99.5      | 0     | 1  |        |        |                   |
 | 307 | 6   |     |           Compute _R moments          | 30    | 5.855   | 5.566     | 0     | 1  |        |        |                   |
 |     | 7   |     |             Other                     | 30    | 5.853   | 99.97     | 0     | 1  |        |        |                   |
 | 339 | 4   |     |       Velocity-space                  | 30    | 301.6   | 71.14     | 0     | 1  |        |        | 2.196e+07 Cells/s |
 | 340 | 5   |     |         semilag-acc                   | 30    | 290.8   | 96.42     | 0     | 1  |        |        |                   |
 | 346 | 6   |     |           cell-semilag-acc            | 1920  | 272.5   | 93.69     | 0     | 8  | 0.1233 | 0.1233 |                   |
 | 347 | 6   | C   |           re-adjust blocks            | 30    | 15.21   | 5.231     | 0     | 1  |        |        |                   |
 | 360 | 7   |     |             Block halo batch kernels  | 30    | 6.066   | 39.87     | 0     | 1  |        |        |                   |
 | 368 | 5   |     |         Compute _V moments            | 30    | 10.79   | 3.577     | 0     | 1  |        |        |                   |
 |     | 6   |     |           Other                       | 30    | 10.79   | 100       | 0     | 1  |        |        |                   |
 | 370 | 4   |     |       Pitch-angle diffusion           | 30    | 17.06   | 4.025     | 0     | 1  |        |        | 3.881e+08 Cells/s |
 | 371 | 5   |     |         pitch-angle-diffusion         | 30    | 17.06   | 100       | 0     | 1  |        |        |                   |
```

arch/gpu_base.cpp
```diff
@@ -245,6 +254,7 @@ __host__ void gpu_clear_device() {
    gpu_trans_deallocate();
    gpu_moments_deallocate();
    gpu_batch_deallocate();
+   if (P::artificialPADiff){ gpu_pitch_angle_diffusion_deallocate(); }
```
markusbattarbee: and if you utilize null-initialized pointers, you don't need an if-clause here.

spatial_cells/block_adjust_gpu_kernels.hpp
```diff
@@ -103,23 +103,19 @@ __global__ void __launch_bounds__(WID3,WID3S_PER_MP) batch_update_velocity_block
...
                gathered_mass[ti] += gathered_mass[ti + s];
             }
             __syncthreads();
          }
-      } else {
-         // Perform loop only until first value fulfills condition
-         for (unsigned int s=WID3/2; s>0; s>>=1) {
-            if (has_content[0]) {
-               break;
-            }
```
markusbattarbee: please reinstate the early exit here - for certain configurations, it can have a significant impact.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,727 @@
...
+   const Real epsilon = 0.0;
+
+   phiprof::Timer diffusionTimer {"pitch-angle-diffusion"};
+
+   const auto LocalCells=getLocalCells();
+
+   size_t numberOfLocalCells = LocalCells.size();
+
+   // Allocate or reallocate memory if necessary
+   gpu_batch_allocate(numberOfLocalCells, 0);
```
markusbattarbee, line 450: oh you have the allocation count here?
I'd suggest indeed having it in the arch directory to have stuff in the same place, easier for future cleanup.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,727 @@
...
+#include <thrust/functional.h>
+#include "vec.h"
+#include "common_pitch_angle_diffusion.hpp"
+#include "../arch/gpu_base.hpp"
+#include "../spatial_cells/block_adjust_gpu.hpp"
+
+#define GPUCELLMUSPACE(var,cellIdx,v_ind,mu_ind) var[(cellIdx)*nbins_v*nbins_mu+(mu_ind)*nbins_v + (v_ind)]
+
+size_t latestNumberOfLocalCells = 0;
+int latestNbins_v = 0, latestNbins_mu = 0;
```
markusbattarbee: So we can work with the assumption that these  two do not change during runtime.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,727 @@
...
+   if (numberOfLocalCells != latestNumberOfLocalCells || nbins_v != latestNbins_v || nbins_mu != latestNbins_mu) {
+      latestNumberOfLocalCells = numberOfLocalCells;
+      latestNbins_v = nbins_v;
+      latestNbins_mu = nbins_mu;
+      return true;
+   }
+   return false;
+}
+
+__global__ void build2dArrayOfFvmu_kernel(
```
markusbattarbee: I tend to have kernel arguments one per line, would that be ok? (here and for the other kernels)

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,727 @@
...
+      SpatialCell& cell                  = *mpiGrid[CellID];
+      
+      host_sparsity[CellIdx]   = 0.01 * cell.getVelocityBlockMinValue(popID);
+      
+      //Initialize to zero
+      density_pre_adjust[CellIdx] = 0.0;
+
+      // Ensure mass conservation
+      if (getObjectWrapper().particleSpecies[popID].sparse_conserve_mass) {
+         //TODO: parallelize, tho this not usually used (I think) and probably not that expensive
```
markusbattarbee: Indeed usually not done, but if we do decide to use it, then this being on CPU will cause all data to be page faulted to device and back two times per cycle :)

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,727 @@
...
+         cell.parameters[CellParams::PERBZVOL] +  cell.parameters[CellParams::BGBZVOL]};
+      const Real Bnorm           = sqrt(B[0]*B[0] + B[1]*B[1] + B[2]*B[2]);
+      const std::array<Real,3> b = {B[0]/Bnorm, B[1]/Bnorm, B[2]/Bnorm};
+      
+      // Save values of b to vector 
+      host_bValues[3*CellIdx] = b[0];
+      host_bValues[3*CellIdx+1] = b[1];
+      host_bValues[3*CellIdx+2] = b[2];
+
+      if (P::PADcoefficient >= 0) {
```
markusbattarbee: This PAD coefficient finding is identical for GPU and CPU so could be split out into the commin_?.cpp file?

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,727 @@
...
+      SpatialCell *cell                  = mpiGrid[CellID];
+      const size_t meshID = getObjectWrapper().particleSpecies[popID].velocityMesh;
+      const vmesh::MeshParameters& vMesh = vmesh::getMeshWrapper()->velocityMeshes->at(meshID);
+
+      host_bulkVX[CellIdx] = (*cell).parameters[CellParams::VX];
+      host_bulkVY[CellIdx] = (*cell).parameters[CellParams::VY];
+      host_bulkVZ[CellIdx] = (*cell).parameters[CellParams::VZ];
+
+      const Real Vmax   = 2*sqrt(3)*vMesh.meshLimits[1];
+      host_dVbins[CellIdx] = Vmax/nbins_v;
```
markusbattarbee: these: `host_dVbins`, `Vmax` are constant for all cells.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
MikaelHuppunen: Actually assert(right>=left) since it will work as intended if right == left

markusbattarbee (commented):
commenting WIP

vlasovsolver/gpu_pitch_angle_diffusion.cpp
MikaelHuppunen: I suggest I will look into this more as a part of the overall optimization and not this pull request

vlasovsolver/common_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,246 @@
...
+      const Real w21 = (betaParallel - beta1)*(Taniso2 - Taniso)  / ( (beta2 - beta1)*(Taniso2-Taniso1) );
+      const Real w22 = (betaParallel - beta1)*(Taniso  - Taniso1) / ( (beta2 - beta1)*(Taniso2-Taniso1) );
+      // Linear interpolation (with fudge factor divisor)
+      return (w11*nu011 + w12*nu012 + w21*nu021 + w22*nu022)/Parameters::PADfudge;
+   }
+}
+
+void computePitchAngleDiffusionParameters(
+   const std::vector<CellID>& LocalCells, dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpiGrid,
+   const uint popID, size_t CellIdx, bool& currentSpatialLoopComplete,
```
markusbattarbee, line 182: here it would in fact be enough to just pass the reference to the cell instead of a vector, the grid object, and the index.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+#include <thrust/functional.h>
+#include "common_pitch_angle_diffusion.hpp"
+#include "../arch/gpu_base.hpp"
+#include "../spatial_cells/block_adjust_gpu.hpp"
+
+#define GPUCELLMUSPACE(var,cellIdx,v_ind,mu_ind) var[(cellIdx)*nbins_v*nbins_mu+(mu_ind)*nbins_v + (v_ind)]
+
+// The original code used these namespaces and as I'm not sure where they are used (which is a problem with namespaces),
+// they are still here
+using namespace spatial_cell;
```
markusbattarbee: this namespace is defined in `spatial_cells/spatial_cell_???.?pp`. Mainly used so one does not need to type out `spatial_cell::SpatialCell` as it's a bit redundant. If you prefer not using that namespace implicitly, that's fine as well, many files have foregone it.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+   }
+
+   if(allSpatialCellTimeLoopsComplete){
+      // No need for any calculations
+      return;
+   }
+
+   // Load CPU data
+   const size_t meshID = getObjectWrapper().particleSpecies[popID].velocityMesh;
+   const vmesh::MeshParameters& vMesh = vmesh::getMeshWrapper()->velocityMeshes->at(meshID);
```
markusbattarbee: a warning - in pretty much every file where we use or adjust the velocity space in our cells, a variable called a "vmesh" points to an object of type `vmesh::VelocityMesh`, not a MeshParameters object. This could cause confusion - thus, please rename it to vMeshParams or something similar.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+
+   const Real Vmax   = 2*sqrt(3)*vMesh.meshLimits[1];
+   Real dVbins = Vmax/nbins_v;
+
+   #pragma omp parallel for
+   for (size_t CellIdx = 0; CellIdx < numberOfLocalCells; CellIdx++) { // Iterate over all spatial cells
+      const auto CellID                  = LocalCells[CellIdx];
+      SpatialCell *cell                  = mpiGrid[CellID];
+
+      host_bulkVX[CellIdx] = (*cell).parameters[CellParams::VX];
```
markusbattarbee: Any reason this isn't `cell->parameters[...]`?

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+
+         const auto CellID                  = LocalCells[CellIdx];
+         SpatialCell& cell                  = *mpiGrid[CellID];
+
+         maxBlockIndex += cell.get_number_of_velocity_blocks(popID);
+         maxCellIndex++;
+      } // End spatial cell loop
+
+      int maxThreadsPerBlock = Hashinator::defaults::MAX_BLOCKSIZE;
+      int blocksPerVelocityCell = (nbins_v*nbins_mu+maxThreadsPerBlock-1)/maxThreadsPerBlock;
```
markusbattarbee: I think this variable is not well named. (In our usual terminology, a velocity cell is one v-space value - i.e. 1/WID3 of a block. :) And I don't think that's the case. 
I see this is a fast ceil division calculation, so a better name might be `muspace_gridsize` I guess?

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+
+      // Construct cellIdx arrays
+      std::vector<size_t> host_cellIdxStartCutoff(maxCellIndex);
+      std::vector<size_t> host_smallCellIdxArray(maxCellIndex);
+      std::vector<size_t> host_remappedCellIdxArray(numberOfLocalCells); // The position of the cell index in the sequence instead of the actual index
+
+      // Calculate indices
+      int remappedCellIdx = 0;
+      vmesh::LocalID totalNumberOfVelocityBlocks = 0;
+      for (size_t CellIdx = 0; CellIdx < numberOfLocalCells; CellIdx++) { // Iterate over all spatial cells
```
markusbattarbee, line 541: couldn't you merge this and the previous loop and just do vector push_backs? If you like, you can reserve them to hold up to `numberOfLocalCells` before the loop to prevent re-allocations.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+   CHK_ERR( gpuMemcpy(dev_bValues, host_bValues, 3*numberOfLocalCells*sizeof(Real), gpuMemcpyHostToDevice) );
+   CHK_ERR( gpuMemcpy(dev_nu0Values, host_nu0Values, numberOfLocalCells*sizeof(Real), gpuMemcpyHostToDevice) );
+   CHK_ERR( gpuMemcpy(dev_sparsity, host_sparsity, numberOfLocalCells*sizeof(Realf), gpuMemcpyHostToDevice) );
+   CHK_ERR( gpuMemcpy(dev_bulkVX, host_bulkVX, numberOfLocalCells*sizeof(Real), gpuMemcpyHostToDevice) );
+   CHK_ERR( gpuMemcpy(dev_bulkVY, host_bulkVY, numberOfLocalCells*sizeof(Real), gpuMemcpyHostToDevice) );
+   CHK_ERR( gpuMemcpy(dev_bulkVZ, host_bulkVZ, numberOfLocalCells*sizeof(Real), gpuMemcpyHostToDevice) );
+   CHK_ERR( gpuMemcpy(dev_VBCs, host_VBCs, numberOfLocalCells*sizeof(vmesh::VelocityBlockContainer*), gpuMemcpyHostToDevice) );
+   
+   dim3 threadsPerBlock(WID, WID, WID);
+   int totalThreadsPerBlock = WID3;
```
markusbattarbee: no need to declare this as dynamic shared memory, WID3 is compile-time constant so you can just declare it at that size in the kernel.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+         }else{
+            spatialLoopComplete[CellIdx] = true;
+         }
+      }
+
+   } // End Time loop
+
+   if (getObjectWrapper().particleSpecies[popID].sparse_conserve_mass) {
+      threadsPerBlock = dim3(WID, WID, WID);
+      totalThreadsPerBlock = WID3;
```
markusbattarbee: no need to declare this as dynamic shared memory, WID3 is compile-time constant so you can just declare it at that size in the kernel.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+      // Copy data to device
+      CHK_ERR( gpuMemcpy(dev_cellIdxStartCutoff, host_cellIdxStartCutoff.data(), maxCellIndex*sizeof(size_t), gpuMemcpyHostToDevice) );
+      CHK_ERR( gpuMemcpy(dev_smallCellIdxArray, host_smallCellIdxArray.data(), maxCellIndex*sizeof(size_t), gpuMemcpyHostToDevice) );
+      CHK_ERR( gpuMemcpy(dev_remappedCellIdxArray, host_remappedCellIdxArray.data(), numberOfLocalCells*sizeof(size_t), gpuMemcpyHostToDevice) );
+
+      // Initialize with zero values
+      CHK_ERR( gpuMemset(dev_fmu, 0.0, numberOfLocalCells*nbins_v*nbins_mu*sizeof(Realf)) );
+      CHK_ERR( gpuMemset(dev_fcount, 0, numberOfLocalCells*nbins_v*nbins_mu*sizeof(int)) );
+
+      totalThreadsPerBlock = Hashinator::defaults::MAX_BLOCKSIZE/2; //Using Hashinator::defaults::MAX_BLOCKSIZE/2 = 512 blocks can lead to better streaming multiprocessor occupancy
```
markusbattarbee: please use different launch parameter names for each kernel, will make it more feasible to keep track of how each kernel is launched.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+      int maxThreadIndex = totalNumberOfVelocityBlocks;
+      blocksPerGrid = (maxThreadIndex+totalThreadsPerBlock-1)/totalThreadsPerBlock;
+
+      // Find cell indices
```
markusbattarbee: More verbose commenting please, explaining also *why* something is done. :) I think I was able to deduce what's going on here, but it'll be helpful in the future to have proper comments and descriptions (especially as this approach you have here is something which I haven't used anywhere else in the code)

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+      getCellIndexArray_kernel<<<blocksPerGrid, totalThreadsPerBlock>>>(
+         dev_cellIdxArray,
+         dev_velocityIdxArray,
+         dev_cellIdxStartCutoff,
+         totalNumberOfVelocityBlocks,
+         maxCellIndex
+      );
```
markusbattarbee, line 586: peek at last error

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+
+   if (getObjectWrapper().particleSpecies[popID].sparse_conserve_mass) {
+      // Ensure mass conservation
+      calculateDensity_kernel<<<blocksPerGrid, threadsPerBlock, sharedMemorySize>>>(
+         dev_densityPreAdjust,
+         dev_VBCs
+      );
+      
+      CHK_ERR( gpuPeekAtLastError() );
+      CHK_ERR( gpuDeviceSynchronize() );
```
markusbattarbee: this sync isn't needed, I think?

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+   if (n == 0) return 1;
+   n--; // Handle exact powers of two
+   n |= n >> 1;
+   n |= n >> 2;
+   n |= n >> 4;
+   n |= n >> 8;
+   n |= n >> 16;
+   return n + 1;
+}
+
```
markusbattarbee, line 54: might as well put launch bounds suggestions in?

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+         for(cLeft = 1; indmu - cLeft > llimit; cLeft++){
+            if(GPUCELLMUSPACE(dev_fcount,cellIdx,indv,indmu - cLeft) != 0){
+               break;
+            }
+         }
+         if( (GPUCELLMUSPACE(dev_fcount,cellIdx,indv,indmu - cLeft) == 0) && (indmu - cLeft == llimit) ) {
+            cLeft  = 0;
+         }
+      }
+
```
markusbattarbee, line 233: if you want, you can place a sync here so warp divergence is contained to as short a region as possible.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+      // Search limits for how many cells in mu-direction should be max evaluated when searching for a near neighbour?
+      // Assuming some oversampling; changing these values may result in method breaking at very small plasma frame velocities.
+      const int rlimit = nbins_mu-1;
+      const int llimit = 0;
+
+      int cLeft = 0;
+      int cRight = 0;
+
+      // !!! DANGER, WARP DIVERGENCE AHEAD !!!
+      // TODO: Can this be avoided?
```
markusbattarbee, line 211: Not in an easy manner at least, I think...

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+      const Realf CellValue = GPUCELLMUSPACE(dev_fmu,cellIdx,indv,indmu) / (2.0 * M_PI * Vmu*Vmu);
+      const Realf absdfdt = abs(dfdt_mu_val); // Already scaled
+      
+      // Save calculated Ddt value
+      if (absdfdt > 0.0 && CellValue > dev_sparsity[cellIdx]) {
+         localDdtValues[threadIndex] = CellValue * PADCFL * (1.0/absdfdt);
+      }
+
+      // Reduction in shared memory
+      __syncthreads();
```
markusbattarbee, line 268: Just commenting here: Nice that you've optimized the loop so we don't need to do all cycles in shared memory, instead using the warp shuffle.

I think, however, that it should be possible to do "both cycles" this way - first gathering the min/max/sum from each warp in the whole domain, then have the first warp read the results of each subsequent warp into its values, and finally reduce those again. Not for this PR, but something to keep in mind.

(I'd imagine these reduction loops aren't all that slow in the big scheme of things)

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+         dev_smallCellIdxArray,
+         dev_fmu,
+         dev_fcount,
+         nbins_v,
+         nbins_mu,
+         maxThreadIndex
+      );
+      
+      CHK_ERR( gpuPeekAtLastError() );
+      CHK_ERR( gpuDeviceSynchronize() );
```
markusbattarbee: these interim device syncs are also not needed unless you're pulling some values to host

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+         epsilon,
+         Parameters::PADCFL,
+         nbins_v,
+         nbins_mu,
+         blocksPerVelocityCell,
+         lastBlockSize
+      );
+      
+      CHK_ERR( gpuPeekAtLastError() );
+      CHK_ERR( gpuDeviceSynchronize() );
```
markusbattarbee: unnecessary

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+      } // End spatial cell loop
+
+      std::vector<Real> host_Ddt (maxCellIndex);
+
+      // Create device arrays
+      size_t *dev_cellIdxArray, *dev_velocityIdxArray, *dev_smallCellIdxArray, *dev_remappedCellIdxArray, *dev_cellIdxStartCutoff;
+      Real *dev_Ddt, *dev_potentialDdtValues;
+      int *dev_cellIdxKeys;
+
+      // Allocate memory
```
markusbattarbee: Ah, more transient cuda mallocs? These could also be persistent and included in overall GPU memory consolidation. In particular as this is inside a while loop!

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+         out_keys, out_values,     // output
+         thrust::equal_to<int>(),              // binary predicate for keys
+         thrust::minimum<Real>()              // reduction operator
+      );
+      
+      CHK_ERR( gpuPeekAtLastError() );
+      CHK_ERR( gpuDeviceSynchronize() );
+
+      Real *dev_Ddt_values = thrust::raw_pointer_cast(out_values);
+      CHK_ERR( gpuMemcpy(host_Ddt.data(), dev_Ddt_values, maxCellIndex * sizeof(Real), gpuMemcpyDeviceToHost) );
```
markusbattarbee: now this copies to a vector instead of pinned host memory, which is slower.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
```diff
@@ -0,0 +1,799 @@
...
+      int* dev_out_keys;
+      Real* dev_out_values;
+      CHK_ERR( gpuMalloc(&dev_out_keys, maxCellIndex * sizeof(int)) );
+      CHK_ERR( gpuMalloc(&dev_out_values, maxCellIndex * sizeof(Real)) );
+      thrust::device_ptr<int> out_keys(dev_out_keys);
+      thrust::device_ptr<Real> out_values(dev_out_values);
+      thrust::device_ptr<int> in_keys(dev_cellIdxKeys);
+      thrust::device_ptr<Real> in_values(dev_potentialDdtValues);
+
+      // Run reduce_by_key
```
markusbattarbee: A bit more documentation please

markusbattarbee (changes requested):
Can you also please perform the visual check of the results of diffusion for HILE (AMD/HIP)? (and post the testpackage output for both CUDA and HIP versions of all 4 versions of the subgrid diffusion testpackage tests)

vlasovsolver/gpu_pitch_angle_diffusion.cpp
MikaelHuppunen: Yes, this is a mistake as it is indeed not blocks per velocity cell but rather per spatial cell.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
MikaelHuppunen: You could. Push_backs are slower than reserving the correct amount of memory to begin with. I suppose I could reserve for ```numberOfLocalCells```. This would of course mean a little bit of extra allocation in some cases but shouldn't be too much and as a benefit the allocation could be done outside the while loop.

vlasovsolver/gpu_pitch_angle_diffusion.cpp
MikaelHuppunen: I agree

vlasovsolver/gpu_pitch_angle_diffusion.cpp
MikaelHuppunen: I'll keep that in mind

MikaelHuppunen:
[nvidia.txt](https://github.com/user-attachments/files/20798284/nvidia.txt)
[nvidia.txt](https://github.com/user-attachments/files/20798284/nvidia.txt)
Test output with nvidia

markusbattarbee:
The subgrid test 4 mass conservation results are not quite as nice as I'd like. Now the question is, is the problem in the CPU or in the GPU version? :)

You can run with the `vg_rhom` diagnostic output value on and with a high cadence of diagnostic output. With mass conservation, the number should remain the same. If it changes in one version or another, there's an issue there.

MikaelHuppunen:
> The subgrid test 4 mass conservation results are not quite as nice as I'd like. Now the question is, is the problem in the CPU or in the GPU version? :)
> 
> You can run with the `vg_rhom` diagnostic output value on and with a high cadence of diagnostic output. With mass conservation, the number should remain the same. If it changes in one version or another, there's an issue there.

Remains basically constant on CPU (3.78744012892e-21 to 3.78744012894e-21), but changes slightly on GPU (3.7848265371e-21 to 3.78449233025e-21). Same thing already happens on dev (so with CPU version of pitch angle diffusion and GPU everything else).

ursg (approved):
I mean, it shares the same structural ugliness as all the other GPU code. But apart from that, it looks good, seems to work and doesn't break anything. So... great!

PR #1129, opened 2025-06-10
Vlasiator GPU to propagate WID3 at once in acceleration as well

markusbattarbee:
To be merged after #1125

This PR changes both the acceleration and translation solvers for GPU mode to act on Realf* buffers, purging any references to vectorclasses (either Agner or other).

The acceleration solver acts on a whole block at once. Care was needed for indexing and thread synchronizations so that writes into blockData do not get race conditions and lead to data corruption. But now it's done!

This also puts all the reconstruction helpers into separate CPU and GPU versions for clarity. This way, if we want specific flexible-vectorclass-versions for the future, they can get their own ones.


markusbattarbee:
Output from Mahti TP run
[slurm-4643178.txt](https://github.com/user-attachments/files/20674132/slurm-4643178.txt)

markusbattarbee:
OK, this is now ready for proper review and merging. 

I experimented with gathering acceleration results into a large shared memory buffer and then doing one loop of writing that out into global memory (the VBC), but that a) wasn't any faster b) fails with WID=8 because the columndata becomes too large to fit in the available shared memory. So let's not do that.

markusbattarbee:
Updated:
[slurm-4676231.txt](https://github.com/user-attachments/files/20761282/slurm-4676231.txt)

markusbattarbee:
Up-to-date and in my view acceptable GPU testpackage output from Mahti:
[slurm-4689699.txt](https://github.com/user-attachments/files/20798999/slurm-4689699.txt)

PR #1131, opened 2025-06-17
Scaling results run on Mahti, with extracted data, plotting scripts and resulting plots.

ykempf:


ykempf:
Actually I'll change the target to master so that it is visible in the wiki right away and not only after the next merge of dev to master, right?

ykempf:
Duh no it's not in wiki it's in README.

ursg:
The CI is still red due to things that I fixed last week already, so let us not be disturbed by them.

PR #1132, opened 2025-06-18
Add Hile-G TP build CI job

markusbattarbee:
Even though HILE-g isn't yet fast enough for TP running, we can at least try building there. Could add the CPU side there as well in the future, this is just a first trial (needs people to set up HILE runners as well)


markusbattarbee:
A lot of the constructs in the ionospheric sysboundary (and a few other things too) seem to have a mismatch of `Real` and `double` resulting, for some reason, in the HILE compilers not compiling but rather complaining a lot. We have debug builds on some other architectures, so we know they aren't totally broken, but just broken on HPE. However, I won't have time to fix all that now - that's left for a further Ionosphere PR for @ursg.

.github/workflows/github-ci.yml
```diff
@@ -296,6 +296,166 @@ jobs:
...
+      - name: Checkout source
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Make clean
+        run: VLASIATOR_ARCH=hile_gpu make clean
+      - uses: ursg/gcc-problem-matcher@master
+      - name: Compile vlasiator (Release, debug build w/ HIP)
+        run: |
+          # srun -C g --job-name CI_debug_G_compile --interactive --nodes=1 -n 1 -c 16 --mem-per-cpu=2G -t 0:20:0 bash -c 'module load papi; module load rocm/6.2.0; module load cray-pmi; module load craype-accel-amd-gfx90a; module load libfabric/1.22.0; export MPICH_GPU_SUPPORT_ENABLED=1; export VLASIATOR_ARCH=hile_gpu; make COMPFLAGS="-DDEBUG_VLASIATOR -DDEBUG_SOLVERS -DDEBUG_IONOSPHERE -DDEBUG_SPATIAL_CELL -DDEBUG_VMESH -DDEBUG_VBC -DDEBUG_ACC " -j 16; sleep 10s'
```
ykempf, line 363: Why is this commented here?

.github/workflows/github-ci.yml
```diff
@@ -296,6 +296,166 @@ jobs:
...
+      - name: Checkout source
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Make clean
+        run: VLASIATOR_ARCH=hile_cpu make clean
+      - uses: ursg/gcc-problem-matcher@master
+      - name: Compile vlasiator (Release, debug build w/ HIP)
+        run: |
+          # srun -C c --job-name CI_debug_C_compile --interactive --nodes=1 -n 1 -c 16 --mem-per-cpu=2G -t 0:20:0 bash -c 'module load papi; module load cray-pmi; module load libfabric/1.22.0; export VLASIATOR_ARCH=hile_cpu; make COMPFLAGS="-DDEBUG_VLASIATOR -DDEBUG_SOLVERS -DDEBUG_IONOSPHERE -DDEBUG_SPATIAL_CELL -DDEBUG_VMESH -DDEBUG_VBC -DDEBUG_ACC " -j 16; sleep 10s'
```
ykempf, line 443: Same here, not needed?

ykempf: Or is the intent to expand and do debug builds here too?

ykempf (commented):
Unless the plan is also to include the debug builds, I guess it is good to go?

.github/workflows/github-ci.yml
markusbattarbee: Yes, I think it would be good to have prod+debug build tests on HILE in the future, but that requires plenty of cleaning of the ionosphere code. Better to have at least non-debug prod build active, IMO, to catch any SPF/DPF issues.

ykempf:
All right then if these debug builds don't get activated it's good to go I suppose.

ykempf:
Fascinating that Flowthrough_damr again has a grid mismatch in this PR.

ykempf (approved):

PR #1135, opened 2025-06-26
Asterix lossy VDF compression

kstppd:
So this is not a PR ready for review just leaving it here as a draft. There are dependencies from the ASTERIX repo so I bet this will also fail to build


datareduction/datareducer.cpp
```diff
@@ -22,15 +22,52 @@
...
 #include "datareducer.h"
 #include "../common.h"
+#include "datareductionoperator.h"
 #include "dro_populations.h"
 #include "../sysboundary/ionosphere.h"
 #include "../fieldtracing/fieldtracing.h"
 
 using namespace std;
 
+namespace {
```
alhom: The spell checker (while nice!) should be its own PR.

vlasiator.cpp
```diff
@@ -27,6 +27,7 @@
 #include <vector>
 #include <sstream>
 #include <ctime>
+#include <chrono>
```
alhom, line 30: Not used?

datareduction/datareducer.cpp
kstppd: Oh right! I forgot this had leaked in here

vlasiator.cpp
kstppd: Used to but not anymore

vlasovsolver/vec.h
```diff
@@ -170,7 +170,6 @@ typedef Vec16fb Vecb;
 #define VECL 16
 #define VPREC 4
 #ifndef VEC_PER_PLANE
-const int VEC_PER_PLANE = (WID*WID/VECL);
```
alhom, line 173: What's the rationale behind this change?

ursg:
Notes from us looking through the features of this PR this morning:
- This is still missing the fetch / build_libraries build process for automagically making MLP methods compile
- Currently, this compresses all recover and restart files. This should be configurable (realistically, we want the restarts to be lossless and the recovers to potentially be lossy)
- Compressed VDFs in bulk files also don't have a config option yet. It makes sense to add them as ```io.system_write_distribution_compression``` or something. The corresponding code is in iowrite.cpp
- The ```ASTERIX``` config settings needs a bit of cleanup and understandable help texts (also good default values)
- (We went on a tangent here and agreed that we'd really like a turing-complete config language, but maybe we shouldn't)
- We want CI steps testing compilation and running of all compression methods. They can run the simplest possible thing that produces a nonthermal distribution function (like the losscone setup), write output after a single timestep and check it against reference data.

kstppd:
To be reopened from a different branch.

PR #1141, opened 2025-07-17
Gpu optimization

MikaelHuppunen:
Optimized GPU kernels


spatial_cells/block_adjust_gpu_kernels.hpp
```diff
@@ -91,41 +96,47 @@ __global__ void __launch_bounds__(WID3,WID3S_PER_MP) batch_update_velocity_block
...
-            if (ti < s) {
-               has_content[ti] = has_content[ti] || has_content[ti + s];
-            }
-            __syncthreads();
+      }
+      
+      const int indexInsideWarp = ti % GPUTHREADS;
+      const int warpIndex = ti / GPUTHREADS;
+
+      // Perform loop only until first value fulfills condition
```
markusbattarbee: There is no longer a loop, so this comment should be updated that it's two consecutive warp votes.

spatial_cells/block_adjust_gpu.cpp
```diff
@@ -361,11 +361,14 @@ void adjust_velocity_blocks_in_cells(
       CHK_ERR( gpuPeekAtLastError() );
       #else
       dim3 grid_vel_halo(largestContentList,nCells,1);
```
markusbattarbee: Since you're now allowing for more v-blocks in parallel, this launch dim3 should have `largestContentList/warpsPerBlockBatchHalo`(ceil_int) instead :)

spatial_cells/block_adjust_gpu_kernels.hpp
```diff
@@ -747,28 +757,31 @@ __global__ void __launch_bounds__(GPUTHREADS*WARPSPERBLOCK, FULLBLOCKS_PER_MP) b
...
 
    const int blockWidth = blockDim.x; // how many GIDs each GPU block manages at once (in parallel)
    const int ti = threadIdx.x; // [0,blockSize)
    const int blockiStart = blockIdx.x * blockWidth;
 
-   const split::SplitVector<vmesh::GlobalID>* __restrict__ velocity_block_with_content_list = neigh_velocity_block_with_content_lists[neighIndex];
    const int nBlocks = velocity_block_with_content_list->size();
 
-   for (int blocki = blockiStart + ti; blocki < blockiStart+blockWidth; blocki += blockWidth) {
+   //for (int blocki = blockiStart + ti; blocki < blockiStart+blockWidth; blocki += blockWidth) { pointless loop, always exactly 1 iteration
```
markusbattarbee: yep, an old remnant, can be just deleted.

spatial_cells/block_adjust_gpu_kernels.hpp
```diff
@@ -747,28 +757,31 @@ __global__ void __launch_bounds__(GPUTHREADS*WARPSPERBLOCK, FULLBLOCKS_PER_MP) b
...
+   {
+      const int blocki = blockiStart + ti;
```
markusbattarbee: and this can be merged into line 774 (FMA!)

vlasovsolver/gpu_1d_ppm_nonuniform.hpp
```diff
@@ -43,21 +43,21 @@ ARCH_DEV inline void compute_ppm_coeff_nonuniform(const Realf* __restrict__ cons
    compute_filtered_face_values_nonuniform(dv, values, k, order, m_face, p_face, threshold, index, stride);
 
    //Coella et al, check for monotonicity
-   m_face = ((p_face - m_face) * (values[k*stride+index] - 0.5 * (m_face + p_face)) >
-             (p_face - m_face)*(p_face - m_face) * (1./6.)) ?
-      3 * values[k*stride+index] - 2 * p_face :
+   m_face = ((p_face - m_face) * (values[k*stride+index] - (Realf)(0.5) * (m_face + p_face)) >
```
markusbattarbee, line 46: Let's add these `(Realf)` casts to `gpu_1d_plm.hpp` and `gpu_1d_ppm.hpp` as well then, if they are necessary. BTW, what's the reasoning here? Doesn't the compiler know to use that precision anyway as as both the target and the other multiplicative terms are of type Realf?

vlasovsolver/gpu_acc_map.cpp
```diff
@@ -908,6 +908,12 @@ __global__ void __launch_bounds__(WID3) reorder_blocks_by_dimension_kernel(
    } // if valid setIndex
 }
```
markusbattarbee, line 910: I had a look at some performance analysis on Mahti (A100s), and it showed register usage at 57 per thread. Thus, I guess it's not really feasible to try and crunch it down to 40 or 32... 

Something to keep in mind though, especially looking at register usage on AMD as well.

vlasovsolver/gpu_acc_map.cpp
```diff
@@ -970,19 +975,37 @@ __global__ void __launch_bounds__(WID3,Hashinator::defaults::MAX_BLOCKSIZE/(WID3
...
+   __shared__ uint setColumnOffset;
+   __shared__ uint numColumns;
 
    // shared memory buffer for reducing looping count per block
-   __shared__ int loopN[WID3];
+   __shared__ int loopN[WID3/GPUTHREADS];
 
-   if (setIndex >= columnData->dev_sizeColSets()) {
-      return;
+   {
```
markusbattarbee, line 987: Oh, hm, would this be useful in other kernels too, reducing the scope of certain elements where possible, freeing up registers etc?

vlasovsolver/gpu_acc_map.cpp
```diff
@@ -1074,34 +1126,45 @@ __global__ void __launch_bounds__(WID3,Hashinator::defaults::MAX_BLOCKSIZE/(WID3
 
                // compute right integrand
```
markusbattarbee: perhaps add a comment that this is to trigger FMA operations

vlasovsolver/gpu_acc_map.cpp
```diff
@@ -1074,34 +1126,45 @@ __global__ void __launch_bounds__(WID3,Hashinator::defaults::MAX_BLOCKSIZE/(WID3
...
                // Write values into block data
-               if (isfinite(tval) && (tval>0) && (targetLID != invalidLID) ) {
-                  gpu_blockData[targetLID * WID3 + tcell] += tval;
-                  // atomicAdd(&gpu_blockData[targetLID*WID3+tcell],tval);
+               if (isfinite(tval) && (tval>(Realf)(0.0)) && (targetLID != invalidLID) ) {
+                  // gpu_blockData[targetLID * WID3 + tcell] += tval;
+
+                  // We use atomicAdd to avoid the need for sync
+                  // It shouldn't be any slower if there is no competition
+                  atomicAdd(&gpu_blockData[targetLID*WID3+tcell],tval);
```
markusbattarbee, line 1163: This should be also checked on AMD hardware, whether it's the case! Floating point atomic operations on AMD can behave *very* differently.... (insert can of worms)

vlasovsolver/gpu_acc_map.cpp
```diff
@@ -1008,26 +1034,52 @@ __global__ void __launch_bounds__(WID3,Hashinator::defaults::MAX_BLOCKSIZE/(WID3
 
          // Min/max Velocity coordinates in acceleration direction for this block
          const Realf min_lagrangian_v_l = v_r0 + blockOffset * dv;
```
markusbattarbee: Looks to me like `min_lagrangian_v_l`, `subcolumnMinGk`, and `minGk` could also go inside thta reduced scope below, retaining the previous readability and comments? (min and max extents on consecutive lines)

vlasovsolver/gpu_moments.cpp
```diff
@@ -70,46 +70,71 @@ __global__ void first_moments_kernel (
...
-       const Real DV3 = blockParamsZ[BlockParams::DVX]*blockParamsZ[BlockParams::DVY]*blockParamsZ[BlockParams::DVZ];
-       const Real VX = blockParamsZ[BlockParams::VXCRD] + (i+HALF)*blockParamsZ[BlockParams::DVX];
-       const Real VY = blockParamsZ[BlockParams::VYCRD] + (j+HALF)*blockParamsZ[BlockParams::DVY];
-       const Real VZ = blockParamsZ[BlockParams::VZCRD] + (k+HALF)*blockParamsZ[BlockParams::DVZ];
-       myMom[0] += f * DV3;
-       myMom[1] += f * VX * DV3;
-       myMom[2] += f * VY * DV3;
-       myMom[3] += f * VZ * DV3;
+      const Realf cellValue = data[blockIndex*WID3+indx];
+      const Real* blockParamsZ = &blockParameters[blockIndex*BlockParams::N_VELOCITY_BLOCK_PARAMS];
```
markusbattarbee: should this also be `__restrict__`? (and same for the second moments kernel)

vlasovsolver/gpu_moments.cpp
```diff
@@ -342,11 +401,14 @@ void gpu_calculateMoments_R(
          continue;
       }
       // Launch kernel calculating this species' contribution to second velocity moments
-      dim3 blockSize(WID,WID,WID);
-      vmesh::LocalID maxVmeshLaunch = sqrt(maxVmeshSizes.at(popID));
+
+      vmesh::LocalID maxVmeshLaunch = (gpuMultiProcessorCount*min(threadsPerMP/WID3, blocksPerMP))/nAllCells;//sqrt(maxVmeshSizes.at(popID));
```
markusbattarbee: that old sqrt-guess can be removed :)

markusbattarbee (changes requested):
Looks very promising!

vlasovsolver/gpu_1d_ppm_nonuniform.hpp
MikaelHuppunen: It seems to understand to use single precision on some machines, but on others it first converts to double precision, computes using double precision, and then converts back to single precision.

vlasovsolver/gpu_acc_map.cpp
MikaelHuppunen: I wasn't able to further reduce register pressure without the kernel being slowed down by the register usage optimizations.

vlasovsolver/gpu_acc_map.cpp
MikaelHuppunen: Yes, I think it's generally better to do this than not to, as long as you don't need to otherwise compromise performance to do so. In reality the register pressure isn't usually reduced that much but it might free up one or two registers, which might be impactful if register pressure is the bottle neck.

vlasovsolver/gpu_acc_map.cpp
MikaelHuppunen: Yes, I could look into it more carefully.

spatial_cells/block_adjust_gpu.cpp
MikaelHuppunen: Yes, seems like I missed that.

vlasovsolver/gpu_acc_map.cpp
MikaelHuppunen: minGk is used for const int gk = minGk + loopgk + k; so it would be out of scope for that instruction. Though I suppose I could define minGk outside the reduced scope and then compute them inside it.

vlasovsolver/gpu_acc_map.cpp
MikaelHuppunen: Seems like doing this is indeed slower on lumi

markusbattarbee: Which is fixed by using hashinator fmihpc/master and turning on unsafe atomics on amd. 🎉

PR #1142, opened 2025-07-24
Fix multi-gpu buffer allocations issue

markusbattarbee:
Change allocations to evaluate base grid size instead number of local cells in case task gets originally just 1 cell. 
Also squelch some compile warnings.


markusbattarbee:
So I've fixed and improved a number of buffer allocations, but currently Flowthrough_damr gets stuck hanging somewhere (deadlock?) when running with several tasks.

ursg:
Hmm, would be great to simply re-run the CI jobs here, now that I fixed a thing or two there, but apparently github has a time limit for rerunning them. I guess I'll just reverse-merge dev into this branch, and see if it performs well on the merge commit.

ursg:
Yep, this looks good now!

PR #1143, opened 2025-07-30
GPU bin pencils

MikaelHuppunen:
Implemented pencil bins on GPU

### DIFF
--- a/arch/gpu_base.hpp
+++ b/arch/gpu_base.hpp
@@ -268,42 +268,36 @@ struct GPUMemoryManager {
 
    // Allocate memory to a pointer by name
    bool allocate(const std::string& name, size_t bytes) {
+      //TODO: only allocate if needs to increase in size
       std::lock_guard<std::mutex> lock(memoryMutex);
       if (gpuMemoryPointers.count(name) == 0) {
          std::cerr << "Error: Pointer name '" << name << "' not found.\n";
          return false;
       }
+      
       if (gpuMemoryPointers[name] != nullptr) {
-         cudaFree(gpuMemoryPointers[name]);  // Re-allocate if needed
+         gpuFree(gpuMemoryPointers[name]);
       }
-      cudaError_t err = cudaMalloc(&gpuMemoryPointers[name], bytes);
-      if (err != cudaSuccess) {
-         std::cerr << "cudaMalloc failed for '" << name << "': " << cudaGetErrorString(err) << "\n";
-         gpuMemoryPointers[name] = nullptr;
-         allocationSizes[name] = 0;
-         return false;
-      }
+
+      CHK_ERR( gpuMalloc(&gpuMemoryPointers[name], bytes) );
       allocationSizes[name] = bytes;
       return true;
    }
 
    // Allocate pinned host memory to a pointer by name
    bool hostAllocate(const std::string& name, size_t bytes) {
+      //TODO: only allocate if needs to increase in size
       std::lock_guard<std::mutex> lock(memoryMutex);
       if (gpuMemoryPointers.count(name) == 0) {
          std::cerr << "Error: Pointer name '" << name << "' not found.\n";
          return false;
       }
+
       if (gpuMemoryPointers[name] != nullptr) {
-         cudaFreeHost(gpuMemoryPointers[name]);  // Free previously allocated host memory
+         gpuFreeHost(gpuMemoryPointers[name]);
       }
-      cudaError_t err = cudaMallocHost(&gpuMemoryPointers[name], bytes);
-      if (err != cudaSuccess) {
-         std::cerr << "cudaMallocHost failed for '" << name << "': " << cudaGetErrorString(err) << "\n";
-         gpuMemoryPointers[name] = nullptr;
-         allocationSizes[name] = 0;
-         return false;
-      }
+
+      CHK_ERR( gpuMallocHost(&gpuMemoryPointers[name], bytes) );
       allocationSizes[name] = bytes;
       return true;
    }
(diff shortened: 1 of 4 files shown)


vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -181,6 +188,58 @@ struct setOfPencils {
...
          activeBins.push_back(bin);
       }
+
+      #ifdef USE_GPU
+      gpuBins();
+      #endif
+   }
+
+   void gpuBins(){
+      #ifdef USE_GPU
```
markusbattarbee: You can even put the whole function inside the ifdef.

arch/gpu_base.hpp
```diff
@@ -242,6 +243,107 @@ struct ColumnOffsets {
...
+
+   // Create a new pointer with a base name, ensure unique name
+   std::string createPointer(const std::string& baseName) {
+      std::lock_guard<std::mutex> lock(memoryMutex);
+      std::string uniqueName = baseName;
+      if (gpuMemoryPointers.count(baseName)) {
+         int& counter = nameCounters[baseName];
+         uniqueName = baseName + "_" + std::to_string(++counter);
+      } else {
+         nameCounters[baseName] = 0;
```
markusbattarbee, line 266: shouldn't this be `=1;`?

vlasovsolver/cpu_trans_pencils.hpp
```diff
@@ -181,6 +188,58 @@ struct setOfPencils {
...
+      if(dev_pencilsInBin == "null"){
+         dev_pencilsInBin = gpuMemoryManager.createPointer("dev_pencilsInBin");
+      }
```
markusbattarbee: You could consider the requested name and the current name both as parameters by reference, and doing the `if null` check inside `createPointer` to neaten this section. An idea for future memory management features, but I like this approach!

vlasovsolver/gpu_trans_map_amr.cpp
```diff
@@ -130,168 +134,177 @@ __global__ void __launch_bounds__(WID3) translation_kernel(
...
-            #else
-            const vmesh::LocalID blockLID = vmesh->getLocalID(blockGID);
-            #endif
-            // Store block data pointer for both loading of data and writing back to the cell
-            if (blockLID != vmesh->invalidLocalID()) {
-               #ifdef DEBUG_VLASIATOR
-               const vmesh::LocalID meshSize = vmesh->size();
-               const vmesh::LocalID VBCSize = cellContainer->size();
-               if ((blockLID>=meshSize) || (blockLID>=VBCSize)) {
+      for (uint nBin = blockIdx.z; nBin < numberOfBins; nBin += gridDim.z) {
```
markusbattarbee: this is in fact a single-step loop by construction, right? :) 
So could be reverted to just using `blockIdx.z` wherever `nBin` is needed.

vlasovsolver/gpu_trans_map_amr.cpp
```diff
@@ -130,168 +134,177 @@ __global__ void __launch_bounds__(WID3) translation_kernel(
...
+               pencilBlocksCount[pencilBlocksCountOffset + pencili] = nonEmptyBlocks;
             }
             __syncthreads();
-         } // End loop over this pencil
-         if (ti==0) {
-            pencilBlocksCount[pencilBlocksCountOffset + pencili] = nonEmptyBlocks;
-         }
-         __syncthreads();
-      } // end loop over all pencils
+         } // end loop over pencils
```
markusbattarbee: `+ in this bin`

vlasovsolver/gpu_trans_map_amr.cpp
```diff
@@ -130,168 +134,177 @@ __global__ void __launch_bounds__(WID3) translation_kernel(
...
-               (pencilBlockData[pencilBlockDataOffset + celli])[ti] = (Realf)(0.0);
-            }
-         }
-      } // end loop over all cells
-
-      __syncthreads();
+         __syncthreads();
+         // Now we reset target blocks
+         for (uint pencilIndex = 0; pencilIndex < dev_binSize[nBin]; pencilIndex++) {
+            const uint pencili = dev_pencilsInBin[dev_binStart[nBin]+pencilIndex];
```
markusbattarbee: As this section now changed from just clearing all target cells once to looping over all pencils and going through their target cells, I want to double-check the logic and counter use here before merging.

markusbattarbee (commented):
Some questions and suggestions, and pending one check

arch/gpu_base.hpp
MikaelHuppunen: In the current implementation the nameCounters[baseName] basically just tells the index of the last pointer with that the given baseName. When it comes to functionality it doesn't really matter what the number is as long as the numbers for different pointers with the same baseName are unique. So it could be changed to nameCounters[baseName] = 1; and std::to_string(counter++); if we want the nameCounters to tell how many pointers there are with a given baseName?

vlasovsolver/cpu_trans_pencils.hpp
MikaelHuppunen: This createPointer method is able to create multiple pointers with the same baseName so we can't really do the null check inside the createPointer method. Here the null check is done to see whether dev_pencilsInBin pointer has been defined for this specific iteration of the setOfPencils struct.

vlasovsolver/gpu_trans_map_amr.cpp
MikaelHuppunen: Yes, with current launch settings it's always exactly one iteration.

arch/gpu_base.hpp
markusbattarbee: ah, my brain had skipped over the `count()` part of that line! All good then.

PR #1146, opened 2025-07-31
CI: WID=8 support, Ukko-DGX GPU TP activation

markusbattarbee:
Activates Ukko GPU TP

Adds WID=8 block size TP on Ukko GPU, but does not yet activate it as it doesn't run fast enough due to memory bloat. Also adds a PR report for it, commented out for now.

Adds reference data generation for WID8 on Ukko-GPU and Carrington

Does not add WID=8 CPU TP running.

Should probably be re-reviewed and debugged once these actions have been merged into master.


ykempf:
Should you also activate running the WID8 TP on Ukko, or is that not ready/too slow or sth?

markusbattarbee:
> Should you also activate running the WID8 TP on Ukko, or is that not ready/too slow or sth?

>>Adds WID=8 block size TP on Ukko GPU, but does not yet activate it as it doesn't run fast enough due to memory bloat. Also adds a PR report for it, commented out for now.

The DGX nodes on Ukko have old V100 cards which have a lot less memory, and using WID=8 there's some memory bloat which makes the Ionosphere_small test run waaaay slow due to swapping. It could be activated to only run the smaller tests which do not swap, I guess? 

Goal is to find the blunder(?) in memory allocations at some point, but that might take some time. And I guess at some point we can switch to Mahti / LUMI GPU TP instead of relying on the local cluster.

ykempf:
Well, I think that it's not a bad idea to build and run on as many realistic architectures as available. The local clusters will remain usable for smaller projects, would be silly to have it break there unnoticed. But OK if that's now too slow, all right.

In future, there could be e.g. a nightly check for "all available machines" and a careful selection for the on-push workflows.

ursg:
Merged with current dev to get conflicts out of the way.

ursg:
So, huh... compilation is broken on HILE-G, and it looks like it's the module system's fault. I guess that has to do with the rocm6.3 update on hile?
(Smells like not-an-issue-of-this-PR)

PR #1156, opened 2025-08-11
HILE-C CI scripts

markusbattarbee:


ykempf (commented):
I am also in a certain seminar. But is there the generate reference data now too here?

ursg (approved):

ursg:
I'm now scratching my head about why that run_TP_HILE-C step still fails. It has reference data. It has no actually failing tests. But it still doesn't seem to create an output tar.gz file, probably because *something* had a nonzero return value there.

ursg:
Ah, I guess it must be the awk syntax error in the test_fp runts that triggers it:
```
awk: VLSV file timestamps match.
----------
cmd. line:1: BEGIN{print (!= 0.0)?1:0}
awk: cmd. line:1:              ^ syntax error
awk: cmd. line:1: BEGIN{print (!= 0.0)?1:0}
awk: cmd. line:1:                     ^ syntax error
```
But... weird that this only triggers in those tests on HILE.

ursg:
so the offending code here is
```bash
       # Check if dt is nonzero
       timeDiff=$(grep "delta t" <<< $C |gawk '{print $8}'  )
       if (( $(awk 'BEGIN{print ('$timeDiff'!= 0.0)?1:0}') )); then
           if (( $( echo "${timeDiff#-} $MAXDT" | awk '{ if($1 > $2) print 1; else print 0 }' ) )); then
               MAXDT=$timeDiff
           fi
           echo "WARNING! VLSV file timestamps differ by ${timeDiff}s."
       else
           echo "VLSV file timestamps match."
       fi
       echo "----------"
```
Where apparently ```timeDiff``` gets an empty value. ```$C``` is the run command for diffing two files. Why would that produce an empty output though? Or one that contains no mention of delta t? It works for the other tests!

ursg:
Well, the awk errors are gone now, but there's still stuff of this kind in Flowthrough_damr:
```
Comparing file Flowthrough_damr/bulk.0000004.vlsv against reference
ERROR Datasets have different size.
 proton/vg_rho_0                                                
 proton/vg_v_0                                                  
ERROR Datasets have different size.
ERROR Datasets have different size.
 proton/vg_v_1                                                  
ERROR Datasets have different size.
 proton/vg_v_2                                                  
ERROR Datasets have different size.
 vg_amr_alpha1_0                                                
ERROR Datasets have different size.
```

ursg:
Heh, that was a silly bug: The HILE CI output was never written to a file with the same name who's presence was later checked for. Hence the CI consistently failed there.

ursg:
Alright, I suppose this looks all ok now. The TP run is still marked red, since the fluxfunction didn't find proper reference data, but that will fix itself... eventually. (When this is merged into master)

PR #1157, opened 2025-08-11
Add hyperresistivity term to electric field in fieldsolver

JonasSuni:
Here's the draft PR for the hyperresistivity term, with latest dev merged in and extra unnecessary stuff removed.

I've tested that it compiles (on carrington).


fieldsolver/ldz_hyper.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+   Real Bmag = sqrt((centBgB->at(fsgrids::bgbfield::BGBX)+centPerB->at(fsgrids::bfield::PERBX))*
+            (centBgB->at(fsgrids::bgbfield::BGBX)+centPerB->at(fsgrids::bfield::PERBX)) +
+            (centBgB->at(fsgrids::bgbfield::BGBY)+centPerB->at(fsgrids::bfield::PERBY))*
+            (centBgB->at(fsgrids::bgbfield::BGBY)+centPerB->at(fsgrids::bfield::PERBY)) +
+            (centBgB->at(fsgrids::bgbfield::BGBZ)+centPerB->at(fsgrids::bfield::PERBZ))*
+            (centBgB->at(fsgrids::bgbfield::BGBZ)+centPerB->at(fsgrids::bfield::PERBZ))
+           );
+   rhoq = centMom->at(fsgrids::moments::RHOQ);
+   limitedRhoq = (rhoq <= Parameters::hallMinimumRhoq ) ? Parameters::hallMinimumRhoq : rhoq ;
```
ursg, line 133: Note that these are now simply taken from their locations on the yee-lattice and used "as is", with no balsara interpolation or volumetric averages.

Since we are deaeling with wiggling on the grid-scale here, should we do better than that? Or can we afford to not care, as long as we get the damping order-of-magnitude correct so those wiggles never even grow?

Anyway, it might at least warrent a comment, that the "center" thing there is not the cell centre.

parameters.cpp
```diff
@@ -374,6 +380,10 @@ bool P::addParameters() {
            "The maximum CFL limit for field propagation. Used to set timestep if dynamic_timestep is true.", 0.5);
    RP::add("fieldsolver.minCFL",
            "The minimum CFL limit for field propagation. Used to set timestep if dynamic_timestep is true.", 0.4);
+   RP::add(
+       "fieldsolver.ohmHyperFactor",
+       "Factor to multiply hyperresistivity term by. Default: 1 (damping becomes strong at grid scale)",
```
ursg: For the future uninitiated, it might be helpful here to add a comment of the sort 
"more means larger scales are being damped".

ursg (approved):
Nice! Pretty straightforward implementation, too.

If only we had the fsgrid refactor already in place, so this wouldn't be such a huge amount of boilerplate code when passing the new fsgrid struct everywhere. But such is life.

fieldsolver/ldz_hyper.cpp
JonasSuni: My logic for doing it on the Yee-lattice without any sort of interpolation or averaging was that this is essentially an artificial term, so as long as it does what we need it to, does it need to be physically accurate?

markusbattarbee: Thing is, this code computes a Bmag value and calculates something from that and the charge density - and charge density is a spatial-cell-average value. Thus, to compare apples with apples, it should also calculate Bmag from (Balsara-reconstructed) cell-averages IMO.

(See if the ordering means those values have already been calculated?)

ykempf: Getting edge-rhoq is trivial and likely already done elsewhere (for Hall) hopefully. Getting the Balsara-reconstructed, correct-order (2nd spatial order to be consistent with the remaining things) would probably require Hall-term-grade arithmetics which was famously handled with wxMaxima in its time and made into code with a careful set of sed commands...

fieldsolver/ldz_hyper.cpp
```diff
@@ -0,0 +1,331 @@
...
+   cint& RKCase
+) {
+   //const std::array<int, 3> gridDims = technicalGrid.getLocalSize();
+   const FsGridTools::FsIndex_t* gridDims = &technicalGrid.getLocalSize()[0];
+   const size_t N_cells = gridDims[0]*gridDims[1]*gridDims[2];
+   phiprof::Timer hyperTimer {"Calculate Hyper term"};
+   int computeTimerId {phiprof::initializeTimer("Ehyper compute cells")};
+
+   phiprof::Timer mpiTimer {"Hyper field update ghosts MPI", {"MPI"}};
+   dPerBGrid.updateGhostCells();
```
markusbattarbee: this now leads to a duplicate ghost update of dPerBGrid if both Ohm and Hyperresistivity are active

JonasSuni: Good catch, I had that in at some point but apparently removed it.

Now it's back in

fieldsolver/ldz_hyper.cpp
```diff
@@ -0,0 +1,333 @@
...
+         break;
+
+      case 2:
+         leftdPerB = dPerBGrid.get(i,j,k-1);
+         rightdPerB = dPerBGrid.get(i,j,k+1);
+         break;
+
+      default:
+         cerr << __FILE__ << ":" << __LINE__ << "Derivative component should be 0 (xx), 1 (yy), or 2 (zz)." << endl;
+         break;
```
kstppd, line 71: I would suggest aborting here

fieldsolver/ldz_hyper.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+      case 2:
+         SecondDerOfCurl = (leftdPerB->at(fsgrids::dperb::dPERBydx) + rightdPerB->at(fsgrids::dperb::dPERBydx) - 2 * middPerB->at(fsgrids::dperb::dPERBydx)) / (dPerBGrid.DX * dPerBGrid.DX * dPerBGrid.DX) - 
+         (leftdPerB->at(fsgrids::dperb::dPERBxdy) + rightdPerB->at(fsgrids::dperb::dPERBxdy) - 2 * middPerB->at(fsgrids::dperb::dPERBxdy)) / (dPerBGrid.DX * dPerBGrid.DX * dPerBGrid.DX);
+         break;
+      
+      default:
+         cerr << __FILE__ << ":" << __LINE__ << "Curl component should be 0 (x), 1 (y), or 2 (z)." << endl;
+         break;
```
kstppd, line 97: Same

kstppd (approved):

PR #1165, opened 2025-09-09
Ci docker and artifact version updates

ursg:
A bit of CI housekeeping here: Updating the docker image version for the azure runners, and updating download-artifact's version to v5 to (hopefully) squelch deprecation warnings.


ursg:
Well, this works as intended and makes the CI pipeline break a little bit less.

PR #1168, opened 2025-09-11
At BSC, run risc-v CI on pioneer cluster instead of synth-hca

ursg:
This hopefully fixes our CI failures without being too slow.


ursg:
Looks like this works!

ursg:
Not going to wait for the test package here, this works as intended.

PR #1170, opened 2025-09-17
Added a GPU memory manager

MikaelHuppunen:


arch/gpu_base.hpp
```diff
@@ -273,14 +276,114 @@ struct GPUMemoryManager {
       return true;
    }
 
+   // Create a new pointer, ensure that there aren't duplicates
+   bool createUniquePointer(const std::string& name) {
```
ursg: Is all the GPU memory handling now actually done by doing string lookups at runtime?
Aren't we, potentially, wasting quite a bit of time there now doing string manipulation, instead of for example storing these names in a big enum? Or do some horrible compile-time constexpr thing with them?

The names are, afterall, all known at compile time.

arch/gpu_base.hpp
```diff
@@ -332,66 +739,122 @@ struct GPUMemoryManager {
...
+      int offset = sessionPointerOffset.at(name);
 
-extern Real *returnReal[];
-extern Realf *returnRealf[];
-extern vmesh::LocalID *returnLID[];
-extern Real *host_returnReal[];
-extern Realf *host_returnRealf[];
-extern vmesh::LocalID *host_returnLID[];
+      // If the pointer did not fit inside the session pointer, retrieve it from the separate map
+      if (offset > dev_sessionAllocationSize){
```
ursg, line 785: I shall be nitpicky now and point out that this comparison creates a new signed / unsigned compiler warning.

spatial_cells/block_adjust_gpu.cpp
```diff
@@ -472,45 +486,45 @@ void adjust_velocity_blocks_in_cells(
    // Finds new Blocks (GID,LID) needing to be added
    // Note:list_with_replace_new then contains both new GIDs to use for replacements and new GIDs to place at end of vmesh
    extract_to_add_caller(
-      dev_allMaps+0, // input maps: this is has_content_maps
-      dev_lists_with_replace_new, // output vecs
+      gpuMemoryManager.getPointer<Hashinator::Hashmap<vmesh::GlobalID,vmesh::LocalID>*>("dev_allMaps")+0, // input maps: this is has_content_maps
```
ursg: Why the "+0" there?

ursg (changes requested):
I like this, it makes the memory handling much prettier, but I find the constant usage of runtime strings for memory lookup pretty inelegant here. I'd suggest to replace it by *something* to not require string handling at runtime.

updateGpuMemoryPointerList.sh
```diff
@@ -0,0 +1,86 @@
...
+# Collects all pointers passed to gpuMemoryManager
+# and writes them into ./arch/gpu_base.hpp between
+# //DEFINITIONS_HERE and //DEFINITIONS_END markers.
+
+shopt -s globstar nullglob
+
+declare -A seen
+
+# --- Collect unique identifiers ---
+for file in **/*.cpp **/*.h **/*.hpp; do
```
ursg, line 11: This is such a wonderful hack, I just have to appreciate it.

But it tells a lot about the state of efficient GPU metaprogramming in 2025 that we have to resort to external script mangling like this...

I'm wondering if this will bite use somewhere, as soon as anything about the code structure changes.

ursg (approved):
This actually looks really nice to me now. With the runtime string handling gone now, I like it!

updateGpuMemoryPointerList.sh
MikaelHuppunen: It shouldn't break easily since it should go through all .cpp .h and .hpp files (except libraries and submodules) regardless of the structure. If you move gpu_base.hpp somewhere else then you'll have to change that path though.

PR #1171, opened 2025-09-22
Dev with asterix optional

kstppd:
# Asterix Compression

Asterix provides lossy compression for VDFs. Current compression methods:
- `ZFP`
- `MLP_MULTI`
- `OCTREE`

TODO:   
- `HERMITE` which is pretty much ready but maybe will be added in a future PR.

**All methods work for both CPU and GPU Vlasiator. However running the MLP on CPUs is strongly discouraged for performance reasons**

The install libraries scripts will install ZFP and OCTREE automagically but the MLP has to be installed manually. It requires either `nvcc` or `hipcc` to be installed on the system. For instructions look at the [Asterix](https://github.com/fmihpc/asterix) repo.

```bash
pip install meson
git clone https://github.com/fmihpc/asterix
cd asterix/  
mkdir subprojects
meson wrap install gtest
meson wrap install curl
meson wrap install spdlog 
meson wrap install fmt
meson setup build  --buildtype=release -Dtiny_cuda_arch_code=70/80/90
meson compile -C build --jobs=8
meson test -C build
```
Some one-liner that is good for NVIDIA systems and saves you from the build system havoc:    
```
cd asterix/
nvcc src/vdf_compressor_nn.cu -DUSE_GPU -DSKIP_HOSTBLAS  -DTINYAI_MEMORY_GB=64  -DNOPROFILE --std=c++20 -I./include -arch=sm_86  --shared  -Xcompiler=-fPIC -o libvlasiator_vdf_compressor_nn.so -lcublas 
```

The compression method is selected at runtime with `--Asterix.method`.

## Compile time Flags
Compression methods are optional and enabled at compile time.

```bash
-DASTERIX_ZFP
-DASTERIX_OCTREE
-DASTERIX_MLP
```

Example: For an example look into MAKE/Makefile.kstppd

## Compression Parameters

```text
--Asterix.mlp_layers
    Hidden layer architecture for the MLP.
    Example: 32,32,16

--Asterix.tol
    Tolerance used by the MLP compressor.

--Asterix.octree_tolerance
    Tolerance used by OCTREE compressor.

--Asterix.max_epochs
    Maximum MLP training epochs to be used.

--Asterix.fourier_order
    Fourier feature order used by the MLP compressor.

--Asterix.interval
    Compression interval in seconds [Deprecated and unused].

--Asterix.state
    Enable or disable compression. 
    0 = disabled
    1 = enabled

--Asterix.method
    Compression method to use.
    Example: ZFP, MLP_MULTI, OCTREE

--Asterix.max_vdfs_per_nn
    Maximum number of VDFs grouped into one neural network
    in multi-regression mode.
```

## I/O Compression Options
These options enable compression for different output file types.

```text
--io.system_write_distribution_compressed 
    Compress VDFs written to bulk distribution files.

--io.system_write_restart_compressed
    Compress VDFs written to restart files.

--io.system_write_recovery_compressed
    Compress VDFs written to recovery files.
```


ursg:
Can we get an update on the functionality here? Is this PR now in such a state that no specific Asterix features are enabled by default, so that we can "just merge it"? (Hmm, working CI would really help with this)

vlasovsolver/vec.h
```diff
@@ -220,7 +220,6 @@ typedef Vec16fb Vecb;
 #define VECL 16
 #define VPREC 4
 #ifndef VEC_PER_PLANE
-const int VEC_PER_PLANE = (WID*WID/VECL);
 #endif
```
petschge, line 224: why remove that line?

spatial_cells/spatial_cell_gpu.hpp
```diff
@@ -624,6 +628,7 @@ __global__ static void resize_and_empty_kernel (
 
       // Following functions adjust velocity blocks stored on the cell //
       void adjustSingleCellVelocityBlocks(const uint popID, bool doDeleteEmpty=false);
+      bool add_velocity_block(const vmesh::GlobalID& block,const uint popID);
```
petschge, line 631: Don't we use a lovely mix of snake_case and CamelCase?

submodules/fsgrid
petschge, line 1: This is going to be so much fun to merge with the fsgrid changes...

ioread.cpp
```diff
@@ -354,22 +365,9 @@ bool checkScalarParameter(vlsv::ParallelReader& file,const string& name,T correc
    }
 }
 
-/** Read velocity block mesh data and distribution function data belonging to this process
```
petschge, line 357: Why does that doc-comment get removed?

memory_report.cpp
```diff
@@ -100,7 +100,8 @@ uint64_t get_node_free_memory(){
             }
          } else {
             // Skip to neww line
-            fscanf(in_file, "%*[^\n]\n");
+            auto _id=fscanf(in_file, "%*[^\n]\n");
+            (void)_id;
```
petschge: Is silencing the compiler warning really worth this? Maybe we should check the return value and act on it? And we might as well fix the typo in the comment while we're at it.

vlasiator.cpp
```diff
@@ -1036,7 +1040,7 @@ int simulate(int argn,char* args[]) {
                   "restart",
                   (uint)P::t,
                   true, // add the date of the file to the name
-                  P::restartStripeFactor) == false ) {
+                  P::restartStripeFactor,P::systemWriteRestartCompressed) == false ) {
```
petschge: one line per argument?

vlasiator.cpp
```diff
@@ -1072,7 +1076,7 @@ int simulate(int argn,char* args[]) {
                   "recover",
                   recoverCounter % P::recoverMaxFiles,
                   false, // overwrite so do not put date in file name
-                  P::restartStripeFactor) == false ) {
+                  P::restartStripeFactor,P::systemWriteRecoveryCompressed) == false ) {
```
petschge: one line per argument?

vlasovsolver/vec.h
kstppd: I guess this was accidental. Reverting

spatial_cells/spatial_cell_gpu.hpp
kstppd: Hmmm worth changing? We have this mix everywhere :)

submodules/fsgrid
kstppd: This was accidental and has now been reverted! Thanks!

PR #1172, opened 2025-09-23
Enable precommit CI auto-stage on PRs to format source code

ursg:
This will now (hopefully) autoformat in a reasonable way, without messing up omp pragmas.


ursg:
Thanks to github's hilariously convoluted security model, this PR's new Github Action does of course *not* run, unless it is already present on the target branch. Which it is not. So I shall now create a temporary precommitTest branch on fmihpc for testing this.

PR #1173, opened 2025-09-23
Trivial little change to test the PR CI running

ursg:


ursg:
So... now there is *no action* running on this. But why not? Is it because the action is not yet set up to trigger for the target branch, in the target branch? Let's test that!

PR #1176, opened 2025-09-23
Precommit test

ykempf:


ykempf:
This maketh no sense. In here I got the "approve workflow" button (after I had got both precommit entries red, then a while later the github GUI had all green, I hit F5 and the approval thingy appeared, and that now succeeded apparently. The new commit having been pushed by it before my approval of the workflow, anyway.

ykempf:
Yet in #1177 the same precommit thing fails.

PR #1182, opened 2025-09-24
Precommit test

ykempf:


arch/arch_device_cuda.h
```diff
@@ -120,81 +120,75 @@ extern cudaStream_t gpuStreamList[];
...
-      T *ptr;
-      T *d_ptr;
+      T* ptr;
+      T* d_ptr;
       uint bytes;
       uint is_copy = 0;
       uint thread_id = 0;
 
    public:
+      void syncDeviceData(void) { CHK_ERR(cudaMemcpyAsync(d_ptr, ptr, bytes, cudaMemcpyHostToDevice, gpuStreamList[thread_id])); }
```
ursg, line 188: I don't like that these have been compressed to one-liners now. I wonder what clang-format option would help there?

sysboundary/ionosphere.h
```diff
@@ -171,82 +168,58 @@ namespace SBC {
...
+      std::map<std::array<Real, 3>, std::array<std::pair<int, Real>, 3>> vlasovGridCoupling; /*!< Grid coupling information, caching how vlasovGrid coordinate couple to ionosphere data */
 
-      void setDipoleField(const FieldFunction& dipole) {
-         dipoleField = dipole;
-      };
-      void setConstantBackgroundField(const std::array<Real, 3> B) {
-         BGB = B;
-      }
+      void setDipoleField(const FieldFunction& dipole) { dipoleField = dipole; };
+      void setConstantBackgroundField(const std::array<Real, 3> B) { BGB = B; }
```
ursg, line 174: Here, too: The threshold for single-lining simple functions is too small, IMHO

sysboundary/ionosphere.h
```diff
@@ -171,82 +168,58 @@ namespace SBC {
...
-         Real & minPotentialN,
-         Real & maxPotentialN,
-         Real & minPotentialS,
-         Real & maxPotentialS
-      );
+      void initSolver(bool zeroOut = true);                                      /*!< Initialize the CG solver */
+      iSolverReal Atimes(uint nodeIndex, int parameter, bool transpose = false); /*!< Evaluate neighbour nodes' coupled parameter */
+      Real Asolve(uint nodeIndex, int parameter, bool transpose = false);        /*!< Evaluate own parameter value */
+      void solve(int& iteration, int& nRestarts, Real& residual, Real& minPotentialN, Real& maxPotentialN, Real& minPotentialS, Real& maxPotentialS);
+      void solveInternal(int& iteration, int& nRestarts, Real& residual, Real& minPotentialN, Real& maxPotentialN, Real& minPotentialS, Real& maxPotentialS);
```
ursg: These, too, are less readable now than before.

sysboundary/ionosphere.h
```diff
@@ -171,82 +168,58 @@ namespace SBC {
...
-         FsGrid< std::array<Real, fsgrids::bfield::N_BFIELD>, FS_STENCIL_WIDTH> & perBGrid,                                                                                                                                                                                                                                
-         FsGrid< std::array<Real, fsgrids::dperb::N_DPERB>, FS_STENCIL_WIDTH> & dPerBGrid,
-         FsGrid< std::array<Real, fsgrids::moments::N_MOMENTS>, FS_STENCIL_WIDTH> & momentsGrid,
-         FsGrid< std::array<Real, fsgrids::volfields::N_VOL>, FS_STENCIL_WIDTH> & volGrid,
-         FsGrid< fsgrids::technical, FS_STENCIL_WIDTH> & technicalGrid
-      );
-      
+      void mapDownBoundaryData(FsGrid<std::array<Real, fsgrids::bfield::N_BFIELD>, FS_STENCIL_WIDTH>& perBGrid, FsGrid<std::array<Real, fsgrids::dperb::N_DPERB>, FS_STENCIL_WIDTH>& dPerBGrid,
+                               FsGrid<std::array<Real, fsgrids::moments::N_MOMENTS>, FS_STENCIL_WIDTH>& momentsGrid, FsGrid<std::array<Real, fsgrids::volfields::N_VOL>, FS_STENCIL_WIDTH>& volGrid,
+                               FsGrid<fsgrids::technical, FS_STENCIL_WIDTH>& technicalGrid);
```
ursg: Wasn't this exactly what the BinPackArguments option was supposed to prevent? Or is this now only affecting arguments in the function definition, and is separately handled for declarations?

datareduction/datareducer.cpp
```diff
@@ -47,507 +44,502 @@ void initializeDataReducers(DataReducer * outputReducer, DataReducer * diagnosti
...
-                     for(int x=0; x<gridSize[0]; x++) {
-                        retval[3*(gridSize[1]*gridSize[0]*z + gridSize[0]*y + x)] =     (*BgBGrid.get(x,y,z))[fsgrids::BGBX]
-                           + (*perBGrid.get(x,y,z))[fsgrids::PERBX];
-                        retval[3*(gridSize[1]*gridSize[0]*z + gridSize[0]*y + x) + 1] = (*BgBGrid.get(x,y,z))[fsgrids::BGBY]
-                           + (*perBGrid.get(x,y,z))[fsgrids::PERBY];
-                        retval[3*(gridSize[1]*gridSize[0]*z + gridSize[0]*y + x) + 2] = (*BgBGrid.get(x,y,z))[fsgrids::BGBZ]
-                           + (*perBGrid.get(x,y,z))[fsgrids::PERBZ];
-                     }
+      for (auto& c : lowercase)
+         c = tolower(c);
```
ursg: This should definitely have curlies.

datareduction/datareducer.cpp
ursg, line 1: The formatting of all these datareducers, that I had been quite vocal about in the fsgrid PR, now actually looks pretty nice!

ursg (changes requested):
I only took some picks of files here and there (and did not scroll through 280 files in 10 minutes!). There are definitely some clang-format options that still need tuning.

arch/arch_device_cuda.h
ursg: I guess ```AllowShortFunctionsOnASingleLine=Empty``` would be a reasonable setting for this?

ykempf: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#allowshortblocksonasingleline

sysboundary/ionosphere.h
ykempf: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#allowshortblocksonasingleline

sysboundary/ionosphere.h
ursg: Ah, it is ```BinPackArguments``` (which are the actual values supplied in a function invocation) vs. ```BinPackParameters``` (which is the parameter list in the function definition and/or declaration.

sysboundary/ionosphere.h
ykempf: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#binpackarguments

sysboundary/ionosphere.h
ykempf: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#binpackparameters

ursg: There is also "ExperimentalAutoDetectBinPacking", but I have no idea what it does.

datareduction/datareducer.cpp
ykempf: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces

ursg:
scrolling through the config reference, i see ```ForEachMacros```, that tells clang-format which custom macros should be treated as for loops. Our arch looping mechanisms are candidates for that.

ykempf:
In other news, we have a successful pair of TP runs! :tada:

sysboundary/ionosphere.h
lkotipal: This might be a consequence of having too permissive line width, I'm assuming it's only going to split calls and definitions if they don't fit

ykempf: This and the below are better imho if forcing them to just have one per line in a blanket fashion. Having variably one or more per line is just going to confuse people.

lkotipal: Yeah `BPPS_AlwaysOnePerLine` is probably _fine_ if not ideal imo. I don't think I agree about it being case-by-case being bad, like 1-3 param functions can be single-line. How does the formatting look with a lower line width?

iowrite.h
```diff
@@ -42,17 +42,57 @@
...
+         fsgrids::bgbfield::N_BGB>,
+      FS_STENCIL_WIDTH>& BgBGrid,
+   FsGrid<
+      std::array<
+         Real,
+         fsgrids::volfields::N_VOL>,
+      FS_STENCIL_WIDTH>& volGrid,
+   FsGrid<
+      fsgrids::technical,
+      FS_STENCIL_WIDTH>& technicalGrid,
```
lkotipal: Oh my god bruh

alhom: Tbh this particular mess could do with a proper fsgrids wrapper anyhow.

sysboundary/ionosphere.h
lkotipal: Well, this turned out to be absolutely awful as it applies the same rule to template parameters which we absolutely do not want one per line, ever. So we'll have to live with `BPPS_OnePerLine` which will occasionally result in these one-liners (we'd have to tone down line width to address that)

common.h
```diff
@@ -32,85 +32,73 @@
...
+   if ((x) != (x)) {                                                                                                                                                                                   \
+      std::cerr << __FILE__ << ":" << __LINE__ << " Illegal value: " << x << std::endl;                                                                                                                \
+      abort();                                                                                                                                                                                         \
    }
 #else
-#define CHECK_FLOAT(x) {}
+#define CHECK_FLOAT(x)                                                                                                                                                                                 \
+   {                                                                                                                                                                                                   \
+   }
```
ursg: This is a weird way to reformat this. It seems to identify the "{}" here as an empty code block, but this totally breaks the ideomatic use of {} to signify an empty piece of code.

Also, why does it even touch this line, if the regex forbids it from touching anything with a # in it?

ursg:
To sum up our findings of today:
a) By and large, the formatting is now looking good.
b) We manually disabled clang-format for some code blocks that are semantically grouped for ease of understanding
c) In principle, we are now happy and would merge it, but it looks like we are triggering a clang-format bug with our regex-ignoring of pragmas: many (all?) comment lines preceding a pragma are alternatingly moved to column one and back to their proper indentation on every even-odd run of clang-format.

Do we have a workaround for that?

ursg:
So your solution is to have a "#" in every comment in front of an omp pragma, hence clang-format ignoring it?
And if we forget that in the future, precommit will hapilly mess up our code again?

"No sir, I don't like it."

PR #1189, opened 2025-10-04
Change vectordipole dipoleInflow magnetic field split into BGB and PerB

### LINKED ISSUE #1188: Magnetospheric simulations using dipoleInflowBXYZ instead of putting IMF everywhere misbehave when IMF is changed
When injecting a magnetic field rotation with corresponding velocity change into a magnetospheric simulation with an initial field set using vector dipole and the dipoleInflowBXYZ parameters, the RD is accompanied by anomalous velocity enhancement in the direction of the velocity change as well as enhanced magnetic field at the y or z boundaries (depending on the geometry of the initial IMF and rotation).

If instead one uses the options to set the initial IMF everywhere in the box on top of the dipole, the issue does not occur. 

Example images with dipoleInflowBXYZ (left) and IMF everywhere (right)

My uneducated guess is that the issue is related to the IMF being put in PerB in the former case and BGB in the latter. A potential solution to the issue could be to put constant IMF in BGB also when using dipoleInflowBXYZ, and then correcting for the tapering behaviour in PerB

JonasSuni:
This fixes #1188 

Left: Old scheme, Right: New scheme
(Note: Both schemes also remove the use of finite differencing at sysboundaries, but that is not included in this PR)

### DIFF
--- a/backgroundfield/vectordipole.cpp
+++ b/backgroundfield/vectordipole.cpp
@@ -73,9 +73,9 @@ double VectorDipole::operator()( double x, double y, double z, coordinate compon
       return 0.0; //set zero field inside dipole
    
    if(r[0]>=xlimit[1]){
-      //set zero or IMF field and derivatives outside "zero x limit"
+      //set zero field and derivatives outside "zero x limit", leaving IMF there
       if(derivative == 0) {
-         return IMF[component]; 
+         return 0.0; 
       } else {
          return 0.0;
       }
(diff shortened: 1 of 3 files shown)


JonasSuni:
Ah, I should probably have checked this compiles first :P

JonasSuni:
Now it should compile.

The second commit changes dipoleInflowB from a Real[3] to std::vector<Real, 3>

JonasSuni:
A caveat to this is that if the PerB correction terms are not applied at the inner boundary, then there will be an anomalous IMF remaining in the IB cells. 

More investigation is needed

JonasSuni:
The inner boundary seems to be behaving perfectly well. 

However, I was informed yesterday that having the IMF in PerB is the sensible way to go [M. Battarbee, private communication, 2025].

So maybe the effort should be focused on solving the PerB bug instead of this workaround

PR #1191, opened 2025-10-05
Warn when field line tracing is enabled and not in a Magnetosphere project

### LINKED ISSUE #1185: Field line tracing fails with flowthrough project
(and probably other non-magnetosphere projects as well.)

If you run a Flowthrough project but the config file has options for field tracing, the run fails with `std::bad_function_call`.

kstppd:
This PR tries to resolve #1185 . 
I could actually not figure out any way to do the same using Boost's program options so I had to do some manual parsing. In any case this will now wart on ```stderr``` when field tracing is enabled and we are not running a Magnetosphere. The user can still however specify datareducers which will trigger field line tracing. Should we guard that too?

### DIFF
--- a/MAKE/Makefile.kstppd
+++ b/MAKE/Makefile.kstppd
@@ -68,8 +68,8 @@ LIB_PAPI = -L$(LIBRARY_PREFIX)/lib/ -lpapi
 INC_JEMALLOC = -I$(LIBRARY_PREFIX)/jemalloc/include
 LIB_JEMALLOC = -L$(LIBRARY_PREFIX)/jemalloc/lib -ljemalloc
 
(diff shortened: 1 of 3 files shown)


JonasSuni:
I think enabling those datareducers in projects where field line tracing doesn't work should just not do anything (aside from printing a warning in stderr) and not write out the data reducers. Certainly it should not enable field tracing.

kstppd:
Yeah maybe we should come up with a ruleset there.

alhom:
As Yann was saying in the issue, flux rope tracing is still a thing that a user could want to have on in non-Magnetosphere runs. Plus there are other applications that can be built on tracing, like squash factor calculations.

PR #1193, opened 2025-10-16
Reapply initial state in boundary cells after setProjectBField

JonasSuni:
This is necessary when reapplying boundaries at restart for projects that store correction terms in BGB as these correction terms need to be calculated by setProjectBField before PerB can be set correctly in the boundaries.


ursg (approved):
Goodygood. Note to the future: This probably will need a proper workthrough (and an illustration of the actual control flow at initialization) to disentangle what-gets-set-where-in which order.

But this PR is not the point to create that.

PR #1194, opened 2025-10-16
Do not reapply solved B components in boundary cells except at simulation start

JonasSuni:
The argument being that B evolves in these cells and thus resetting B during runtime is suspect.

### DIFF
--- a/sysboundary/inflow.cpp
+++ b/sysboundary/inflow.cpp
@@ -155,10 +155,15 @@ namespace SBC {
       for (uint popID = 0; popID < getObjectWrapper().particleSpecies.size(); ++popID) {
          setCellsFromTemplate(mpiGrid, popID);
       }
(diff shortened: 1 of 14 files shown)


sysboundary/inflow.cpp
```diff
@@ -155,10 +155,15 @@ namespace SBC {
       for (uint popID = 0; popID < getObjectWrapper().particleSpecies.size(); ++popID) {
          setCellsFromTemplate(mpiGrid, popID);
       }
-      setBFromTemplate(mpiGrid, perBGrid, BgBGrid);
+      if (P::isRestart) {
+         setBFromTemplate(mpiGrid, technicalGrid, perBGrid, BgBGrid, false);
```
alhom, line 159: Just pass `!P::isRestart`?

ursg (approved):
This is one of those "how did this ever work?" kind of things

PR #1196, opened 2025-11-07
Added weak scaling scripts

MikaelHuppunen:
Automatically generates .cfg and .sh files for a weak scaling test with given base .cfg and .sh files. Then launches the .sh scripts using sbatch and outputs the results in the corresponding folders.


ykempf:
How about moving that under scaling/ instead of samples/?

PR #1197, opened 2025-11-14
Fixed a bug in GPU memory manager

MikaelHuppunen:
Corrected calculation of maxPointerIndex in sub pointer creation.


ursg:
Thumbs up from me.
Surprisingly large amount of red in CI there though, is that all just caused by the cluster downtimes?
Trying to rerun anyway. Probably makes sense to merge regardless.

MikaelHuppunen:
> Surprisingly large amount of red in CI there though, is that all just caused by the cluster downtimes?

Seems to be the case.

PR #1199, opened 2025-11-21
Improvements to GPU memory manager

MikaelHuppunen:
make now automatically runs the script for updating GPU memory pointers, so it doesn't need to be run manually. Set alignment in session pointer to at least 256 bits.


ursg (approved):
Ok, much more elegant than running it by hand!

PR #1200, opened 2025-11-25
Viiki migration slurm command update

ursg:
now that carrington and ukko are migrated to Viikki, the clusters have all been integrated together, with kale, ukko and carrington nodes sitting happily side-by-side in the same rack.

This means that no explicit "-M carrington" is required anymore. Not sure if we *want* to even explicitly pin our jobs to the carrington nodes, but would maybe be ok to run wherever?

Anyway, this PR will first show us whether running like this works in the first place.


ursg:
Hooray, semi-good news: By running it on ukko instead of carrington, the testpackage now actually compiles and runs again!

ursg:
Obsoleted by newer CI fix PRs.

PR #1201, opened 2025-12-02
Add Mimalloc as an alternative to jemalloc

ursg:
Will this lead to improved performance? Less memory use? Better portability? Let's find out!

(Actually, this first commit only tests compilation on Github Actions, testpackage build and test coming up next.)


ursg:
Tested with a random ionosphere test restart at scale (100 nodes of mahti):
mallocs
Jemalloc appears better, both in terms of memory HWM *and* runtime:

Jemalloc: total run time 2643.68 s
System allocator: total run time 2595.32 s
Mimalloc: total run time 2717.11 s

So neither any performance, nor memory benefit.
I'm still in favour of merging this, because choice is good. Who knows how this will go on future architectures.

PR #1202, opened 2025-12-05
Copy fsgrid moments to outflow after filtering them

JonasSuni:
Since boundary cells are not filtered, we should copy fsgrid moments to outflow cells after each filter pass to retain consistency of the copy conditions.


ykempf:
Should this include a mechanism to bypass when no Outflow is present?

JonasSuni:
Could be done. Do you think looping over local cells to check whether they're outflow is that slow though?

ykempf:
You could set some flag on first pass or so, but you could also store some class member variable when the outflow  constructor is called? @ursg or @lkotipal will have good suggestions I'm sure.

ursg (approved):

PR #1203, opened 2025-12-05
changed the outdated library paths in MAKE files to be in /wrk-vakka/

lassejsc:
path /proj/ no longer works on the nodes


lassejsc:
This is outdated currently

PR #1209, opened 2025-12-17
Workflow check for whether to run testpackage

lassejsc:
The workflow will now skip the testpackage runs if only documentation has changed.

This adds two new steps for the workflow: `get_git_diff` and `get_git_diff_check`. During first one the new script `CI_git_diff.sh` is run to get the git diff between the PR branch and its target.  The script uses grep with regex to check whether only documentation files have changed. If it is so, then running of the testpackage is skipped. 

`get_git_diff_check` acts merely as an if statement that checks if the `get_git_diff` tells to run the tests.

### DIFF
--- a/.github/workflows/CI_git_diff.sh
+++ b/.github/workflows/CI_git_diff.sh
@@ -1,7 +1,7 @@
 
 #!/bin/bash 
 
-git_diff=$(git diff --name-only upstream/vlasiator_gpu...)
+git_diff=$(git diff --name-only origin/dev...)
 declare -i full_count=$(echo $git_diff | wc -w)
 declare -a unallowed_strings=(".cpp\b" ".hpp\b" ".c\b" ".sh\b" )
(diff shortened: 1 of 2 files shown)


.github/workflows/CI_git_diff.sh
```diff
@@ -0,0 +1,27 @@
...
+
+#Check if git diff contains any changes to the actual code
+for unallowed_string in ${unallowed_strings[@]}
+do
+  if [[ $(echo $git_diff | grep -Po $unallowed_string) ]]; then
+    exit 0
+  fi
+done
+
+declare -a file_check=("\w+.md" "INSTALL" "CITATION.cff" "Doxyfile." "\bdoc/")
```
alhom: These can be skipped as well:
`.vscode/`
`doxytmp/`
`scaling/`
`visit_variables/`

lassejsc: added

PR #1212, opened 2026-01-30
Timevarying inflow 3D test

lassejsc:
Timevarying inflow test that produces a rotational discontinuity where initially the flow is set to be along Y-axis and after 3 seconds it flips to be entirely on Z-axis.  Currently writes 3 bulk files and simulates 30 seconds

On carrington: total run time 88.5888 s, total simulated time 30.2 s
HILE: total run time 38.6716 s, total simulated time 30.2 s

Example plots of vg_v along z
2_z_map_proton_vg_v_z_0000002
1_z_map_proton_vg_v_z_0000001
0_z_map_proton_vg_v_z_0000000

Example plots of vg_v along y
2_z_map_proton_vg_v_y_0000002
1_z_map_proton_vg_v_y_0000001
0_z_map_proton_vg_v_y_0000000


JonasSuni:
Regarding the idea of baking this test into the ionosphere test or magnetosphere3d, that's not going to work because those have purely southward imf afaik, and for this test we need the Bx component

PR #1214, opened 2026-02-10
Added writing of mesh domain extents 

lassejsc:
Added a function to write the domain extents, i.e the smallest box to fit all the cells that are in the domain. Does same thing as what analysator funnction [get_mesh_domain_extents()](https://github.com/fmihpc/analysator/blob/e5738edd15418c584997dedcf4aaed43aff713fd/analysator/pyVlsv/vlsvreader.py#L4291) is supposed to do apart from the reshaping of the return value.

### DIFF
--- a/iowrite.cpp
+++ b/iowrite.cpp
@@ -643,24 +643,20 @@ bool writeDomainSizes( Writer & vlsvWriter,
 
 bool writeDomainExtents(Writer& vlsvWriter, const string& meshName, const std::vector<CellID>& local_cells,
                         dccrg::Dccrg<SpatialCell, dccrg::Cartesian_Geometry>& mpiGrid) {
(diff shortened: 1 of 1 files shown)


ursg (approved):

ursg:
The dataset written by this PR is now called MESH_DOMAIN_EXTENTS, and not MESH_DOMAIN_SIZES, that the analysator code is looking for. Is that deliberately chosen to be different?

lassejsc:
> The dataset written by this PR is now called MESH_DOMAIN_EXTENTS, and not MESH_DOMAIN_SIZES, that the analysator code is looking for. Is that deliberately chosen to be different?

Analysator function I linked does the same thing as what is now being written into MESH_DOMAIN_EXTENTS, the MESH_DOMAIN_SIZES is a different thing.

PR #1215, opened 2026-02-11
Changing -M to --constraint for slurm command calls to fix github_CI

lassejsc:
the -M flag causes errors as it is no longer available on the cluster.


MAKE/Makefile.ukko_dgx
```diff
@@ -71,10 +71,10 @@ VLASIATOR_LIBRARY_PREFIX = $(CURDIR)/libraries-carrington
 # Both ukko-dgx and ukko-a100 use the same modules and libraries
 
 LIB_PROFILE = -L$(VLASIATOR_LIBRARY_PATH)/lib -lphiprof
-INC_PROFILE = -I$(VLASIATOR_LIBRARY_PATH)/include
+INC_PROFILE = -isystem $(VLASIATOR_LIBRARY_PATH)/include
```
lassejsc: Oops meant -I -> -isystem, this was also changed for carrington makefile so i assume this makes the compiler find the headers

lassejsc: I guess -I also does that as well, goes to show build systems are bit alien to me still

ursg:
Even though not everything is green yet, this PR improves CI to the point that I'd say it makes sense to merge it immediately.

PR #1217, opened 2026-02-20
Some const correctness fixes

lassejsc:
Checking const correctness, touches a lot but maybe not everything, mostly did not delve too deep into the GPU side of things, though did check some files and they seemed okay.

If the value given as a parameter be it pass by reference or pointer is not changed, it is marked as `const`.

I will try to commit these file by file so it would be easier to undo something later, this should not change how the code functions.


lkotipal:
> Not sure if pass-by-values should be const or not

As per the [Bible](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rconst-immutable) it's fine if they are not

PR #1234, opened 2026-03-13
Build libraries for compilation when generating reference data

JonasSuni:
I have no idea what I'm doing, this is the first time I've touched CI.


ursg (approved):
s

PR #1247, opened 2026-03-17
Removed boost build from hile gpu for now since that did not build on hile

lassejsc:
Removed boost build and using the one in /usr/lib64

Boost just refuses to build on hile I guess.

### DIFF
--- a/.github/workflows/github-ci.yml
+++ b/.github/workflows/github-ci.yml
@@ -850,7 +850,7 @@ jobs:
         tar -czf testpackage-output.tar.gz testpackage_check_description.txt testpackage_output_variables.txt
         MORO
         )
-        srun --job-name CI_package_results --constraint "carrington|ukko" -N 1 -c 1 --mem=1G bash -c "$PARSE_OUTPUT_CMD"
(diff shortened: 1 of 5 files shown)


lassejsc:
This should work right now, I think the run_TP failing is something else, but I guess should be tested after that has been fixed

.github/workflows/github-ci.yml
```diff
@@ -840,7 +840,7 @@ jobs:
         # Delete any old run directories
         rm -rf run_20*
         export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$GITHUB_WORKSPACE/libraries-carrington/lib
-        sbatch -W -o testpackage_run_output.txt ./small_test_carrington_github_ci.sh
+        sbatch -W -o testpackage_run_output.txt ./small_test_carrington_github_ci.sh || (srun --constraint="ukko|carrington" cat testpackage_run_output.txt || cat testpackage_run_output.txt )
```
ursg: Oh what a wonderfully ugly workaround for our filesystem being unreliable...

ursg (approved):

.github/workflows/github-ci.yml
lassejsc: Realized this was actually unnecessary, better fix was to add `|| true` so it continues past this and prints the log.

PR #1251, opened 2026-03-19
Dispersion Plotter

petschge:
Better user experience without tqdm (or when PTNOINTERACTIVE=1, e.g. in a job script).
Support for high frequency modes
more quantities are considered for plotting


ursg:
I see this also includes #1245, which is probably not on purpose?
(But anyway a good idea, so why not!)

ursg (approved):

ursg:
I suppose this can be closed, since it is superseded by #1325?

PR #1252, opened 2026-03-26
Schedule CI runs for master weekly...

alhom:
... to 1) make use of the Carrington reservation and 2) ensure CI keeps functioning.

Noting that the scheduled jobs _only_ run on the default branch, i.e. `master` for now. Should the default be `dev` instead, even in general?


JonasSuni (approved):
Looks good and shouldn't break anything

PR #1258, opened 2026-04-13
Update arch_dt.cpp

MaxenceAnt:
Modification of
 blockContainer->getParameters(popID)
 to 
blockContainer->getParameters()

The previous version caused problems when computing dt for more than one population, as this could result in extremely low values.


MaxenceAnt:
Modification of
blockContainer->getParameters(popID)
to
blockContainer->getParameters()

The previous version caused problems when computing dt for more than one population, as this could result in extremely low values.

PR #1259, opened 2026-04-13
Update vlasovmover.cpp

MaxenceAnt:
The adjustVelocityBlocks in the subcycle loop has been modified for acceleration. The previous version caused "bailing out at vlasovsolver/cpu_acc_map.cpp".  The new version may slightly increase computational time if there are a lot of subcycles.


MaxenceAnt:
The adjustVelocityBlocks in the subcycle loop has been modified for acceleration. The previous version caused "bailing out at vlasovsolver/cpu_acc_map.cpp". The new version may slightly increase computational time if there are a lot of subcycles.

PR #1262, opened 2026-04-14
Do not exit testpackage when fluxfunction files differ

JonasSuni:
Ugly hack to exit only when diff exits with code 2, otherwise proceed


ursg (approved):
It's horrible, I like it!

PR #1268, opened 2026-04-22
CLI11

lassejsc:
Very badly WIP

Idea is to replace boost program options and throw boost out and replace it with CLI11, which only need a single header file (very nice!) 

(This has been quite awful to replace)

Current status: All of the testpackages should run (?) at least most, mostly, maybe....

Some stuff to note:

syntax of the cfg files stays about the same apart from for example multipeak. parameters now needing to be inputted as a list instead of the old way of having multiple lines of `parameter=value`  (This still works for output for example currently, not sure why it doesnt work for the multipeak parameters)

Help and version currently don't work as I've been working on first getting it to actually read the parameters correctly..
(Currently touches files in testpackage that will be reverted later)


lassejsc:
this was too much of a hottake, see #1269 for a less controversial version

PR #1269, opened 2026-04-23
Less controversial CLI11 PR

lassejsc:
The gist of the CLI11 is that the `Readparameters::add` now takes in the variable to which the value should be assigned alongside the default value. There is no longer a need for `Readparameters::get` and as such there will be a lot of lines removed. 

The class level `addParameters` and `getParameters` remain, former is used for adding the parameters, so `Readparameters::add` calls and the latter is used for processing those parameters, see for example `getPopulationParameters` in `object_wrapper.cpp`.

A sort of meta level logic of how the parameters go is that we first call the `::add`s , which in most cases creates a vector of pointers of for example `particleSpecies` (see `object_wrapper.cpp`) which are later read into a vector of  species values. This is due to the fact that I need to pass a pointer to the parameter value, within say a struct , to CLI11. Initially I tried to do this with making a pointer to a variable within the vector of values, but this has the issue of getting nullified if the vector changes size, say when we push new species to the vector.  

This pattern of `::addParameters`(within which `Readparameters::add`s are called) call before `parse()` and `::getParameters` call after it gets repeated to almost all the files that added some parameters i.e stuff in `projects/`, `sysboundary/` , `object_wrapper.cpp` and `parameters.cpp`

Other notes:
We still do a double parse(actually now quadruple parse, but it seems fast) since to initialize all the particle populations for example, we need to know how many populations we have etc, if this was done during one read it would need to be done at specific time, for example before reading and initializing the project as that may add population specific parameters. I tried to have it as single parse but I couldn't get it to work, I am not 100% it can be done with CLI11 but it's not a big issue.

Afaik this touches almost nothing that happens in the actual meat of the simulation.

NOTE: Comments besides subcommands in the config files doesn't seem to work, you can blame CLI11 for that, minor thing though but do not put
```
[something] #comment
value=bla
```

Found another quirk that need to be accounted for: Mainly you cannot have multiple sections/subcommands in the config since that is just how CLI11 works, for example
```
[proton_outflow]
face = y-
face = y+
face = z-
face = z+

[proton_outflow]
face = x-
```
 will complain about "Boundary 1 is invalid", which should be fixed too and is related to the overall issue of checking the validity of the boundary conditions. (Will see if i have time to fix this, it now should print an error about invalid boundary settings in some direction)
 
 
 Additionally there is `--check_cfg` option which checks the cfg for validity, basically just does the parameter reading as usual but quits before it starts the execution. Usage: ``vlasiator --check_cfg --run_config="bla.cfg"``
 
 # Vlasiator parameter reading some descriptive notes

Parameter handling is divided into two functions: adders and getters. Adders add the parameters to the `CLI::App` as `CLI::Option`s. Parameters with dot add subcommands to the app, for example `gridbuilder.dt` adds a subcommand `gridbuilder` (instance of `CLI::App`) to the main `CLI::App` with an option `dt`. As CLI11 automatically assigns the variables with the parsed values given to it during option adding, the use of getters is only for checking parameters and possibly doing some operations on them.

The getters should be encapsulated in the classes that also add parameters, for example in the object wrapper we have the methods `addPopulationParameters()` and `getPopulationParameters()`. Not to be confused with the `Readparameters::add` that does the actual parameter adding to the `CLI::App` instance which the `addPopulationParameters()` uses to add the parameters.

Parse happens in vlasiator.cpp and is structured as follows:

- add ParticlePopulations parameter from `object_wrapper.cpp`
- add parameters from `parameters.cpp`
- add `sysboundary.cpp` parameters

First parse happens here, and we only parse parameters added above since some of those parameters add more parameters based on the input. For example after parsing ParticlePopulations more parameters will be added for those particle populations in multiple places like sysboundary, projects and object wrapper.

After the first parse:

- Add Population specific parameters from `object_wrapper.cpp`

After this we handle `--help` and `--version` calls and a second parse for parsing the additional paremeters added for all the particle populations after the ParticlePopulations was parsed.  Similar thing happens for sysboundaries, we add more parameters based on what was passed to the initial sysboundary parameters for reading in the boundary types.

After the second parse:

- Population parameters are checked ( i.e the getter `getPopulationParameters` in object_wrapper)
- Add boundary specific parameters
- Add project parameters

Then the final parse to parse everything with `extras=false` so any unrecognized parameters will raise an error. Additionally, we parse the parameters added via `addComposing` which mimics the old way some paremeters were handled that CLI11 does not support.

Finally after this we use getters for sysboundaries and project, etc


ursg:
Just scrolling through the diff again, I see that RP::get has now completely gone away, instead replaced by the storage target directly being specified as part of RP::add. That's completely fine, and I would actually say that it makes the parameter handling code in the usage locations cleaner. So thumbs up for that!

But I'm a bit confused about the calling style of RP::add now: Sometimes you call it with explicit template specialization information, and sometimes without? Does it infer the type correctly if it is unabiguous (like std::string) and needs some guidance when it chooses the wrong numeric type?

PR #1272, opened 2026-04-27
Roihu CPU 🔥

ykempf:
Collection of things I did to the repo while working on roihu_cpu
- MPI-IO switch changed – needs a better mechanism maybe/put it in Makefile.ARCH?
- use VLASIATOR_ARCH for folder of libraries
- allows to also use a centralised module-loading script using the same name
- fixed support for AGNER 8D
- included the cfg checking snippet discussed the other day

### DIFF
--- a/MAKE/Makefile.roihu_cpu
+++ b/MAKE/Makefile.roihu_cpu
@@ -62,7 +62,7 @@ testpackage: CXXFLAGS += -DUSE_JEMALLOC -DJEMALLOC_NO_DEMANGLE
 
 #======== Libraries ===========
 
-LIBRARY_PREFIX =  /projappl/project_2001659/pfaukemp/CPU/vlasiator-libraries/libraries-roihu_cpu/
+LIBRARY_PREFIX =  $(CURDIR)/libraries-roihu_cpu/
 
 # module libraries
 LIB_PAPI = -lpapi
--- a/MAKE/Makefile.roihu_cpu
+++ b/MAKE/Makefile.roihu_cpu
@@ -70,6 +70,10 @@ LIB_BOOST = -lboost_program_options
 
 # compiled libraries
 
+LIB_PROFILE = -L./submodules/phiprof/lib/ -lphiprof -Wl,-rpath=$(CURDIR)/submodules/phiprof/lib
+LIB_VLSV = -L./submodules/vlsv/ -lvlsv -Wl,-rpath=$(CURDIR)/vlasiator/submodules/vlsv
+
+
 INC_ZOLTAN = -I$(LIBRARY_PREFIX)/include
 LIB_ZOLTAN = -L$(LIBRARY_PREFIX)/lib -lzoltan -Wl,-rpath=$(LIBRARY_PREFIX)/lib
 
--- a/MAKE/Makefile.roihu_cpu
+++ b/MAKE/Makefile.roihu_cpu
@@ -70,9 +70,11 @@ LIB_BOOST = -lboost_program_options
 
 # compiled libraries
 
-LIB_PROFILE = -L./submodules/phiprof/lib/ -lphiprof -Wl,-rpath=$(CURDIR)/submodules/phiprof/lib
-LIB_VLSV = -L./submodules/vlsv/ -lvlsv -Wl,-rpath=$(CURDIR)/vlasiator/submodules/vlsv
+INC_PROFILE = $(LIBRARY_PREFIX)/include
+LIB_PROFILE = -L$(LIBRARY_PREFIX)/lib -lphiprof -Wl,-rpath=$(LIBRARY_PREFIX)/lib
 
+INC_VLSV = $(LIBRARY_PREFIX)/include
+LIB_VLSV = -L$(LIBRARY_PREFIX)/vlsv -lvlsv -Wl,-rpath= $(LIBRARY_PREFIX)lib
 
 INC_ZOLTAN = -I$(LIBRARY_PREFIX)/include
 LIB_ZOLTAN = -L$(LIBRARY_PREFIX)/lib -lzoltan -Wl,-rpath=$(LIBRARY_PREFIX)/lib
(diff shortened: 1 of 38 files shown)


Makefile
```diff
@@ -59,7 +59,7 @@ FIELDSOLVER ?= ldz_main
 #Skip deprecated C++ bindings from OpenMPI
 COMPFLAGS += -D OMPI_SKIP_MPICXX
 # Allow MCA io to be set to ompio, otherwise the code is overriding and setting ^ompio. (OpenMPI only, no effect with other MPI implementations.)
-# COMPFLAGS += -DVLASIATOR_ALLOW_MCA_OMPIO
```
ykempf, line 62: This is the major news that OMPIO on OpenMPI 5 works, IO fails if going for the default.

vlasovsolver/vec.h
```diff
@@ -162,7 +162,11 @@ const int VEC_PER_BLOCK = (WID*VEC_PER_PLANE);
 //user Agner's AVX512 optimized datatypes, double precision accuracy
 #include "vectorclass.h"
 typedef Vec8d Vec;
+#if VECTORCLASS_H >= 20000
```
ykempf, line 165: This is the second "big news".

tools/check_vlasiator_cfg.sh
```diff
@@ -197,6 +197,22 @@ else
 fi
 
 
+for i in "x" "y" "z"
```
ykempf, line 200: This is a new sanity check for cfgs as discussed on Zulip last week.

ykempf (commented):
And all the rest is the new feature of having only one place in `modules` to record a given architecture's module set.

MAKE/Makefile.roihu_cpu
```diff
@@ -0,0 +1,87 @@
...
+
+FLAGS =
+
+#WID=8
+#WID=4
+#VECL=4
+#VECL=8
+
+# Roihu has OpenMPI 5 so this needs to be activated
+COMPFLAGS += -DVLASIATOR_ALLOW_MCA_OMPIO
```
ykempf, line 30: This is moved here instead of Makefile.

ykempf:
Note that in CI I have named some library folders and archives with the arch name, e.g. libraries-carrington became libraries-carrington_gcc_openmpi as that matches the name obtained when building the libraries with the script. Makes it easier to match the library tar file with the arch it's on.
However: there's still throughout `libraries-docker` whereas the ARCH flag for that is `github_actions`, and `ukkogpu` vs `ukko_dgx`. Should these be made consistent too?

build_fetched_libraries.sh
```diff
@@ -47,8 +47,9 @@ elif [[ $PLATFORM == "-leonardo_dcgp_intel" ]]; then
 elif [[ $PLATFORM == "-hile_cpu" || $PLATFORM == "-hile_gpu" || $PLATFORM == "-lumi_2403" ]]; then
    make -j $PARALLEL CCC=CC CC=cc CCFLAGS="-fpic -O2 -std=c++17 -DCLOCK_ID=CLOCK_MONOTONIC -fopenmp" LDFLAGS="-fopenmp"
 elif [[ $PLATFORM == "-lumi_hipcc" ]]; then
-    make -j $PARALLEL CC=hipcc
-    CCFLAGS="-fpic -O2 -std=c++17 -DCLOCK_ID=CLOCK_MONOTONIC -fopenmp" LDFLAGS="-lstdc++ -fopenmp -lgomp -lroctx64 -lroctracer64"
+   make -j $PARALLEL CC=hipcc CCFLAGS="-fpic -O2 -std=c++17 -DCLOCK_ID=CLOCK_MONOTONIC -fopenmp" LDFLAGS="-lstdc++ -fopenmp -lgomp -lroctx64 -lroctracer64"
```
ykempf, line 50: This was on 2 lines, so the flags weren't passed.

build_fetched_libraries.sh
```diff
@@ -47,8 +47,9 @@ elif [[ $PLATFORM == "-leonardo_dcgp_intel" ]]; then
...
+elif [[ $PLATFORM == "-roihu_cpu" ]]; then
+   make -j $PARALLEL CCC=mpic++ CCFLAGS="-fpic -O2 -std=c++17 -DCLOCK_ID=CLOCK_MONOTONIC -fopenmp" LDFLAGS="-fopenmp"
```
ykempf, line 52: I guess this won't hurt, and who knows, maybe will get rid of empty phiprofs. Of course 100 jobs with this and 200 jobs without running for 1 min each on 10 nodes did not get any empty phiprofs, so no idea if it actually helps.

MAKE/Makefile.roihu_cpu
ykempf: @kstppd

kstppd: Thanks Yann!

ykempf:
Last commit message here to bring it to your attention.

Move zoltan to using Trilinos package
Yann's branch atm, until the bugfix is upstreamed.
Includes stubs for roihu_cpu_aocc but that's WIP.
Also includes direct path to Yann's cmake as current one on roihu is too old.

ykempf:
I suppose you'll want this to put the Zoltan library path to lib64 in every Makefile? But the wiki instructions will be changed separately. Or what do you prefer?

modules/roihu_cpu.sh
kstppd, line 1: One minor issue is we have to chmod +x these new module files.

ykempf: You think so? they're being sourced/bashed everywhere so far.

ykempf:
Note this last commit, if that cleanup is not wished for/done incorrectly, let me know. But on roihu it left the dangling symlink and rerunning the script failed on symlinking the new build folder as the old link was still there.

ykempf:
Side-note FYI on the upcoming GPU version: this installs an Eigen for Asterix but Roihu has an Eigen module. Maybe you don't need your own install then.

modules/roihu_gpu.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+module load cmake
```
ykempf, line 2: Note this is not installed yet but is put on the list of modules to be installed.

ursg (approved):
build_libraries_riscv is failing now, but I would guess that's not up to *this* PR to resolve.

Otherwise thumbs up!

PR #1273, opened 2026-04-27
Timeclass branch with current dev merged up

alhom:
This should merge as-is so it should actually run current CI.

Let's see how much it dies.


vlasiator.cpp
```diff
@@ -1168,32 +1556,58 @@ int simulate(int argn,char* args[]) {
...
+         }
          addTimedBarrier("barrier-check-dt");
          if(dtIsChanged) {
             phiprof::Timer updateDtimer {"update-dt"};
             //propagate velocity space back to real-time
-            if( P::propagateVlasovAcceleration ) {
+            if( P::propagateVlasovAcceleration) {
                // Back half dt to real time, forward by new half dt
-               calculateAcceleration(mpiGrid,-0.5*P::dt + 0.5*newDt);
+               calculateAcceleration(mpiGrid,-0.5); //This sets cells back to previous TIME_R
```
alhom: This breaks the main codepath!

spatial_cells/spatial_cell_cpu.cpp
```diff
@@ -706,21 +894,21 @@ namespace spatial_cell {
          const Real b = population.sparseDynamicMinValue1 - k * population.sparseDynamicBulkValue1;
          Real x;
          if ( population.sparseDynamicAlgorithm == 1 ) {
-            x = this->populations[popID].RHO;
+            x = this->get_population(popID,timeclass).RHO;
```
alhom, line 899: Check that this RHO is filled

spatial_cells/spatial_cell_cpu.cpp
ursg, line 199: This implicitly assumes that the velocity_block_with_content_list is up to date for the same timeclass that this function has been invoked with.
If not the iterations below would fall short and leave further block_indices untouched.

ursg, line 223: This *also* assumes that the neighbor cell has its velocity_block_with_content_list correctly set for the timeclass we are now looking at. Is that always true at timeclass interfaces?

spatial_cells/spatial_cell_cpu.cpp
```diff
@@ -250,18 +286,18 @@ namespace spatial_cell {
...
                //No content, and also no neighbor have content -> remove
                //and increment rho loss counters
-               const Real* block_parameters = get_block_parameters(popID)+blockLID*BlockParams::N_VELOCITY_BLOCK_PARAMS;
+               const Real* block_parameters = get_block_parameters(popID,timeclass)+blockLID*BlockParams::N_VELOCITY_BLOCK_PARAMS;
                const Real DV3 = block_parameters[BlockParams::DVX]
                  * block_parameters[BlockParams::DVY]
                  * block_parameters[BlockParams::DVZ];
                Real sum=0;
                for (unsigned int i=0; i<WID3; ++i) {
-                  sum += get_data(popID)[blockLID*SIZE_VELBLOCK+i];
```
ursg, line 259: Do we actually need to collect and measure loss of ghosts? Or should this summation only be performed for timeclass -1?

vlasovsolver/vlasovmover.cpp
```diff
@@ -406,99 +625,314 @@ uint getAccelerationSubcycles(const SpatialCell* spatial_cell, const Real dt, co
...
+         tc-1 after-acc   |----x-------x-------x-------|
+         tc-1 after-trans |x-------x-------x-------x---|
+         tc-0 after-acc   |----o---x---o-------o---x---|
+         tc-0 after-trans |x---------------x-----------|
+         legend:
+         - x: true state 
+         - o: ghost state
+         */
+      int tc_d = i-spatial_cell->get_tc();
+      payload.timeclass = i;
```
ursg: This is confusing: The payload's timeclass is *first* set to tcToPropagate (which is the current timeclass of the cell proper) and then updated to be i?

vlasovsolver/vlasovmover.cpp
```diff
@@ -406,99 +625,314 @@ uint getAccelerationSubcycles(const SpatialCell* spatial_cell, const Real dt, co
...
+            std::cerr << "c " << spatial_cell->parameters[CellParams::CELLID] <<" " << i << "tc with " << spatial_cell->get_velocity_blocks(popID,i)->size() <<" blocks\n";
+
+            payload.dt = dt/pow(2,tc_d)*1/2;
+            outvec.push_back(payload);
+         }
+         else if (spatial_cell->get_timeclass_turn_v(i)) {
+            payload.dt = dt/pow(2,tc_d);
+            outvec.push_back(payload);
+         }
+         else if (spatial_cell->requested_timeclass_copy_ghosts.count(i) && spatial_cell->get_timeclass_turn_v(i)){
```
ursg: This else-if branch will only even be invoked if the previous ones evaluated to false, which means that ```spatial_cell->get_timeclass_turn_v(i)``` must have been false.
Hence, this if will also always be false and never be called.

alhom: This might explain some missing VDFs!

spatial_cells/spatial_cell_cpu.cpp
alhom: Since the ghosts participate in translation and may propagate losses from there, it is a good diagnostic to keep available.

vlasovsolver/vlasovmover.cpp
```diff
@@ -507,66 +941,146 @@ void calculateAcceleration(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+
+      //now cellsToPropagateSet contains all cells which have been propagated and whose moments need updating
+
+   } //else
+
+   std::cout << "---------------------------- ACC finished --------------------\n";
+
+   
+   //converting cellsToPropagateSet to vector
+   for (auto cell : cellsToPropagateSet) {
```
MaxenceAnt, line 1138: Why not put this in the if ? if dt=0 there is no cells ToPropagate

vlasovsolver/vlasovmover.cpp
```diff
@@ -507,66 +941,146 @@ void calculateAcceleration(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+      cellsToPropagateVector.push_back(cell);
    }
 
    // Recalculate "_V" velocity moments
-   calculateMoments_V(mpiGrid,cells,true,(dt==0));
+   calculateMoments_V(mpiGrid,cellsToPropagateVector,true,(dt==0));
```
MaxenceAnt: Because this should not do much if dt=0

MaxenceAnt: Because this should not do much if dt=0

vlasovsolver/vlasovmover.cpp
```diff
@@ -507,66 +927,146 @@ void calculateAcceleration(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
+                           //    }
+                           //    std::cerr << "\n";
+                           // }
+                           
+                           // if ( SC->get_timeclass_turn_v() == true){ // propagate only if it is the cell's turn)
+                           //    propagatedCells.push_back(cells[c]);
+                           //    cellsToPropagateSet.insert(cells[c]);
+                           // }
+                           // vector<AccelerationPayload> harvest;
+                           setAccelerationTimeGhosts(propagatePayloads, mpiGrid[cells[c]], 0, dt_cell);
```
MaxenceAnt: Shouldn't we need to put popID instead of 0 ?

vlasovsolver/vlasovmover.cpp
alhom: Moment recalculation may be needed even with `dt=0`.

vlasovsolver/vlasovmover.cpp
```diff
@@ -406,99 +625,300 @@ uint getAccelerationSubcycles(const SpatialCell* spatial_cell, const Real dt, co
...
+         else if (spatial_cell->get_timeclass_turn_v(i) || (P::tstep == 0 && P::fractionalTimestep == 0)) {
+            
+            // if it is slower-tc's turn, we are synced at after-trans state
+            // -> Copy state, but needs to acc by half-tc-0-dt ("always init")
+            std::cerr << __FILE__ << ":"<<__LINE__ << " c"<< (int)spatial_cell->parameters[CellParams::CELLID]<<"\n";
+            spatial_cell->set_population(spatial_cell->get_population(popID,spatial_cell->get_tc()),popID,i);
+            // spatial_cell->set_velocity_mesh_ghost(popID, i);
+            // spatial_cell->set_velocity_blocks_ghost(popID, i);
+            // spatial_cell->get_population(popID,i).N_blocks = spatial_cell->get_number_of_velocity_blocks(popID, i);
+            double dtt = 5./4.*dt/pow(2,tc_d);
```
MaxenceAnt: Why it is different there?

alhom:
There's now a trivial conflict with `dev`, leaving that unresolved until we get a bit more better handle on what is broken currently to not bang the CI too much.

vlasovsolver/vlasovmover.cpp
```diff
@@ -507,66 +924,146 @@ void calculateAcceleration(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>&
...
-               acceleratedCells.swap(temp);
+               
+               propagatePayloads.swap(temp);
             }
             // Accelerate population over one subcycle step
-            calculateAcceleration(popID,(uint)globalMaxSubcycles,step,mpiGrid,acceleratedCells,dt);
+            // std::cout << "--------------------- an ACC subcycle " << step << " ------------------\n";
+            calculateAcceleration(popID,globalMaxSubcycles,step,mpiGrid,propagatePayloads,dt);
+            for(auto& payload:propagatePayloads){
+               cellsToPropagateSet.insert(payload.cellptr->get_cellid());
```
MaxenceAnt, line 1116: If multipop, some cells will be duplicated

alhom: The std::set data structure handles that

spatial_cells/spatial_cell_cpu.hpp
```diff
@@ -400,66 +442,66 @@ namespace spatial_cell {
...
+   inline const Realf* SpatialCell::get_data(const uint popID, const int timeclass = -1) const {
       debug_population_check(popID);
-      return populations[popID].blockContainer->getData();
+      return this->get_velocity_blocks(popID, timeclass)->getData();
    }
 
-   inline Realf* SpatialCell::get_data(const vmesh::LocalID& blockLID,const uint popID) {
-      debug_population_check(popID,blockLID);
+   inline Realf* SpatialCell::get_data(const vmesh::LocalID& blockLID,const uint popID, const int timeclass = -1) {
+      debug_population_check(popID,timeclass, blockLID);
```
MaxenceAnt: THis debug is supposed to be popID, blockLID, timeclass

spatial_cells/spatial_cell_cpu.cpp
```diff
@@ -287,10 +323,10 @@ namespace spatial_cell {
     sense in given block.
     Also returns false if given block doesn't exist or is an error block.
     */
-   bool SpatialCell::compute_block_has_content(const vmesh::LocalID& blockLID,const uint popID) const {
+   bool SpatialCell::compute_block_has_content(const vmesh::GlobalID& blockLID,const uint popID, const int timeclass) const {
```
MaxenceAnt: This is supposed to be a vmesh::LocalID

alhom:
So I suggest pulling this to a new branch, considering _maintainers are allowed to edit_ doesn't seem to be working.

grid.cpp
```diff
@@ -1130,78 +1363,228 @@ void initializeStencils(dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Geometry>& mpi
...
+   for (auto it : neighborhood){
+      all_neighborhoods.emplace(it);
+   }
+   if (!mpiGrid.add_neighborhood(Neighborhoods::VLASOV_SOLVER_Z, std::vector<neigh_t>(neighborhood.begin(), neighborhood.end()))){
       std::cerr << "Failed to add neighborhood Neighborhoods::VLASOV_SOLVER_Z \n";
       abort();
    }
+   neighborhood.clear();
+
+   for (int d = -VLASOV_STENCIL_WIDTH-3; d <= VLASOV_STENCIL_WIDTH+3; d++) {
```
petschge: Why does this exist for the x direction but is commented out in y and z direction?

alhom: Looks like old debugging crud.

PR #1277, opened 2026-05-08
Refactor derivatives

petschge:
This PR refactors fieldsolver/derivatives.cpp in five clean commits. It doesn't change what is actually happening and passes all my tests and bring the code in line with how we are going to do things once the fsgrid rework lands.


petschge:
fsgrid has landed, so this PR no longer makes sense.

Issue #1257, opened 2026-04-13
Working example configurations


petschge:
Ensure that all the projects come with a working and runable cfg file. At least Dispersion, KHB and Magnetosphere currently fail that.

petschge:
This should be fixed now that the PR has been merged.

PR #1280, opened 2026-05-19
Fixing project config files

lassejsc:
#1257

Trying to fix the project config files such that they *should* work out of the box.

Currently all but Harris works due to missing wall_+/-.dat

Diffusion and Firehose were deprecated

KHB was replaced by the KHB from samples/

All should run roughly with the same slurm setup on turso as the testpackage. Some do need bit more memory.


lassejsc:
We still need to decide which projects should be deprecated if any.

lassejsc:
This will possibly have a slightly annoying merge conflict with #1280 , this should be prioritized and any mention of the removed projects (Diffusion and Firehose) should be removed from `projects/project.cpp` should fix it. 

Not a hard thing to quickly fix during merging but more of a reminded to the fact

ursg:
I suppose the CLI11 merge caused some small conflicts here

PR #1274, opened 2026-04-29
Initial makefile and library script adjustments for roihu-gpu

ursg:
This is somewhat oblique to #1272, but so are the node architectures!

PR #1281, opened 2026-05-21
Updated makefile, test script, and library script adjustments for roihu-gpu

MikaelHuppunen:
Makefile and library script adjustments based on #1274


ykempf (commented):
Same comment here? :) I suppose the small test script is needed from here maybe?

PR #1282, opened 2026-05-21
Better io file write path handling in config

### LINKED ISSUE #1231: Output not creating subdirectories
It's possible to set output names such as `system_write_file_name = bulk/bulk` but unless you create the bulk directory yourself, that fails with `FAILED TO WRITE GRID ATvlasiator.cpp 948` (also note the missing space on that error message).

lassejsc:
Made the file writing path handling etc better, now supports paths in filename and also ~actually makes those folders~, fixed missing space in the error in vlasiator.cpp also.
Has no issue with config like
```
system_write_file_name = foo/bar/bulk
system_write_path = ./testing
```

Fixes #1231

### DIFF
--- a/iowrite.cpp
+++ b/iowrite.cpp
@@ -1885,7 +1885,7 @@ bool writeGrid(
    phiprof::Timer writeReducedTimer {"writeGrid-reduced"};
    // Create a name for the output file and open it with VLSVWriter:
    stringstream fname;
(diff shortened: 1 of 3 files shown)


ursg (approved):
I like it!

JonasSuni:
CI looks clean enough

PR #1284, opened 2026-05-26
Update UCX_NET_DEVICES devices

lassejsc:


lassejsc:
ugh autoformat accident again

PR #1285, opened 2026-05-26
fix UCX_NET_DEVICES to use eth5 and mlx5 on carrington tests

lassejsc:
Update: no eth5, it shouldnt use that, instead it should use the mlx, had to add couple exports suggested by HPC people to fix this and got linked the following:

https://vuw-research-computing.github.io/raapoi-docs/advanced/OpenMPI_users_guide/


JonasSuni (approved):
Has been tested and is sensible

PR #1286, opened 2026-05-29
Dependency version bump

petschge:
Reminder that we regularly want to bump the pinned versions of our dependencies and see what breaks.


ursg:
I suppose this PR should be retargeted to point at dev?

ursg:
This has gotten some conflict with the CI update from yesterday. Let's see if that's quick to fix.

PR #1287, opened 2026-05-29
Add option to use NIF instead of dHT frame in IPShock

JonasSuni:
Compiles, but hasn't been tested yet.


JonasSuni:
Has now been tested, gives reasonable-looking results (see Zulip topic) compared to old version.

JonasSuni:
Did a proper vector potential-based interpolation of the magnetic field. Testing now.

JonasSuni:
Ok this does not work super well, continuing testing

PR #1294, opened 2026-06-08
Ionosphere downmap smoothtop

alhom:
Adds an option to filter the electric current density to-be-downmapped to the ionosphere.

New parameters:
```
ionosphere.downmapSamplingMode Pointwise | Boxcar27 [Pointwise]
ionosphere.downmapSamplingWidth double [1]
```

The first sets the smoothing method out of two given. Pointwise is the default and has the same behaviour as previously. Boxcar27 does a 27-point cube with the downmapping point in the middle and equally weights all points. The width of the cube can be controlled by the other parameter (default 1, Harrison-Stetson constant).

Ionosphere_small test results for reference:
Pointwise:
image

Boxcar27, Width = 0.4
image

Boxcar27, Width = 1
image

Notes:
- This samples equally-weighted samples. Adding e.g. Gaussian-weighted sampling is trivial. 
- Boxcar27 samples the CurlB interpolator and uses that instead of the Pointwise CurlB for the first approximation. Plenty of fiddling to be done on if the sampling should be done already on the FAC values, should the sampling stencil be of some magnetic-field driven shape... left for discussion.

### DIFF
--- a/sysboundary/ionosphere.cpp
+++ b/sysboundary/ionosphere.cpp
@@ -1920,15 +1920,12 @@ namespace SBC {
             int samplingFAC = 0;
             std::array<Real, 3> curlB;
             std::array<FsGridTools::FsIndex_t,3> lfsc = getLocalFsGridCellIndexForCoord(fsgrid, nodes[n].xMapped);
+
             // Local cell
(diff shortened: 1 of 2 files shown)


JonasSuni:
Doesn't change default behaviour, so worth having in dev

PR #1297, opened 2026-06-11
Added a thingy to skip copying .vlsv files from the tests/ folders

lassejsc:
Had annoying accident with this so I'm making a fix for it.


lassejsc:
I don't know why this seems to not behave nicely with slurm jobs, so closing

PR #1298, opened 2026-06-11
Set relative diff to 0 if absolute is 0

lassejsc:
Currently if absolute is 0 and the reference dataset is zeros it will print relative as -1, this changes it so if absolute is 0 the relative will be 0 too in such a case.


ursg (approved):
Good idea!

PR #1302, opened 2026-06-15
roihu_gpu small updates

ykempf:
This helps with the fact slurm inherits the modules upon sbatch.

Note cuda/13.0.2 is available with gcc 15 but this was now working off @kstppd's existing scripts.


ykempf:
And the explanation for not using the latest version is that something is not working with it.

ykempf:
And another commit to activate PAPI memory reports on roihu_gpu too.

ykempf (commented):
And then see what I added here – I ran some GPU tests with this.

PR #1307, opened 2026-06-23
Adding fieldline tracing to the magnetosphere_3d_small test

lassejsc:
Does what it says on the tin, I think this could be good addition since I realized the testpackage doesn't test that. The fieldline tracing params are from the FHA run (comment if something is wonky with the params)


ykempf (approved):
Looks sensible I guess. And I assume you'll take my 2 cents as gold so I'll tag this as "approve". ;)

PR #1308, opened 2026-06-23
Vlsvdiff error raise

lassejsc:
Testing adding check for the return value in case the vlsvdiff returns some non zero exit value.

Not sure if this works how i imagine it would work.


ursg:
This is now in a conflicted state, but that might be as trivial as renaming the run file.

PR #1310, opened 2026-06-24
Gathering GPU memory report from all tasks

MikaelHuppunen:


arch/gpu_base.cpp
```diff
@@ -260,6 +260,32 @@ __host__ uint gpu_getAllocationCount() {
...
+   }
+
+   using get_pool_size_function = std::size_t(*)();
+   auto poolSizeFunction = reinterpret_cast<get_pool_size_function>(dlsym(handle, "get_pool_size"));
+   if (!poolSizeFunction){
+      return 0;
+   }
+
+   return poolSizeFunction();
+}
```
ursg, line 287: This might be the dirtiest hack I have seen in the Vlasiator codebase *by far*.

I love it, but do we really want to go to these depths?

ursg:
Apart from the horrible, horrible hack you made there to get pancake memory statistics, this looks pretty straightforward.

Should we... I don't know... just merge it?

arch/gpu_base.cpp
MikaelHuppunen: If there's a cleaner way, I'm open to suggestions. It works, and handles the cases where MPI Pancake is not preloaded or the function is not implemented. I'll leave it up to you to decide if it's too dirty XD

PR #1311, opened 2026-06-25
CI for LUMI

lassejsc:
Just a testing branch now

todo:
 - actually testing
 - the srun commands should be checked (will probably make some more generalized srun thing maybe through a script or alias?

Currently i just set it to do the hipcc one.


lassejsc:
This is a test, merging into a new test branch since we suspect that they github secret wont work unless it is in a workflow in a branch on the main repo

PR #1312, opened 2026-06-26
CI rewrite thingy

lassejsc:
Possible CI rewrite, the whole CI workflow is an absolute spaghetti mess tbh and I think it's time to clean it up.

My idea is to generalize it  as much as possible such that different `srun` parameters are handled in another bash script and then using some yaml matrix shenanigans to make it all work.

Here for testing since that's the best way to test.


lassejsc:
See #1320

PR #1316, opened 2026-06-29
Improve fetch(_and_build)_libraries.sh

pedr-nz:
- [X] Shallow clone git repos
 - [ ] ~~Allowing script to run in other directories~~

### DIFF
--- a/build_fetched_libraries.sh
+++ b/build_fetched_libraries.sh
@@ -197,7 +197,7 @@ cd build
 cmake .. \
     -DTOCTREE_L2ERROR=true \
     -DCMAKE_BUILD_TYPE=Release \
-    -DEigen3_DIR="$WORKSPACE/submodules/eigen/build/" \
+    -DEigen3_DIR="$SH_FILES_DIR/submodules/eigen/build/" \
     -Dzfp_DIR="$ZFP" \
     -DCMAKE_INSTALL_PREFIX=$WORKSPACE/libraries${PLATFORM}
 make install
--- a/fetch_and_build_libraries.sh
+++ b/fetch_and_build_libraries.sh
@@ -4,7 +4,7 @@ set -e   # Abort on error
 
 WORKSPACE=`pwd`
 
-SH_FILES_DIR=`dirname $0`
+export SH_FILES_DIR=`dirname $0`
 if [[ z$SH_FILES_DIR == "z" ]]; then
 	SH_FILES_DIR="."
 fi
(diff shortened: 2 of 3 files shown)


fetch_and_build_libraries.sh
```diff
@@ -27,12 +32,12 @@ else
    ln -s $BUILDDIR library-build
 fi
 
-bash ./fetch_libraries.sh ${PLATFORM:1}
+bash $(SH_FILES_DIR)/fetch_libraries.sh ${PLATFORM:1}
```
lassejsc: This should be without the parenthesis ()

pedr-nz: Ah yeah true, how about should I wrap the whole string in quote or not? The rest of the code doesn't seem to do that but I usually do.

fetch_and_build_libraries.sh
```diff
@@ -4,6 +4,11 @@ set -e   # Abort on error
 
 WORKSPACE=`pwd`
 
+SH_FILES_DIR=`dirname $0`
```
lassejsc: Does this actually work? 

Note that in the build script for example it uses $WORKSPACE which i believe is set to pwd 

```
cmake .. \
blabla
    -DEigen3_DIR="$WORKSPACE/submodules/eigen/build/" \
blabla
```
so I'm not sure it will build correctly as it will now download all the libraries to the current folder which may not be the root folder. I think this could just be removed as I don't see a point in ever fetching the libraries to non root folder of the repo...

fetch_and_build_libraries.sh
lassejsc: I think either one should work. I would personally use without quotes

fetch_and_build_libraries.sh
pedr-nz: I'm trying to build libraries in separate folders for different machine without having to just cloning the repo over and over (or rebuilding them every time I switch between machines).

pedr-nz: But yeah I didn't notice this, went to sleep yesterday after pultti's CMake version doesn't satisfy requirement.

petschge:
didn't we have a PR running around that included $VLASIATOR_ARCH into the name of the libraries directory?

fetch_and_build_libraries.sh
lassejsc: valid, though i'm more thinking if the end user will accidentally run it in another folder...

lassejsc:
> didn't we have a PR running around that included $VLASIATOR_ARCH into the name of the libraries directory?

That is already merged i believe? If you leave $1 blank it will use $VLASIATOR_ARCH

fetch_libraries.sh
lassejsc, line 1: These changes are very nice

pedr-nz:
Yeah I guess we can just merge the commit that organise commit hash, and don't use the further commits that try to build them in different directories. 
I was just trying to see if it worked last night because somehow different cubbli machines don't cross-run on things built on other cubbli machine.

build_fetched_libraries.sh
lassejsc, line 171: Note that there is a spot here too.

I still think it would be better to not have it be possible to run it in different directories since it can cause more issues for the end-user etc or have some sort of a warning that requires user input, but im on the side of not having it.

Not to mention it makes the future changes to the build scripts ever so slightly less straight forward if you want to maintain that.

pedr-nz: Yeah I also spotted that just now as I'm testing the script, but I agree we don't necessarily need that ability. Should I just close this PR and made another PR just for the shadow-cloning on fetch_libraries.sh?

lassejsc: > Yeah I also spotted that just now as I'm testing the script, but I agree we don't necessarily need that ability. Should I just close this PR and made another PR just for the shadow-cloning on fetch_libraries.sh?

Yes I think that would be good

PR #1317, opened 2026-06-30
Shallow clone git repos to save space&time

pedr-nz:


pedr-nz:
Should we also pin Trilinos to a specific commit?

petschge:
I'm tempted to say yes. Not sure why Yann pinned it to a branch instead of a commit or tag.

ykempf:
I'm waiting on progress on https://github.com/trilinos/Trilinos/pull/15236 and am expecting there might be more commits on that branch, which would allow you to continue tracking that progress without having to update the commit in the build script.

pedr-nz:
> I'm waiting on progress on https://github.com/trilinos/Trilinos/pull/15236 and am expecting there might be more commits on that branch, which would allow you to continue tracking that progress without having to update the commit in the build script.

I can maybe make the git_use_commit function take optional hash, and with that we can leave the hash empty for now which will track HEAD and pin it once it's finished? What does Patrick think?

petschge:
works for me

ursg:
Can you rebase this on top of the current dev again, now that CI has been cleaned up? (If only just to get a new set of CI checkmarks)

PR #1318, opened 2026-07-01
Fix vslv writer bug in ompio 4.x

petschge:
By automagically setting OMPI_MCA_io=^ompio when required. This fixes the vlsvextract bug that Lauri was running into.


JonasSuni (approved):
Looks good to me, has been tested by Patrick and it works, and it doesn't touch main vlasiator so it's fine.

PR #1300, opened 2026-06-11
Made it possible to launch block adjust functions on GPU with more than 65535 cells

MikaelHuppunen:

PR #1319, opened 2026-07-01
Launching prepare to receive blocks for all cells at once in grid

MikaelHuppunen:


spatial_cells/spatial_cell_gpu.cpp
```diff
@@ -33,6 +33,8 @@
 
 using namespace std;
 
+const int maxCellsPerIteration = 65535;
```
ursg, line 36: This constant is now redundant with the one defined in PR #1300, isn't it?

PR #1320, opened 2026-07-02
CI cleanup 

lassejsc:
Will document more later,  for now the gist of it is that I've combined the different workflows for different architectures into a single matrix job which makes the workflow a ton cleaner.

In addition i've made some use of the yaml tags `&` to reuse steps etc, this also makes things cleaner if a new test is added etc.

Right now it should work and do all the same things, there are still things that maybe unnecessary like uploading some artifacts.

It does make the workflow overview less cool but more concise 

Some basic documentation:

# github CI

### Turso testpackage
The testpackage that runs on turso does the following in summary:
- Build libraries
- Build tools
- Compile vlasiator binary
  - There are couple relevant flags in the strategy matrix in the `.github/workflows/github-ci.yml`
    - `compile_tp`: `true` will compile with the `testpackage` flag and `false` without it. Latter is should be referred to as `prod`/`production` in the CI.
    - `debug`: `true` will use compilation flags from `compile_flags_tp`/`compile_flags_prod` (`compile_tp:true`/`false`) in `.github/workflows/slurm_run.sh`
- Run testpackage 
    - ONLY if `run_tp:true` and `compile_tp:true` then the github ci test script `testpackage/small_test_[VLASIATOR_ARCH]_github_ci.sh` will be run.
      - For example: `small_test_hile_cpu_github_ci.sh` for hile_cpu
      - This will also run the fluxfunction test.
 

## Adding new platform to test
Adding a new platform to test in the CI is a roughly a two(three) step process:
- Add the platform specific run commands/flags to `.github/workflows/slurm_run.sh`
- Add a new matrix entry for platform in the `.github/workflows/github-ci.yml` 
- (Optionally) Add a new `small_test_[VLASIATOR_ARCH]_github_ci.sh` script for the testpackage tests. Note that this is likely not trivial and currently only couple architectures have this.  

### run/compilation flags
Each platform needs its specific srun flags defined, these are done in the  `.github/workflows/slurm_run.sh`. There are 4 different type of flags:
- `core_flags`: the number of cores/tasks/nodes to use for compiling or heavier srun calls
- `constraint`: for setting the partition etc to use for compiling the testpackage
- `constraint_small`: for setting the constraint for lighter calls like build libs/file handling
- `mem_flags`: the amount of memory to use for compilation

For now the small_test testpackage flags etc are called with sbatch so these are set in the script itself.

Compilation flags can be set in `compile_flags_tp`/`compile_flags_prod` when `debug:true`. See the script for an example.

### Adding a new matrix entry to the workflow
See the matrix entries in the workflow file, the format is the following
```
          - name: prod         #name that is displayed in github summary
            runner: hile       #Tag of the github runner to run on
            arch: hile_gpu     #VLASIATOR_ARCH
            compile_tp: false  #whether to compile w/ testpackage flag
            run_tp: false      #run testpackage
            debug: false       #whether to use the custom defined compilation flags
```
See the summary in [Turso testpackage](#turso-testpackage) for the boolean flags. 

(The link above should push you to the correct section but i havent tested it since it doesnt seem to maybe work as well in the PR preview)


lassejsc:
This is in a condition that can be reviewed but since the runners are still down it probably should be tested across all platforms

ursg:
Oooh, those shell scripts are pretty, I like this a lot!

PR #1324, opened 2026-07-02
Testing lumi CI 

lassejsc:
empty PR to see if not touching the workflow file will now run the lumi CI with the secret key


lassejsc:
Removing this as the secrets only work for pushes but no PRs

PR #1327, opened 2026-07-09
Fixed balsaraFsgrid such that it compiles, runs and produces *something*

lassejsc:
Not sure whether it makes sense or not, I added parameter reading etc to it since it seemed logical there because I don't know how it was supposed to set the parameters previously. Now you can just set the params via the usual `--run_config` for example and it will skip the unknown parameters.

A lot of the compilation was trial and error + looking at the ionosphereSolver app


ursg:
I suppose this will now need to be updated *again*, with CLI11 in the main branch? Also, if it is fixed and working, should it be automagically build as part of CI to ensure that it keeps working?

Issue #1163, opened 2025-08-21
Test restart reading on same and different task counts


ykempf:
Restart IO is done differently when the task count changes, which should be tested as well.

PR #1328, opened 2026-07-10
testpackage addition: restart read on single(or multiple) mpi rank

lassejsc:
Since the restart IO depends on the number of ranks here https://github.com/fmihpc/vlasiator/blob/dev/ioread.cpp#L1422 . By default the restarts are written with the `--ntasks` number of tasks so reading them with a single rank will use the `else` part of the code there.  Not sure if this fully answers to #1163

Update:

Changed `single_cell` in `test_definitions_small.sh` to be `ranks` which sets the number of ranks to be used when running the test. Setting this to `"half"` as done in the new restart read tests will used half the $SLURM_NTASK ranks (floored division)


ykempf:
Single rank in itself is already a good test but I think another rank count, but parallel, would be needed, say 8 ranks if the base case is run on 16, or so.

PR #1332, opened 2026-08-04
Creating less custom datatypes to lower Open MPI memory consumption

MikaelHuppunen:


ursg:
This can go into the CPU spatial cell just as well, I suppose.

Even further, we could consider even identifying if that whole hindexed data type *happens* to lie in a continuous memory block, and send it as such, thus taking some amount of work away from the pancake.

MikaelHuppunen:
> This can go into the CPU spatial cell just as well, I suppose.
> 
> Even further, we could consider even identifying if that whole hindexed data type _happens_ to lie in a continuous memory block, and send it as such, thus taking some amount of work away from the pancake.

Yes, I did test this also on the CPU side. In my test case, the OMPI local copies of data types did not accumulate, so it made no difference. I should test it with a case where the problem is present on the CPU side, to see if it's effective. I can also check whether the continuous memory block occurs enough to make it worth implementing.

PR #1334, opened 2026-08-27
Bump dccrg commit

pedr-nz:
Bump the dccrg to the commit where boost dependency in the main header is removed.


ursg:
This is basically identical to #1333 and just points the submodule pointer directly to the correct commit id

ursg:
This is rebased on top of the CL11 PR now, and *should* completely remove any boost requirement.

ursg:
I can confirm by compiling locally (and not wresting the half-broken CI into shape first) that this compiles and links completely without boost now!

PR #1337, opened 2026-09-01
Fix typo in sysboundary.cpp

JonasSuni:


ykempf (approved):
Leet it be, this is a good PR to go in. But wait for the CI to complete!

PR #1340, opened 2026-09-01
Remove the backgroundfield/tests directory that's unused since 2014

ursg:
It doesn't compile anymore.
It references headers that don't exist anymore.
Let's toss it!


ykempf:
Whose test was that anyway?

ykempf:
Though having some better practices than cargo-cult in the Romberg integration magic of the background field would not hurt. ;) In another PR I guess...

PR #1341, opened 2026-09-01
Fix vlsvextract where CLI11 requires positional argument to be named

pedr-nz:


JonasSuni (approved):
Fixes vlsvextract so that it works as before

PR #1345, opened 2026-09-07
Add lumi runner to github-ci matrix.

ursg:
Will fail in its current form, but at least it should give us the *correct* error message.

### DIFF
--- a/testpackage/small_test_lumi_2503_github_ci.sh
+++ b/testpackage/small_test_lumi_2503_github_ci.sh
@@ -19,8 +19,8 @@ reference_dir="/projappl/project_462001599/testpackage"
 cd $SLURM_SUBMIT_DIR
 #cd $reference_dir # don't run on /proj
 
-bin="/pfs/lustrep4/projappl/project_462001599/actions-runner-linux-x64-2.337.0/_work/vlasiator/vlasiator/vlasiator"
-diffbin="/pfs/lustrep4/projappl/project_462001599/actions-runner-linux-x64-2.337.0/_work/vlasiator/vlasiator/vlsvdiff_DP"
+bin="/projappl/project_462001599/actions-runner-linux-x64-2.337.0/_work/vlasiator/vlasiator/vlasiator"
+diffbin="/projappl/project_462001599/actions-runner-linux-x64-2.337.0/_work/vlasiator/vlasiator/vlsvdiff_DP"
 
 #compare agains which revision
 reference_revision="current"
--- a/testpackage/small_test_lumi_2503_github_ci.sh
+++ b/testpackage/small_test_lumi_2503_github_ci.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+#SBATCH -t 01:30:00        # Run time (hh:mm:ss)
+#SBATCH --job-name=ctestpackage
+#SBATCH -A project_462001599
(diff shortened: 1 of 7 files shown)


testpackage/small_test_lumi_2503_github_ci.sh
```diff
@@ -0,0 +1,64 @@
...
+# folder for all reference data
+reference_dir="/users/lrb_462001599_ci/testpackage"
+cd $SLURM_SUBMIT_DIR
+#cd $reference_dir # don't run on /proj
+
+bin="/pfs/lustrep3/users/lrb_462001599_ci/actions-runner-linux-x64-2.337.0/_work/vlasiator/vlasiator/vlasiator"
+diffbin="/pfs/lustrep3/users/lrb_462001599_ci/actions-runner-linux-x64-2.337.0/_work/vlasiator/vlasiator/vlsvdiff_DP"
+
+#compare agains which revision
+reference_revision="current"
```
JonasSuni, line 26: Make sure that you have a symlink "current" that points to the same directory as "CI_reference" (the fact that we refer to two different ref directories still confuses me greatly and I don't understand it)

testpackage/small_test_lumi_2503_github_ci.sh
```diff
@@ -0,0 +1,64 @@
...
+# if 0 then we check the v1
+create_verification_files=0
+
+# folder for all reference data
+reference_dir="/users/lrb_462001599_ci/testpackage"
+cd $SLURM_SUBMIT_DIR
+#cd $reference_dir # don't run on /proj
+
+bin="/pfs/lustrep3/users/lrb_462001599_ci/actions-runner-linux-x64-2.337.0/_work/vlasiator/vlasiator/vlasiator"
+diffbin="/pfs/lustrep3/users/lrb_462001599_ci/actions-runner-linux-x64-2.337.0/_work/vlasiator/vlasiator/vlsvdiff_DP"
```
JonasSuni: So these then imply that the ci will always build to the _work directory and not a random temporary directory etc. Is this intentional?

JonasSuni (commented):
Seems sensible to me at first glance

testpackage/small_test_lumi_2503_github_ci.sh
petschge: I have that, yeah

testpackage/small_test_lumi_2503_github_ci.sh
petschge: That is how it's currently set up. Not sure if that is smart.

JonasSuni: If you only have one runner, then I guess it's fine. However, should the compilation be done in /projappl and the running on /scratch? Or did the LUMI support give you instructions to run everything in /users? What's the quota on the different disks for the robot account?

ursg: A second runner would have its own _work, too. So I think this is fine.

JonasSuni: This is a hardcoded path that points to the robot account's home directory though, wouldn't that always be the same?

JonasSuni (approved):
Looks good to me, at least for now.

PR #1346, opened 2026-09-07
Pencil datareducer

veetihaaja:
As per title, added data into CellParams and further into a datareducer to show

1. What pencil that cell is part of (locally)
2. If the cell is a seed cell for pencil generation
3. If the cell is an active cell (with ghost translation)

Note: the pencil IDs are local, so in an MPI simulation, the id can change at process boundaries

### DIFF
--- a/vlasovsolver/cpu_trans_pencils.cpp
+++ b/vlasovsolver/cpu_trans_pencils.cpp
@@ -1514,6 +1514,38 @@ void prepareSeedIdsAndPencils(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Ge
    }
    buildPencilsTimer.stop();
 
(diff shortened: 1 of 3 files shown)


vlasovsolver/cpu_trans_pencils.cpp
```diff
@@ -1418,6 +1423,40 @@ void prepareSeedIdsAndPencils(const dccrg::Dccrg<SpatialCell,dccrg::Cartesian_Ge
    vector<CellID> seedIds;
    getSeedIds(mpiGrid, propagatedCells, dimension, seedIds);
    getSeedIdsTimer.stop();
+
+
+   for (auto c : localCells){
```
ykempf: Not a very optimal loop – and is this running only when an IO is upcoming?

PR #1349, opened 2026-09-12
Attempt to fix samples/ cfg files by removing sameline comments

ursg:
This might even turn CI green! Fingers crossed.


ursg:
Ok, this fixed samples/Magnetosphere3D, but the others are still red. So those are up next!

ursg:
Check cfg is green!

PR #1352, opened 2026-09-14
Update CLI11 to version 2.7.2

ursg:
This fixes internal compiler errors on roihu_gpu.


ursg:
All CI tests built and ran just fine, except for filesystems, http-caches and github all conspiring against us. Anyway, good enough.

PR #1356, opened 2026-09-16
Fix PAPI library include and library flags to be version agnostic

JonasSuni:
For the lumi_2503 makefile.


JonasSuni:
This has been tested and it works on LUMI, showing the memory reports in at least one manually run test job

PR #1357, opened 2026-09-16
use qdparser and hack a new parsing system

kstppd:
Why not ?


readparameters.cpp
```diff
@@ -21,88 +21,298 @@
...
+      line.erase(std::remove_if(line.begin(), line.end(), [](unsigned char ch) { return std::isspace(ch); }), line.end());
+      if (line.empty()) {
+         continue;
+      }
+
+      if (line.front() == '[') {
+         if (line.back() == ']') {
+            section = line.substr(1, line.size() - 2);
+         } else {
+            std::cerr << "Invalid configuration line, found a line starting with '[' which doews not  end wit "
```
ursg: There's a *typo* here! Literally unmergable!

ursg (changes requested):
This is a remarkably smaller amount of code than CLI11. If it works just as well, why not?

readparameters.cpp
alhom: Two, even!

ursg:
(Also it doesn't compile, so maybe fix that first)

ursg (approved):

PR #1360, opened 2026-09-16
Replace git submodules with fetch_submodules.sh

ykempf:
fsgrid, dccrg, eigen, vectorclass, vectorclass-addon and hashinator are no longer tracked as git submodules; fetch_libraries.sh now calls fetch_submodules.sh to clone them into ./submodules/. CI workflows fetch them explicitly (or via the existing fetch_libraries.sh / fetch_and_build_libraries.sh calls) after checkout instead of using actions/checkout's submodules option.

Here's our suggestion to alleviate some woes encountered by vlasiator cloners.


ursg (changes requested):
How about doing shallow clones of the no-longer-submodules, too?

JonasSuni (commented):
I like it! Since we're anyway always fetching and building the other libraries now, it makes more sense to have everything in the same (or similar) workflow. As Yann said, this also avoids issues with remembering to get the submodules as well as with updating the commits (I know it's easy for some people but I still struggle when having to update submodules)

ykempf:
And both myself and @petschge seem to have been able to build from fresh clones with little hassle.

ykempf:
Suggestion: remove the README mention of these libraries altogether as it's no longer going to be such a top-level acute issue. (Can be pushed later to let the current CI iteration complete.)
Instead, have an update of the wiki on this too.

alhom:
Sure! Good to have a single workflow.

`rm -rf` for the submodule folders though - if I need to update my libraries while having something fixed/tested/modified in our own submodules for development I would be rather crass if I lost that work unexpectedly. Guarding against existing modifications/WIP submodules repos is IMO needed.

ursg (approved):

alhom (approved):

PR #1361, opened 2026-09-16
Build phiprof with nvtx ranges on roihu_gpu

ursg:
For funky tracing and profiling!


ykempf:
Worked for me at least!
image

PR #1363, opened 2026-09-18
Get off stupid ukko nodes

petschge:
like ukko-paavo for the carrington_gcc_openmpi CI runs.


JonasSuni (approved):
Looks good

ursg:
Should we wait for this to actually *run* CI on ukko before merging it?

PR #1366, opened 2026-09-18
add a tiny expression parser to enable expressions in config files

kstppd:
This is backwards compatible hopefully.
What is happening here is that we are just enabling expressions to be evaluated in config files
You can even define local variables with a dollar sign in front of them!
So we can now do things like 
```
[proton_vspace]
local velres = 40e3 # 40 km/s resolution
vx_length = 26 
vy_length = 26
vz_length = 26
vx_min = -.5 * vx_length * 4 * velres
vx_max =  .5 * vx_length * 4 * velres
vy_min = -.5 * vy_length * 4 * velres
vy_max =  .5 * vy_length * 4 * velres
vz_min = -.5 * vz_length * 4 * velres
vz_max =  .5 * vz_length * 4 * velres
```


ursg:
> You can even define local variables with a dollar sign in front of them!

The PR description is outdated!

ursg:
For proper software engineering practices, I think this should have some test coverage. Can you pop in some expressions into at least one of the testpackage runs? And/or add some test that specifically tests the behaviour with invalid syntax (so that we do not end up debugging misbehaviour due to a typo in production for three days...)

Maybe your LLM friends would be up for this kind of task. :)

ursg:
Also, vlsvdiff does not compile with the current state of this PR.

PR #1368, opened 2026-09-21
Change access to zeroth element

petschge:
of std containers to use .data() instead which is not undefined (but most likely still a bad idea) on empty containers.

This should not actually be merged, but serves as a list of files that need manual review. To generate it, I ran

```
sudo apt install clang-tidy bear libomp-dev
export VLASIATOR_ARCH=home
git clone git@github.com:fmihpc/vlasiator.git vlasiator_containercheck
cd vlasiator_containercheck
git checkout dev
./fetch_and_build_libraries.sh home
make clean
bear -- make -B -j 4
mkdir -p build && mv compile_commands.json build/
run-clang-tidy -p build -checks='-*,readability-container-data-pointer' -fix
``

### CODE REVIEW

**Comment by petschge** [comment]
Using
```
run-clang-tidy-19 -p build -checks='-*,readability-container-data-pointer' -header-filter='.*' -fix
```
I found two more occurances.

